PR #4, opened 2024-05-21
ui: bootstrap react pages, dependencies and github action

edlerd:


ui/babel.config.js
```diff
@@ -0,0 +1,29 @@
+module.exports = {
```
mas-who: for context, is babel required by vite?

ui/index.html
```diff
@@ -0,0 +1,19 @@
+<!doctype html>
+
+<html lang="en" dir="ltr">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+
+    <title>LXD Site Manager</title>
+    <link rel="shortcut icon" href="/assets/img/favicon-32x32.png" type="image/x-icon">
```
mas-who, line 9: probably not relevant for now, but we should discuss with @piperdeck and core team for a logo.

ui/package.json
```diff
@@ -0,0 +1,82 @@
+{
+  "name": "lxd-site-manager-ui",
+  "version": "0.0.1",
+  "author": "Canonical Webteam",
+  "license": "LGPL-3.0-only",
+  "type": "module",
+  "scripts": {
+    "clean": "rm -rf node_modules yarn-error.log *.log build/ .jekyll-metadata .bundle playwright-report test-results build",
+    "build": "npx vite build",
+    "format-js-eslint": "eslint 'src/**/*.{json,jsx,tsx,ts}' 'tests/**/*.ts' --fix",
```
mas-who, line 13: Since we use typescript, maybe these scripts should follow the *-ts-* pattern? e.g. format-ts-eslint, format-ts etc

ui/package.json
```diff
@@ -0,0 +1,82 @@
...
+    "format-js-prettier": "prettier 'src/**/*.{json,jsx,tsx,ts}' 'tests/**/*.ts' --write",
+    "format-js": "yarn format-js-eslint && yarn format-js-prettier",
+    "lint-scss": "stylelint 'src/**/*.scss'",
+    "lint-js-eslint": "eslint 'src/**/*.{json,tsx,ts}' 'tests/**/*.ts' .eslintrc.cjs babel.config.js",
+    "lint-js-prettier": "prettier 'src/**/*.{json,tsx,ts}' 'tests/**/*.ts' .eslintrc.cjs babel.config.js --check",
+    "lint-js": "yarn lint-js-eslint && yarn lint-js-prettier",
+    "hooks-add": "husky install",
+    "hooks-remove": "husky uninstall",
+    "start": "vite --host",
+    "test-e2e": "npx playwright test"
```
mas-who: I think we would still have unit tests, should we use vitest similar to how we do it in lxd-ui?

ui/package.json
```diff
@@ -0,0 +1,82 @@
+{
```
mas-who, line 1: code coverage will be interesting for this project. What would be the steps to combine coverage reports?

ui/package.json
```diff
@@ -0,0 +1,82 @@
...
+    "format-js-eslint": "eslint 'src/**/*.{json,jsx,tsx,ts}' 'tests/**/*.ts' --fix",
+    "format-js-prettier": "prettier 'src/**/*.{json,jsx,tsx,ts}' 'tests/**/*.ts' --write",
+    "format-js": "yarn format-js-eslint && yarn format-js-prettier",
+    "lint-scss": "stylelint 'src/**/*.scss'",
+    "lint-js-eslint": "eslint 'src/**/*.{json,tsx,ts}' 'tests/**/*.ts' .eslintrc.cjs babel.config.js",
+    "lint-js-prettier": "prettier 'src/**/*.{json,tsx,ts}' 'tests/**/*.ts' .eslintrc.cjs babel.config.js --check",
+    "lint-js": "yarn lint-js-eslint && yarn lint-js-prettier",
+    "hooks-add": "husky install",
+    "hooks-remove": "husky uninstall",
+    "start": "vite --host",
```
mas-who, line 22: I assume for this project we would mainly serve the UI via the go API, even in development? I guess we wouldn't need haproxy anymore, but we would have to build the ui first. I wonder if there is some kind of scripting mechanism in go to achieve this in a project like this?

ui/package.json
```diff
@@ -0,0 +1,82 @@
...
+    "test-e2e": "npx playwright test"
+  },
+  "dependencies": {
+    "@canonical/react-components": "0.52.0",
+    "@tanstack/react-query": "5.17.15",
```
mas-who: should we use the latest versions for these dependencies since it would be easy to upgrade now? i.e. `yarn upgrade --latest`. Tried this locally, ui builds fine and playwright also still works.

ui/package.json
```diff
@@ -0,0 +1,82 @@
+{
+  "name": "lxd-site-manager-ui",
+  "version": "0.0.1",
```
mas-who, line 3: should we enforce a specific node version to use?

ui/package.json
```diff
@@ -0,0 +1,82 @@
...
+    "postcss-cli": "11.0.0",
+    "prettier": "3.2.4",
+    "sass": "1.69.7",
+    "stylelint": "16.1.0",
+    "stylelint-config-standard-scss": "13.0.0",
+    "stylelint-order": "6.0.4",
+    "stylelint-prettier": "5.0.0",
+    "stylelint-scss": "6.0.0",
+    "typescript": "5.3.3",
+    "vite": "5.2.8",
```
mas-who: Should we still have the rollup resolution for vite similar to that in lxd-ui? Assuming later on lxd-site-manager snap would have to be built on weird OS architectures

ui/playwright.config.ts
```diff
@@ -0,0 +1,58 @@
...
+  retries: process.env.CI ? 2 : 0,
+  /* Opt out of parallel tests on CI. */
+  workers: 1,
+  /* Reporter to use. See https://playwright.dev/docs/test-reporters */
+  reporter: process.env.CI ? "blob" : "html",
+  /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
+  use: {
+    /* Maximum time each action such as `click()` can take. Defaults to 0 (no limit). */
+    actionTimeout: 0,
+    baseURL: "http://localhost:8414/",
```
mas-who, line 32: this will probaly change later to the go API assets endpoint.

ui/tsconfig.json
```diff
@@ -0,0 +1,41 @@
+{
```
mas-who, line 1: can we add a tsconfig.json to the root of the project with the following content:

```
{
    "extends": "./ui/tsconfig.json",
}
```

This is more for VS code to enable IntelliSense. VS code uses the project root to find a tsconfig.json, since when working on the lxd-site-manager project, we would have the whole project open instead of just the /ui directory, IntelliSense breaks. It's not a critical issue, but it would make life easier for VScode users on the team.

.github/workflows/pr.yaml
```diff
@@ -0,0 +1,27 @@
+name: PR checks
+on: pull_request
+
+jobs:
```
mas-who, line 4: should we add the CLA check now? That's relevant for all canonical projects right?

mas-who (commented):
Thanks for this! First steps of a great project :) Left some comments, not all of them is critical for current stage of the project so feel free to disregard them where applicable.

ui/babel.config.js
edlerd: nope, removing this.

ui/package.json
edlerd: I'd keep it consistent among our projects.

ui/package.json
edlerd: Yes, vitest seems right, adding the cmd.

ui/package.json
edlerd: I'd run go with an entrypoint script and put haproxy in front to have the vite dev server with live updates to develop the ui. I have a follow up PR nearly ready with this setup. Would keep it for a 2nd step.

ui/package.json
edlerd: Yes, updated all the things to latest just now.

ui/package.json
edlerd: We can add it once the need for it arises. Would keep it clean for now.

ui/playwright.config.ts
edlerd: Yeah, true. For now it seem right though.

ui/tsconfig.json
edlerd: Adding it.

ui/playwright.config.ts
mas-who: Aah yes, this is just me thinking out loud

.github/workflows/pr.yaml
edlerd: good idea, adding it now.

ui/package.json
edlerd: yes, good idea.

ui/package.json
edlerd: Yes, that's a fun challenge for the future. I think go has a a coverage mode, so we can run the go backend with that enabled and trigger e2e tests and collect the coverage for go easily. As for ts we can use the same solution as for lxd-ui.

ui/index.html
edlerd: I think we can use the lxd logo.

ui/package.json
mas-who: Yeah, that's definitely a fun task for the future :slightly_smiling_face:

mas-who (approved):

mas-who:
Thanks for the changes :+1:

PR #6, opened 2024-05-28
ui: add dev server to run the ui easily, update readme

edlerd:
- added dotrun configuration to combine the ui with the backend and have hot reloading enabled locally
- update readme with renamed commands.


ui/README.md
```diff
@@ -2,13 +2,25 @@
...
+
+Run the backend go server with yarn on your host
+    
+    cd ui
+    yarn backend-run
+
+Bootstrap the database on the first run. This step can be skipped on successive runs
+
+    cd ui
+    yarn backend-init
```
mas-who, line 15: are we considering to have the ui as the entry point for development?

ui/haproxy.cfg
```diff
@@ -0,0 +1,21 @@
...
+  bind BIND_ADDRESS
+  acl is_api path_beg /1.0
+  use_backend go_backend if is_api
+  default_backend ui_serve
+
+backend ui_serve
+  server ui_server 127.0.0.1:3000
+
+backend go_backend
+  http-request replace-header Host .* 172.19.0.1:9001
```
mas-who: why do we have the backend running on 172.19.0.1 and not just 127.0.0.1?

ui/package.json
```diff
@@ -19,9 +19,10 @@
     "lint-js": "yarn lint-js-eslint && yarn lint-js-prettier",
     "hooks-add": "husky install",
     "hooks-remove": "husky uninstall",
-    "start": "vite --host",
-    "test-e2e": "npx playwright test",
-    "test-js": "vitest --run"
+    "start": "./entrypoint 0.0.0.0:${PORT}",
+    "backend-run": "cd .. && make && go run ./cmd/lxd-site-mgrd --state-dir ./state/dir1/",
+    "backend-init": "cd .. &&  go run ./cmd/lxd-site-mgr --state-dir ./state/dir1/ init \"member1\" 172.19.0.1:9001 --bootstrap",
```
mas-who: I think we should probably add `go run ./cmd/lxd-site-mgr --state-dir state/state_dir_1 waitready` before the initialisation to ensure the daemon has started.

ui/src/pages/sites/SiteList.tsx
```diff
@@ -1,7 +1,36 @@
...
+import { queryKeys } from "util/queryKeys";
+import { fetchSites } from "api/sites";
+import { useQuery } from "@tanstack/react-query";
+import { MainTable } from "@canonical/react-components";
 
 const SiteList: FC = () => {
-  return <div>Here be sites</div>;
+  const { data: sites = [] } = useQuery({
+    queryKey: [queryKeys.sites],
+    queryFn: () => fetchSites(),
```
mas-who:
```suggestion
    queryFn: fetchSites,
```

ui/src/types/site.d.ts
```diff
@@ -0,0 +1,5 @@
+export interface LxdSite {
+  Addresses: string[];
```
mas-who: I don't think these fields should be capitalised.

mas-who:
Works nicely with HAProxy, thanks for this! Just a few minor comments.

Is it worthwhile to figure out later how to spin up the dev server using go instead?

ui/haproxy.cfg
```diff
@@ -0,0 +1,21 @@
+global
+  daemon
+
+defaults
+  mode  http
+  timeout connect 500000
+  timeout client  500000
+  timeout server  500000
+
+frontend anbox_proxy
```
mas-who:
```suggestion
frontend site_manager_proxy
```

ui/src/types/site.d.ts
tomponline: LxdSite should be LXDSite

ui/README.md
edlerd: I was thinking this as the readme for a ui developer. Though we might not have the distinction ui backend developer on the project yet.

ui/haproxy.cfg
edlerd: Dotrun is starting haproxy in a docker container together with the vite dev server. We reach the backend on  172.19.0.1 from inside dotruns docker container.

ui/package.json
edlerd: Didn't encounter an issue with this yet, but agree this is a good idea. Will change it.

ui/src/types/site.d.ts
edlerd: This is capitalized because the backend sends it currently like this. Once we fix it there, we can update the types as well as the usages. Currently using lower cased fields will fail as field names in javascript are case sensitive.

edlerd: As for LxdSite / LXDSite I'd keep the former as that is consistent with lxd-ui.

ui/haproxy.cfg
mas-who: I see, that makes sense. But when I run `yarn backend-init` with `go run ./cmd/lxd-site-mgr --state-dir ./state/dir1/ init \"member1\" 172.19.0.1:9001 --bootstrap`, it fails with `Error: Post "http://control.socket/cluster/control": context deadline exceeded`. Not sure if you've seen this happen? Changing the address to 127.0.0.1:9001 resolved the issue for me

edlerd: Are you using `yarn run-backend` to start the backend? You should, otherwise you might start it on a different ip or port.

mas-who: Yeah that's the sequence, first `yarn backend-run` then `yarn backend-init`. See screencast below, not sure if I'm doing something wrong?

[Screencast from 04-06-2024 09:10:59.webm](https://github.com/canonical/lxd-site-manager/assets/151511568/03645a62-d305-4fd3-a295-faae9fc96b68)

ui/src/types/site.d.ts
tomponline: Hrm OK, strange that the lxd-ui is using "LXD" incorrectly as well, but consistency amongst UIs has value too.

ui/haproxy.cfg
edlerd: I slightly updated the config so the bootstrap listens on all interfaces. Hope that step succeeds.

Though haproxy needs the hosts ip and is still using 172.19.0.1 in haproxy.cfg. Is this ip maybe different on your machine?

ui/src/types/site.d.ts
tomponline: > This is capitalized because the backend sends it currently like this. Once we fix it there, we can update the types as well as the usages. Currently using lower cased fields will fail as field names in javascript are case sensitive.

Makes sense. Where can I see the proposed API docs and struct definitions for review before the UI progresses far in consuming them? Thanks 

In general I believe we should try and stick with the conventions that the LXD API uses (i.e lower cased field names in the actual API JSON, which can then optionally be mapped back to differently cased language specific struct fields as the language allows).

E.g.

```
{
  "metadata": [
    "/1.0"
  ],
  "status": "Success",
  "status_code": 200,
  "type": "sync"
}
```

edlerd: The spec on [lxd site manager](https://docs.google.com/document/d/1q0mlkekLO7iTVnktHFJU2VTGAE9ojtR7av1N3QHZGo8/edit#heading=h.gd8pfy4tx04g) is detailing some of the structures we aim for.

I agree with the proposed api response structure. Site manager actually already is responding in exactly this way. So the specific fields are under metadata and status/status_code and type are on the top level of the json response. I think this is coming in from the microcluster framework code.

tomponline: @roosterfish @masnax care to chime in here please? WRT to JSON field name capitalisation in REST APIs and whether microcluster has diverged from the conventions used by LXD?

ui/haproxy.cfg
mas-who: Yeah my IP is 10.0.0.139, I tested using this IP and it's working now. Isn't this IP different per machine? Maybe we can use something like `hostname -I | awk '{print $1}'` to set a environment variable for the host address?

edlerd: We can't run the hostname in a dotrun script as we are already in the container.
In lxd-ui we use an environment variable to set this ip, because it can differ based on your setup, I'll make the change here accordingly.

ui/src/types/site.d.ts
edlerd: I think the upper case field names come from lxd-site-manager code. (edit: i.e. [here](https://github.com/canonical/lxd-site-manager/blob/main/api/sites.go#L43))

@tomponline Does LXD define field names in lowercase also in the database and go internal structs? Or only the api fields are lowercase and the go-internal names start with a capital letter?

mas-who: I think this is just due to the field tags not being added. The `Site` struct was setup during the sprint, it was a quick example to show how to use microcluster and not representative of what we have in the spec.

Edit: if the Site struct in api/types/site is updated to:
```
type Site struct {
	Name      string   `json:"name"`
	Addresses []string `json:"address"`
	Status    string   `json:"status"`
}
```
Microcluster should take care of the json serialisation accordingly.

tomponline: > Does LXD define field names in lowercase also in the database and go internal structs? Or only the api fields are lowercase and the go-internal names start with a capital letter?

Yes the field names in the DB are lower case underscore separated, take a look at

https://github.com/canonical/lxd/blob/main/lxd/db/cluster/schema.go#L8

When we talk about the REST API, the underlying fields in the actual JSON transferred over the wire are lower case underscore separated, but they are marshalled/unmarshalled into Go structs that have UC first field names (and camel case) so they are exported in Go and comply with Go's linter rules.

edlerd: Thanks for clarifying, I included the change to have the lowercase json field names and updated the types in the ui code.

tomponline: Thanks!

mas-who (approved):

PR #9, opened 2024-06-04
feat: serve ui from go server [WD-11656]

mas-who:
## Done

1. added api endpoints to serve ui static assets
2. updated ui build configs to include `/1.0/ui` as base prefix for routing

## Notes
1. The static assets will be copied from the ui directory and embedded in the go binary during compilation
2. multiple endpoints must be registered for various static file paths due to limitation from microcluster where api endpoints are registered individually behind `/1.0` prefix.
3. For development, the UI is served from vite dev server which can be reached on `http://0.0.0.0:8414/`. For testing and production environments (future), the UI is served from each cluster members API e.g. `https://0.0.0.0:9001/1.0/ui`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,8 @@ default: build
 # Build targets.
 .PHONY: build
 build:
+	cd ui && yarn build
+	cd api && go generate
 	go install -v ./cmd/lxd-site-mgr
 	go install -v ./cmd/lxd-site-mgrd
 
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,6 @@ ui/node_modules
 ui/playwright-report
 test-results
 haproxy-local.cfg
+.vscode
+api/static
+*__debug*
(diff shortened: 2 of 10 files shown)


edlerd (commented):
Code looks good, when I checkout and run make it fails atm. Shall we update the makefile to also run `cd ui && yarn build` and `cd api && go generate` for make to succeed on 1st run?

mas-who:
> Code looks good, when I checkout and run make it fails atm. Shall we update the makefile to also run `cd ui && yarn build` and `cd api && go generate` for make to succeed on 1st run?

Aah good idea, will add that quick

Makefile
```diff
@@ -4,6 +4,8 @@ default: build
 # Build targets.
 .PHONY: build
 build:
+	cd ui && yarn build
+	cp -r ui/build/ui api/static
```
edlerd: Should this use `go:generate`?

edlerd (approved):
LGTM, just one question below.

Makefile
mas-who: 100%, sorry not sure how I missed that

PR #11, opened 2024-06-04
feat: add oidc authentication support [WD-11658]

mas-who:
## Done

- [x] Copied over oidc package from lxd (including latest zitadel v3 update)
- [x] Setup global state for site manager and pass to all endpoint handlers
- [x] Setup oidc endpoints for site manager (`/oidc/login`, `/oidc/callback`, `/oidc/logout`)
- [x] Removed what's not needed for site manager from oidc package
- [x] Implemented authentication check on protected api endpoints
- [x] Added UI login tests
- [x] Adjusted demo server setup for oidc authentication


auth/auth.go
```diff
@@ -0,0 +1,598 @@
...
+		return nil, AuthError{Err: fmt.Errorf("Failed to verify access token: %w", err)}
+	}
+
+	// Check that the token includes the configured audience.
+	audience := claims.GetAudience()
+	if o.audience != "" && !shared.ValueInSlice(o.audience, audience) {
+		return nil, AuthError{Err: fmt.Errorf("Provided OIDC token doesn't allow the configured audience")}
+	}
+
+	userInfo, err := rp.Userinfo(accessToken, oidc.BearerToken, claims.Subject, o.relyingParty)
```
markylaing: One of the reasons LXD needs to cache identities is so that the email address of a user that logs in via a bearer token can be found without this call. The OIDC library will make a HTTP request here to the IdP to get user information on every request.

auth/auth.go
```diff
@@ -0,0 +1,598 @@
...
+	err := o.ensureConfig(r.Host)
+	if err != nil {
+		_ = response.ErrorResponse(http.StatusInternalServerError, fmt.Errorf("OIDC callback failed: %w", err).Error()).Render(w)
+		return
+	}
+
+	handler := rp.CodeExchangeHandler(func(w http.ResponseWriter, r *http.Request, tokens *oidc.Tokens[*oidc.IDTokenClaims], state string, rp rp.RelyingParty) {
+		sessionID := uuid.New()
+		secureCookie, err := o.secureCookieFromSession(sessionID)
+		fmt.Printf("SecureCookie: %v\n", secureCookie)
```
markylaing: Looks like some debug logging?

auth/auth.go
```diff
@@ -0,0 +1,598 @@
...
+}
+
+// WriteHeaders writes the OIDC configuration as HTTP headers so the client can initatiate the device code flow.
+func (o *Verifier) WriteHeaders(w http.ResponseWriter) error {
+	w.Header().Set("X-LXD-OIDC-issuer", o.issuer)
+	w.Header().Set("X-LXD-OIDC-clientid", o.clientID)
+	w.Header().Set("X-LXD-OIDC-audience", o.audience)
+
+	return nil
+}
```
markylaing: These headers written to the response when OIDC authentication fails. The CLI uses them to perform a device login flow in it's end and get an access token. You could use a similar approach, but I'd recommend changing the headers themselves e.g. `X-LXD-OIDC-issuer` -> `X-LXD-site-manager-OIDC-issuer`.

cmd/lxd-site-mgrd/main.go
```diff
@@ -86,9 +104,21 @@ INSERT INTO sites_addresses (site_id, address) VALUES (3, 'https://192.168.0.2:8
...
+			// The oidc verifier needs the cluster cert for generating secure cookies and we only have access to cluster cert after the daemon has started.
+			oidcVerifier.SetClusterCert(s.ClusterCert)
+			return nil
+		},
+	}
+
+	oidcEndpoints := []rest.Endpoint{
+		api.OidcLoginCmd(oidcVerifier),
+		api.OidcCallbackCmd(oidcVerifier),
+		api.OidcLogoutCmd(oidcVerifier),
```
markylaing: When changing OIDC configuration, you will likely need access to the `oidcVerifier` in scope of a handler. It might be a good idea to have your own state type, similar to the microcluster state which has the `oidcVerifier` as a field, then pass that as an argument when instantiating your handler funcs.

auth/auth.go
```diff
@@ -381,8 +381,19 @@ func (o *Verifier) setRelyingParty(host string) error {
...
+	proto := getHeaderOrDefault(r, "X-Forwarded-Proto", "")
+	if proto == "" {
+		if r.TLS != nil {
+			proto = "https"
+		} else {
+			proto = "http"
+		}
+	}
+	host := getHeaderOrDefault(r, "X-Forwarded-Host", r.Host)
+	callbackURL := fmt.Sprintf("%s://%s/1.0/oidc/callback", proto, host)
```
markylaing: I'm not sure what the best approach is here. This is a good idea, but IIRC I've also had this working with [sticky sessions](https://www.haproxy.com/blog/enable-sticky-sessions-in-haproxy). We'll have to have a think about the behaviour of LXD when behind a load balancer and perhaps we can come to a general solution.

auth/auth.go
```diff
@@ -381,8 +381,19 @@ func (o *Verifier) setRelyingParty(host string) error {
...
+	// Construct the callback URL based on X-Forwarded headers if applicable
+	proto := getHeaderOrDefault(r, "X-Forwarded-Proto", "")
+	if proto == "" {
+		if r.TLS != nil {
+			proto = "https"
+		} else {
+			proto = "http"
+		}
+	}
+	host := getHeaderOrDefault(r, "X-Forwarded-Host", r.Host)
```
markylaing: Generally for code style we like to have a newline after each code block or each logical set of calls.

markylaing:
Last comment for now is more about repo structure. Golang codebases now tend to follow [this convention](https://github.com/golang-standards/project-layout?tab=readme-ov-file#go-directories). Many of our repos don't follow this convention, I'm not certain as to the reasons for this, but it's good practise to try and follow. 

Since you don't want other go programs to import any of this codebase yet, I would initially create an `internal` directory and then move your `auth` and `api` packages into it :)

auth/auth.go
mas-who: I see, this is referring to the [identityCache](https://github.com/canonical/lxd/blob/9387ded1aee7888f37dd1a8714cada3f7ad31d7e/lxd/auth/oidc/oidc.go#L150) right? From what I can see, the `identityCache` is coupled to some local concepts in LXD. Am I correct to say that in site manager we only need the `Identifier`, `Name` and `Subject` fields from the `identity.CacheEntry` struct? I wonder if we should import the identity package as a dependency in site manager to reuse the existing cache structure, not sure if it will work since site manager does not have tls authentication and also we won't be implementing fine grained authorisation yet. Maybe we can setup a site manager version of the cache wdyt?

auth/auth.go
mas-who: We are not planning to implement oidc for CLI in site manager. I was actually going to ask how auth validation works in LXD once the user authenticates. Based on what you said, does that mean we should also implement oidc for CLI if we want to leverage what LXD has for auth validation?

cmd/lxd-site-mgrd/main.go
mas-who: Thanks will do that :slightly_smiling_face:

auth/auth.go
mas-who: Will look into sticky sessions. I think there is a wider discussion here as well. Are we happy with progressing to make relevant changes in the site manager version of the oidc package for now, get a sense of the requirement then later look into generalising the oidc package in lxd?

auth/auth.go
mas-who: aah right, the thinking here is that the logical set of calls is for constructing the callback url, I think putting this block into a function might improve readability?

mas-who:
> Last comment for now is more about repo structure. Golang codebases now tend to follow [this convention](https://github.com/golang-standards/project-layout?tab=readme-ov-file#go-directories). Many of our repos don't follow this convention, I'm not certain as to the reasons for this, but it's good practise to try and follow.
> 
> Since you don't want other go programs to import any of this codebase yet, I would initially create an `internal` directory and then move your `auth` and `api` packages into it :)

This is great! :smile: will make a separate PR to readjusted the codebase structure

tomponline:
> > Last comment for now is more about repo structure. Golang codebases now tend to follow [this convention](https://github.com/golang-standards/project-layout?tab=readme-ov-file#go-directories). Many of our repos don't follow this convention, I'm not certain as to the reasons for this, but it's good practise to try and follow.
> > Since you don't want other go programs to import any of this codebase yet, I would initially create an `internal` directory and then move your `auth` and `api` packages into it :)
> 
> This is great! 😄 will make a separate PR to readjusted the codebase structure

Yes please make good use of `internal` to avoid becoming an unexpected dependency elsewhere.

.gitignore
```diff
@@ -1,10 +1,12 @@
 .idea
 state
+!internal/state
```
edlerd:
```suggestion
internal/state
```
Typo?

ui/haproxy.cfg
```diff
@@ -8,15 +9,20 @@ defaults
...
-  use_backend go_backend if is_api !is_ui
+  # These are required headers for site manager if user is using a proxy or load balancer 
+  http-request set-header X-Forwarded-For %[src]
+  http-request set-header X-Forwarded-Host %[req.hdr(host)]
+  http-request set-header X-Forwarded-Port %[dst_port]
+  http-request set-header X-Forwarded-Proto https
+  bind BIND_ADDRESS ssl verify optional crt keys/lxd-sm.pem ca-file keys/lxd-sm.crt
+  acl is_backend path_beg /1.0 path_beg /oidc
+  acl is_ui path_beg /ui
+  use_backend go_backend if is_backend !is_ui
```
edlerd:
```suggestion
  use_backend go_backend if is_backend
```
Can simplify here.

internal/api/api.go
```diff
@@ -0,0 +1,94 @@
...
+	"github.com/canonical/lxd/shared/api"
+	microclusterState "github.com/canonical/microcluster/state"
+
+	"github.com/canonical/lxd-site-manager/internal/identity"
+	"github.com/canonical/lxd-site-manager/internal/oidc"
+	"github.com/canonical/lxd-site-manager/internal/state"
+)
+
+type endpointHandler func(*microclusterState.State, *http.Request) response.Response
+type endpointHandlerWithState func(*state.SiteManagerState, *microclusterState.State, http.ResponseWriter, *http.Request) response.Response
```
edlerd:
```suggestion
```
I don't think we need this.

internal/api/api.go
```diff
@@ -0,0 +1,94 @@
...
+
+		return response.ManualResponse(func(w http.ResponseWriter) error {
+			if a.SetCookies {
+				err := a.WriteCookies(w, s.OidcVerifier)
+				if err != nil {
+					return response.Forbidden(err).Render(w)
+				}
+			}
+
+			return handler(s, innerState, w, r).Render(w)
```
edlerd:
```suggestion
			return handler(innerState, r).Render(w)
```
The other two params are not needed in the handlers.

internal/api/sites.go
```diff
@@ -8,32 +8,45 @@ import (
...
+		Get: rest.EndpointAction{
+			Handler:        wrapHandlerWithAuth(s, siteGet),
+			AllowUntrusted: true,
+			AccessHandler:  authHandler(s),
+		},
+	}
 }
 
-func sitesGet(s *state.State, r *http.Request) response.Response {
+func sitesGet(s *state.SiteManagerState, innerState *microclusterState.State, w http.ResponseWriter, r *http.Request) response.Response {
```
edlerd:
```suggestion
func sitesGet(innerState *microclusterState.State, r *http.Request) response.Response {
```
Can this be removed?

edlerd: Also in the siteget below.

edlerd (changes requested):

auth/auth.go
mas-who: Since we would be using access token verification currently in site manager, `identityCache` isn't really needed. Although I did setup a simpler version of `identityCache` here to mimic the same flow in lxd, not sure if I should remove this rather?

internal/api/api.go
mas-who: As discussed, we will keep the interface simple for now :slightly_smiling_face:

mas-who:
@edlerd  I've added e2e tests for login/logout as well. It requires use of repository secrets though so CI will fail on a forked PR before this is merged. Also I don't have access to Settings in this repo. Here is a success [action run](https://github.com/mas-who/lxd-site-manager/actions/runs/9676264964) for the login e2e tests on my fork.

.github/workflows/pr.yaml
```diff
@@ -101,7 +102,7 @@ jobs:
       - name: Install Dotrun
         run: |
           sudo pip3 install dotrun requests==2.31.0 # requests version is pinned to avoid breaking changes, can be removed once issue is resolved: https://github.com/docker/docker-py/issues/3256
-
+
```
edlerd:
```suggestion

```

internal/api/sites.go
```diff
@@ -8,32 +8,45 @@ import (
...
+		Get: rest.EndpointAction{
+			Handler:        wrapHandlerWithAuth(s, siteGet),
+			AllowUntrusted: true,
+			AccessHandler:  authHandler(s),
+		},
+	}
 }
 
-func sitesGet(s *state.State, r *http.Request) response.Response {
+func sitesGet(innerState *microclusterState.State, r *http.Request) response.Response {
```
edlerd:
```suggestion
func sitesGet(s *microclusterState.State, r *http.Request) response.Response {
```
Keep naming simple? Also relevant below.

internal/api/types/rest.go
```diff
@@ -0,0 +1,11 @@
...
+
+import (
+	"net/http"
+
+	"github.com/canonical/lxd/lxd/response"
+	"github.com/canonical/microcluster/state"
+)
+
+// EndpointHandler represents a function that handles an API endpoint.
+type EndpointHandler func(*state.State, *http.Request) response.Response
```
edlerd: Should this type be upstreamed to microcluster?

ui/haproxy.cfg
```diff
@@ -8,15 +9,19 @@ defaults
...
-  bind BIND_ADDRESS
-  acl is_api path_beg /1.0
-  acl is_ui path_beg /1.0/ui
-  use_backend go_backend if is_api !is_ui
+  # These are required headers for site manager if user is using a proxy or load balancer 
+  http-request set-header X-Forwarded-For %[src]
+  http-request set-header X-Forwarded-Host %[req.hdr(host)]
+  http-request set-header X-Forwarded-Port %[dst_port]
+  http-request set-header X-Forwarded-Proto https
+  bind BIND_ADDRESS ssl verify optional crt keys/lxd-sm.pem ca-file keys/lxd-sm.crt
```
edlerd: The bind should be the first line, so the config list the ingress first and the egress last. In between, we have handling of the request.

internal/api/types/rest.go
mas-who: Yes we should add it to this [file](https://github.com/canonical/microcluster/blob/main/rest/rest.go) in microcluster. Will create an issue for it.

ui/haproxy.cfg
mas-who: yeah that makes sense :+1:

edlerd (approved):
LGTM :+1:

auth/auth.go
mas-who: As per previous discussions, we will look into forwarding requests to the correct host for this.

auth/auth.go
mas-who: As per previous discussion, we do not need these headers in site manager for now since we are not implementing authentication for the CLI. Will keep these headers for now in case if we need to implement device grant flow later.

edlerd (approved):

mas-who:
@edlerd @markylaing the latest commit implements forwarding idp callback request to the member that initiated the login flow, please have a look when you get some time. 

FYI: There is a bug in microcluster at the moment which will cause CI to fail. I have a [PR](https://github.com/canonical/microcluster/pull/163) open to fix it.

Edit: Fix was merged so CI should pass now.

.gitignore
```diff
@@ -1,5 +1,6 @@
...
```
markylaing: I'm finding this commit a bit confusing. What does it have to do with OIDC? Should it be in a separate PR to be merged first?

This particular line is also a bit confusing to me. Why was `state` ignored to begin with?

internal/oidc/oidc.go
```diff
@@ -23,6 +18,11 @@ import (
 	"github.com/zitadel/oidc/v3/pkg/oidc"
 	"github.com/zitadel/oidc/v3/pkg/op"
 	"golang.org/x/crypto/hkdf"
+
+	"github.com/canonical/lxd/lxd/identity"
+	"github.com/canonical/lxd/lxd/response"
+	"github.com/canonical/lxd/shared"
+	"github.com/canonical/lxd/shared/api"
```
markylaing: I don't think these imports should have moved. Imports should be in three blocks:
1. Go stdlib imports.
2. External non-stdlib dependencies.
3. In-project package dependencies.

Each block should be in alphabetical order. The `gci` linter should catch this.

internal/identity/cache.go
```diff
@@ -0,0 +1,80 @@
...
+
+// CacheEntry represents an oidc identity.
+type CacheEntry struct {
+	Identifier   string
+	Name         string
+	IdentityType string
+	Subject      string
+}
+
+// Get returns a single CacheEntry by its authentication method and identifier.
```
markylaing: The comment is out of date now.

internal/identity/cache.go
```diff
@@ -0,0 +1,80 @@
...
+	// identifier is the email address of an oidc identity
+	entries map[string]*CacheEntry
+	mu      sync.RWMutex
+}
+
+// CacheEntry represents an oidc identity.
+type CacheEntry struct {
+	Identifier   string
+	Name         string
+	IdentityType string
```
markylaing: Do we need `IdentityType` if they are all OIDC identities?

internal/oidc/oidc.go
```diff
@@ -10,6 +10,10 @@ import (
 	"strings"
 	"time"
 
+	"github.com/canonical/lxd/lxd/identity"
+	"github.com/canonical/lxd/lxd/response"
+	"github.com/canonical/lxd/shared"
+	"github.com/canonical/lxd/shared/api"
```
markylaing: These have moved around a few times between commits.

internal/oidc/oidc.go
```diff
@@ -10,7 +10,7 @@ import (
 	"strings"
 	"time"
 
-	"github.com/canonical/lxd/lxd/identity"
+	"github.com/canonical/lxd-site-manager/internal/identity"
```
markylaing: As this import is in the same module it should be in the last block.

cmd/lxd-site-mgrd/main.go
```diff
@@ -65,6 +70,26 @@ func (c *cmdDaemon) run(cmd *cobra.Command, args []string) error {
...
+	// setup global oidc verifier
+	httpClientFunc := func() (*http.Client, error) {
+		return util.HTTPClient("", nil)
+	}
+
+	oidcVerifier, err := oidc.NewVerifier(
+		oidc.OidcIssuer,
+		oidc.OidcClientID,
+		oidc.OidcAudience,
+		func() *shared.CertInfo { return &shared.CertInfo{} },
```
markylaing: Why is this returning an empty certificate? I think this needs to be the cluster certificate.

internal/oidc/oidc.go
```diff
@@ -24,6 +24,13 @@ import (
 	"golang.org/x/crypto/hkdf"
 )
 
+// TODO: these should be removed.
+const (
+	OidcIssuer   = "https://dev-xjrvvfikbsv4jxn7.us.auth0.com/"
+	OidcClientID = "gpSHEK5SooxiKlFU1lx4xnWJMf78ydwy"
+	OidcAudience = "https://dev-xjrvvfikbsv4jxn7.us.auth0.com/api/v2/"
+)
```
markylaing, line 22: As you have said, these should be removed. Can you get these from a environment variables for now?

cmd/lxd-site-mgrd/main.go
markylaing: Never mind, I see in the next commit that the cluster cert is being set on a microcluster hook. Perhaps the verifier can be set then as well?

internal/oidc/oidc.go
```diff
@@ -287,7 +287,7 @@ func (o *Verifier) Callback(w http.ResponseWriter, r *http.Request) {
 
 		// Send to the UI.
 		// NOTE: Once the UI does the redirection on its own, we may be able to use the referer here instead.
-		http.Redirect(w, r, "/ui/", http.StatusMovedPermanently)
+		http.Redirect(w, r, "/ui", http.StatusMovedPermanently)
```
markylaing: Is this related to OIDC?

ui/entrypoint
```diff
@@ -7,9 +7,19 @@ cp haproxy.cfg haproxy-local.cfg
...
 
+# generate certificates for dev environment
+if [ ! -d "keys" ]; then
+    mkdir -p keys
+    openssl req -nodes -x509 -newkey rsa:2048 -keyout keys/lxd-sm.key -out keys/lxd-sm.crt -subj "/C=GB/ST=London/L=London/O=LXD SM/OU=dev/CN=localhost" -days 3000
+    openssl dhparam -out keys/dhparams.pem 2048
+    cat keys/lxd-sm.key keys/lxd-sm.crt > keys/lxd-sm.pem
+    cp keys/lxd-sm.key keys/lxd-sm.crt.key
+    echo 'finished generating certificates'
+fi
```
markylaing: I think LXD has some utils for generating certificates that match what it generates by default if that is of interest? https://github.com/canonical/lxd/blob/762f7dc5c3dc4dbd0863a796898212d8fbe3f7c3/test/includes/certificates.sh

Also if this is for development, do you need the certificate to be valid for 3000 days?

internal/oidc/oidc.go
```diff
@@ -595,3 +597,26 @@ func NewVerifier(issuer string, clientID string, audience string, clusterCert fu
 
 	return verifier, nil
 }
+
+func getHeaderOrDefault(r *http.Request, headerName string, defaultValue string) string {
```
markylaing: Now that the member address is encoded in the OIDC state, do you need these functions for getting headers? Also, do they rely upon the correct configuration of the external load balancer?

internal/api/api.go
```diff
@@ -4,7 +4,85 @@ import (
...
+	return func(innerState *microclusterState.State, r *http.Request) response.Response {
+		if r.TLS == nil {
+			return response.Forbidden(nil)
+		}
+
+		if s.OidcVerifier == nil {
+			return response.Forbidden(nil)
+		}
+
+		if s.OidcVerifier.IsRequest(r) {
```
markylaing: Since only OIDC is allowed, I would reverse this if statement. E.g.:
```golang
if !s.OidcVerifier.IsRequest(r) {
    return response.Forbidden(nil)
}

// Rest of OIDC logic ...
```

internal/api/api.go
```diff
@@ -4,7 +4,85 @@ import (
...
 )
 
-type endpointHandler func(*microclusterState.State, *http.Request) response.Response
+func authHandler(s *state.SiteManagerState) types.EndpointHandler {
+	return func(innerState *microclusterState.State, r *http.Request) response.Response {
+		if r.TLS == nil {
+			return response.Forbidden(nil)
+		}
+
+		if s.OidcVerifier == nil {
```
markylaing: Also since OIDC is an acronym should we rename this to `OIDCVerifier`?

internal/api/api.go
```diff
@@ -4,7 +4,85 @@ import (
...
+				}
+
+				cacheEntries := []identity.CacheEntry{newCacheEntry}
+				if err := s.IdentityCache.ReplaceAll(cacheEntries); err != nil {
+					return response.Forbidden(err)
+				}
+			}
+
+			return response.EmptySyncResponse
+		}
```
markylaing: I just want to check I'm following this logic correctly.

- If there is an error when getting the identity from the cache by their email address, and that error is not a 404 Not Found error, return a 403 Forbidden. Should this be a 500 internal error? Or some other error more reflective of what occurred inside the identity cache?
- Otherwise, if there was a 404 Not Found error, or if the identity we found in the cache is different from the identity that authenticated, then replace all identities in the cache with the current one. What happens when more than one identity authenticates? Also, will propagating this to other members be added in a subsequent PR?

internal/api/api.go
```diff
@@ -4,7 +4,85 @@ import (
...
+			return response.Forbidden(nil)
+		}
+
+		if s.OidcVerifier == nil {
+			return response.Forbidden(nil)
+		}
+
+		if s.OidcVerifier.IsRequest(r) {
+			authCtx := &oidc.AuthCtx{}
+			result, err := s.OidcVerifier.Auth(innerState.Context, authCtx, r)
```
markylaing: Why do you need pass an `oidc.AuthCtx` into the `Auth` function? Can this be set in the request context instead?

internal/api/api.go
```diff
@@ -4,7 +4,85 @@ import (
...
+}
+
+// wrapHandlerWithAuth wraps an endpoint handler with authentication tasks.
+// At this point, the request has already been authenticated by the OIDC verifier.
+// The context here is that if the id token got refreshed, we need to update the cookies based on the SetCookies flag in the request context.
+func wrapHandlerWithAuth(s *state.SiteManagerState, handler types.EndpointHandler) types.EndpointHandler {
+	return func(innerState *microclusterState.State, r *http.Request) response.Response {
+		a := oidc.GetAuthContext(r)
+
+		return response.ManualResponse(func(w http.ResponseWriter) error {
```
markylaing: Would it be possible to pass a `http.ResponseWriter` into `(oidc.OidcVerifier).Auth` so that we can write/delete cookies as necessary? Otherwise this is splitting auth logic into lots of different places and will become hard to keep track of I think.

ui/vite.config.ts
```diff
@@ -9,7 +9,7 @@ export default defineConfig({
     port: 3000,
     proxy: {
       "/ui/assets": {
-        target: "http://0.0.0.0:8414/",
+        target: "https://0.0.0.0:8414/",
```
markylaing: Is this for development purposes? I'm not sure what it has to do with the commit message?

internal/oidc/context.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/google/uuid"
+	"github.com/gorilla/securecookie"
+)
+
+type authCtxKey string
```
markylaing: Generally I'm not sure about moving this logic outside of the authorizer itself.

Why can't the authorizer accept a `http.ResponseWriter` argument? Is this because you would need to wrap it in a `response.Manual` response? I guess that complicates matters...

internal/oidc/oidc.go
```diff
@@ -601,25 +601,13 @@ func NewVerifier(issuer string, clientID string, audience string, clusterCert fu
...
-	proto := getHeaderOrDefault(r, "X-Forwarded-Proto", "")
-	if proto == "" {
-		if r.TLS != nil {
-			proto = "https"
-		} else {
-			proto = "http"
-		}
+	var proto string
+	if r.TLS != nil {
+		proto = "https"
```
markylaing: The protocol should always be `https` shouldn't it? IIRC microcluster doesn't have any non-TLS listeners?

internal/api/auth.go
```diff
@@ -1,13 +1,20 @@
 package api
 
 import (
+	"context"
```
markylaing: There's quite a lot going on in this commit and it's my EOD so I'll come back to this on Monday. Perhaps we can have a call to discuss :)

markylaing (commented):
I've left some general feedback and comments. For the commit which will allow forwarding for the OIDC callback to the correct member, I think it's best if we have a call to discuss :)

.gitignore
mas-who: `state` is the state directory for the cluster so we want to ignore it. The directory `internal/state` holds the package for the site manager state definition which we want to track changes. Thanks for pointing this out though, I see there is a better way to do this. To ignore only the top level `state` directory, we just need to change `state` to `/state/` in `.gitignore`

internal/oidc/oidc.go
mas-who: Thanks, I will have a look across the commits and clean it up :+1: 

Edit: the reason why the imports moved between commits is because I ran the gci linter in later commits than the one that introduced the dependencies, will avoid this moving forward (note to self)

internal/identity/cache.go
mas-who: Yes nice catch, `IdentityType` is not relevant in site maanger :+1: will remove it

internal/oidc/oidc.go
mas-who: I've cleaned up the commits so that the dependencies are correctly imported in the oidc package in the first commit

cmd/lxd-site-mgrd/main.go
mas-who: Nice yeah if it's much simpler to just set the verifier in the `OnStart` hook. I've also squashed the commits since there is no need to create an additional `SetClusterCertificate` method in the oidc package.

internal/oidc/oidc.go
mas-who: No, I've updated the commit message :+1:

internal/oidc/oidc.go
mas-who: No I actually removed this in the second last commit. There is no requirements for the external load balancer now. Previously I needed the `X-FORWARDED` headers in the haproxy due to an incorrect configuration, will explain in the call.

ui/vite.config.ts
mas-who: Nice catch, I've moved it to the next commit which is for adding UI tests

internal/api/api.go
mas-who: Thanks for catching these, I've made the following changes:
 - Return internal error if we end up with an error that is not `404 Not Found` when fetching from the cache
 - Created a `(identity.Cache).ReplaceOrCreate` method which would simply just overwrite a cached identity by its identifier key in the cache or create an entry if the identifier does not exist.
 - I am planning on propagating the identity cache changes at a later stage, especially for when we have oidc configs stored in the database.

internal/oidc/oidc.go
mas-who: My immediate next task is to create the `/config` endpoint and these settings will be stored in the configs table and these constant variables will be removed. For now, I think we can keep it here since it affects CICD workflows and I don't really want to set up everything based on environment variables knowing that later I will update the workflow anyway. wdyt?

internal/api/api.go
mas-who: I have moved the removed the `AuthCtx` argument from all oidc functions.

internal/oidc/oidc.go
mas-who: yes I've made it so that the callback url will always have the `https` protocol

internal/api/auth.go
mas-who: As discussed, I've made the following changes
- Removed source host from the forwarded callback request payload
- Created two functions, `handleCallback` and `handleForwardedCallback`. If the request is a cluster notification, process the request with `handleForwardedCallback`.
- Found a way to use the exact cluster member client using `(microcluster.Microcluster).RemoteClient`

ui/entrypoint
mas-who: Aah yes for development we use the cert for haproxy which the browser will communicate with, we don't really want to replace the cert during development.

internal/api/api.go
mas-who: With the upstream [PR](https://github.com/canonical/microcluster/pull/164) merged, I have adjusted so that `(oidc.OIDCVerifier).Auth` returns a `response.Response` if we need to modify the response.

internal/oidc/context.go
mas-who: This file is now completely removed.

.gitignore
markylaing: I'm not following here. Do you mean `{repo_root}/state` is being used for local development?

We use `TEST_DIR=$(mktemp -d -p "$(pwd)" tmp.XXX)` when testing and ignore all files/folders prefixed with `tmp` in the repo. I think that this makes things a bit clearer but it's up to you.

internal/oidc/oidc.go
edlerd: Let's merge the config endpoints first, so we can rely on them here and not commit the hard coded values.

markylaing: Ok sounds good.

internal/api/api.go
mas-who: remove the identity cache since it's not needed here.

internal/client/auth.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+// CallbackForwardCmd forwards a oidc callback request to the oidc/callback endpoint on.
+func CallbackForwardCmd(ctx context.Context, c *client.Client, sourceRequest *http.Request) (oidc.OidcTokens, error) {
+	queryCtx, cancel := context.WithTimeout(ctx, time.Second*30)
+	defer cancel()
+
+	// Get all the source query params and append it to the target URL
+	sourceQueryParams := sourceRequest.URL.Query()
+	targetURL := api.NewURL().Path("oidc/callback")
```
mas-who: Path("oidc", "callback")

internal/api/auth.go
mas-who: Create issues in microcluster and site manager to detail reasons why we have some weird code. We should also comment a bunch referencing the issues.

internal/api/types/rest.go
```diff
@@ -0,0 +1,14 @@
...
+
+	"github.com/canonical/lxd/lxd/response"
+	"github.com/canonical/microcluster/state"
+)
+
+// AccessHandler represents a function that handles an API endpoint with access control.
+type AccessHandler func(state *state.State, r *http.Request) (bool, response.Response)
+
+// EndpointHandler represents a function that handles an API endpoint.
+type EndpointHandler func(*state.State, *http.Request) response.Response
```
markylaing: @masnax @roosterfish Should we define these types in microcluster?

mas-who:
@markylaing as discussed, I've done the following:
 - cleaned up commits to not have conflicting changes. Also cherry picked some commits and created #35 which should be merged before this PR.
 - removed the hard coded oidc configs. Now the oidc settings will be set by the `/1.0/config` endpoint, and updates will be propagated to all running cluster members. When a new member joins, we will get the oidc settings from dqlite and set it on the new member daemon.
 - removed the identity cache previously defined in this project since it's not needed.
 - added comments highlighting issues relating to not being able to construct the request efficiently for forwarding due to microcluster restrictions. I will create the GH issues once this PR is merged.
 - I decided to keep the e2e test commit in this PR. Since enabling oidc access handler on some endpoints will result in the CI failing if authentication is not included in the e2e tests.

@edlerd I'm going to need some assistance in setting up some additional environment secrets for GA i.e. `OIDC_ISSUER`, `OIDC_CLIENT_ID`, `OIDC_AUDIENCE`

mas-who:
@markylaing #35 is now merged and I have dropped the commits in this PR and rebased.

internal/oidc/oidc.go
```diff
@@ -0,0 +1,655 @@
...
+// verification fails) it will attempt to refresh the ID token.
+func (o *Verifier) authenticateIDToken(ctx context.Context, idToken string, refreshToken string) (result *AuthenticationResult, resp response.Response, e error) {
+	var claims *oidc.IDTokenClaims
+	var err error
+	if idToken != "" {
+		// Try to verify the ID token.
+		claims, err = rp.VerifyIDToken[*oidc.IDTokenClaims](ctx, idToken, o.relyingParty.IDTokenVerifier())
+		if err == nil {
+			return &AuthenticationResult{
+				IdentityType: api.IdentityTypeOIDCClient,
```
markylaing: Did we say we'd remove `IdentityType` for now since only OIDC authentication is possible?

internal/oidc/oidc.go
```diff
@@ -0,0 +1,655 @@
...
+
+		_ = response.SyncResponse(true, OidcTokens{IDToken: tokens.IDToken, RefreshToken: tokens.RefreshToken}).Render(w)
+	}, o.relyingParty)
+
+	handler(w, r)
+}
+
+// WriteTokenToCookies writes the ID and refresh tokens to the response cookies based on a new session ID.
+// The cookies are encrypted using the cluster private key.
+func (o *Verifier) WriteTokenToCookies(w http.ResponseWriter, idToken string, refreshToken string) error {
```
markylaing, line 253: Does this need to be exported now that the auth context is no longer used?

internal/oidc/oidc.go
```diff
@@ -0,0 +1,655 @@
...
+	}
+
+	return nil
+}
+
+// WriteHeaders writes the OIDC configuration as HTTP headers so the client can initatiate the device code flow.
+func (o *Verifier) WriteHeaders(w http.ResponseWriter) error {
+	w.Header().Set("X-LXD-OIDC-issuer", o.issuer)
+	w.Header().Set("X-LXD-OIDC-clientid", o.clientID)
+	w.Header().Set("X-LXD-OIDC-audience", o.audience)
```
markylaing: These are not needed without the CLI. (If you do need OIDC for a CLI at some point, we'll also need to change the headers from `X-LXD-OIDC-*` to something like `X-LXD-site-manager-OIDC-*`).

internal/oidc/oidc.go
```diff
@@ -0,0 +1,655 @@
...
+		}
+	}
+
+	return result, resp, err
+}
+
+// authenticateAccessToken verifies the access token and checks that the configured audience is present the in access
+// token claims. We do not attempt to refresh access tokens as this is performed client side. The access token subject
+// is returned if no error occurs.
+func (o *Verifier) authenticateAccessToken(ctx context.Context, accessToken string) (*AuthenticationResult, error) {
```
markylaing: Also not needed without the CLI. Although you might need this later. At that point we can revisit making these functions available via the LXD shared package.

internal/state/state.go
```diff
@@ -0,0 +1,31 @@
...
+
+	"github.com/canonical/lxd-site-manager/internal/oidc"
+)
+
+// SiteManagerState holds the global state of the site manager.
+type SiteManagerState struct {
+	// OIDCVerifier is the OpenID Connect verifier used for user authentication and validate authentication for protected API endpoints.
+	OIDCVerifier *oidc.Verifier
+	// MicroCluster is the MicroCluster instance.
+	MicroCluster *microcluster.MicroCluster
```
markylaing, line 16: This seems very circular. It's not a blocker, but could you create an issue in microcluster to address this please?

The microcluster state should contain everything you need in a handler. IIRC this gives you access to a method that can make requests to a single member without making a client for all members in the cluster? If so, I think this should be defined on state.

markylaing (dismissed):
I think we can move forward with this for now.

I'm not a fan of the additional endpoint for the OIDC callback (nor adding the cookies to the request body), but as this is circumventing a limitation of microcluster it'll have to do for now. There may be a better way this more generally (see https://github.com/canonical/lxd/issues/13644#issuecomment-2213288260).

When we have a solution to https://github.com/canonical/lxd/issues/13644 we will also move the `oidc` package into `lxd/shared` and make it more of a library. At that point we'll need to revisit this and make changes.

internal/oidc/oidc.go
mas-who: Aaah yes, thanks will remove.

internal/oidc/oidc.go
mas-who: We need `WriteTokenToCookies` for handling the tokens returned in the response of a forwarded idp callback request. This is due to the microcluster restriction that we do not receive the full response object from a client query function, this means I can't extract the session cookies from the response headers and forward them to the browser. Instead, I have to return the oidc tokens, generate the cookies and write them to the response for the browser.

internal/oidc/oidc.go
mas-who: I will remove everything from the oidc package that's not needed in site manager for now :+1:

internal/oidc/oidc.go
mas-who: removing this method as well as usage :+1:

internal/state/state.go
mas-who: Yes agreed, I will open an issue for this.

FYI, there is another way to forward a request to another cluster member in microcluster. For a `rest.EndpointAction`, we can use the `ProxyTarget` flag to forward a request to a member. However, this requires the `target` query param to be set in the idp callback url, which I think isn't appropriate for the oidc flow.

mas-who: created an [issue](https://github.com/canonical/microcluster/issues/171) in microcluster for this.

mas-who:
> I think we can move forward with this for now.
> 
> I'm not a fan of the additional endpoint for the OIDC callback (nor adding the cookies to the request body), but as this is circumventing a limitation of microcluster it'll have to do for now. There may be a better way this more generally (see [canonical/lxd#13644 (comment)](https://github.com/canonical/lxd/issues/13644#issuecomment-2213288260)).
> 
> When we have a solution to [canonical/lxd#13644](https://github.com/canonical/lxd/issues/13644) we will also move the `oidc` package into `lxd/shared` and make it more of a library. At that point we'll need to revisit this and make changes.

@markylaing I've created the following issues to keep track of the above:
#39 - Improve approach to handling oidc login flow across multiple cluster members
#40 - use LXD shared oidc library when ready
#41 - remove `POST /oidc/callback endpoint` (created upstream microcluster [issue](https://github.com/canonical/microcluster/issues/170))

markylaing:
> > I think we can move forward with this for now.
> > I'm not a fan of the additional endpoint for the OIDC callback (nor adding the cookies to the request body), but as this is circumventing a limitation of microcluster it'll have to do for now. There may be a better way this more generally (see [canonical/lxd#13644 (comment)](https://github.com/canonical/lxd/issues/13644#issuecomment-2213288260)).
> > When we have a solution to [canonical/lxd#13644](https://github.com/canonical/lxd/issues/13644) we will also move the `oidc` package into `lxd/shared` and make it more of a library. At that point we'll need to revisit this and make changes.
> 
> @markylaing I've created the following issues to keep track of the above: #39 - Improve approach to handling oidc login flow across multiple cluster members #40 - use LXD shared oidc library when ready #41 - remove `POST /oidc/callback endpoint` (created upstream microcluster [issue](https://github.com/canonical/microcluster/issues/170))

Cool. I've also created an issue to define the handler types in microcluster so that you don't need to do that here.

.github/workflows/pr.yaml
```diff
@@ -125,9 +126,31 @@ jobs:
           lxd-site-mgr --state-dir ./state/dir1/ waitready
           lxd-site-mgr --state-dir ./state/dir1/ init "member1" 0.0.0.0:9001 --bootstrap
 
+      - name: Add dummy data to site manager
+        run: |
+          set -eux
+          sudo chmod 0777 ../lxd-site-manager
+          scripts/generate_sites.sh
+
```
edlerd:
```suggestion
```
This shouldn't be part of this PR.

entrypoint
```diff
@@ -2,4 +2,16 @@
...
+# initialise the site manager cluster
+./lxd-site-mgr --state-dir ./state/dir1 waitready
+./lxd-site-mgr --state-dir ./state/dir1 init "member" 0.0.0.0:443 --bootstrap
+
+# initialise data for site manager
+./scripts/generate_sites.sh
+./lxd-site-mgr --state-dir ./state/dir1 config set oidc.issuer=$OIDC_ISSUER oidc.client.id=$OIDC_CLIENT_ID  oidc.audience=$OIDC_AUDIENCE
+
+# keep container running
+tail -f /dev/null
```
edlerd:
```suggestion
tail -f /dev/null

```

internal/api/auth.go
```diff
@@ -0,0 +1,162 @@
...
+		Get:  rest.EndpointAction{Handler: oidcLogout(s), AllowUntrusted: true},
+	}
+}
+
+func oidcLogin(s *state.SiteManagerState) types.EndpointHandler {
+	return func(innerState *microclusterState.State, r *http.Request) response.Response {
+		loginHandler := func(w http.ResponseWriter) error {
+			// Encode state with uuid and member address.
+			stateToken := oidc.StateToken{
+				MemberAddress: innerState.Address().URL.String(),
```
edlerd: This will make the internal address publicly visible. I am not sure if this is a good idea. Can we make the whole forwarding of requests a separate PR to have only the login in here?

ui/src/App.tsx
```diff
@@ -31,6 +33,10 @@ const App: FC = () => {
     return null;
   }
 
+  useEffect(() => {
+    setIsAuthenticated(!isAuthError);
+  }, [isAuthError]);
+
```
edlerd: The loading experience of the login page is blinking in and out the navigation for some time. Can we hide the navigation until the auth is loaded to avoid this flickering? Or always display the navigation, also logged out.

ui/src/context/auth.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+interface ContextProps {
+  isAuthenticated: boolean;
+  setIsAuthenticated: (value: boolean) => void;
+}
+
+const initialState: ContextProps = {
+  isAuthenticated: false,
+  setIsAuthenticated: () => {},
+};
```
edlerd: This should be better abstracted and decoupled.

We should not expose `setIsAuthenticated` outside this hook. Figure the isAuthenticated state inside this hook with a local query. Invalidate the cache for that query on logout to trigger a refresh, but do not directly couple it to this hook.

internal/oidc/oidc.go
```diff
@@ -0,0 +1,582 @@
...
+
+const (
+	defaultConfigExpiryInterval = 5 * time.Minute
+)
+
+// Verifier holds all information needed to verify an access token offline.
+type Verifier struct {
+	accessTokenVerifier *op.AccessTokenVerifier
+	relyingParty        rp.RelyingParty
+	identityCache       *identity.Cache
```
edlerd:
```suggestion
```
Seems to be unused?

internal/oidc/oidc.go
```diff
@@ -0,0 +1,582 @@
...
+	cookieNameSessionID = "session_id"
+)
+
+const (
+	defaultConfigExpiryInterval = 5 * time.Minute
+)
+
+// Verifier holds all information needed to verify an access token offline.
+type Verifier struct {
+	accessTokenVerifier *op.AccessTokenVerifier
```
edlerd:
```suggestion
```
Unused?

internal/oidc/oidc.go
```diff
@@ -0,0 +1,582 @@
...
+func (e AuthError) Unwrap() error {
+	return e.Err
+}
+
+// StateToken is used to encode the state of the OIDC client in a URL which is used to prevent CSRF attacks (https://datatracker.ietf.org/doc/html/rfc6749#section-10.12).
+// MemberAddress is the address of the member that initiated the login flow. In a distributed system, we can use this address to ensure that the login flow is completed on the same member.
+// ID is a unique identifier for the state token.
+type StateToken struct {
+	MemberAddress string
+	ID            string
```
edlerd:
```suggestion
```
unused?

internal/oidc/oidc.go
```diff
@@ -0,0 +1,582 @@
...
+	var result *AuthenticationResult
+	if idToken != "" || refreshToken != "" {
+		// When authenticating via the UI, we expect that there will be ID and refresh tokens present in the request cookies.
+		result, resp, err = o.authenticateIDToken(ctx, idToken, refreshToken)
+		if err != nil {
+			return nil, resp, err
+		}
+	}
+
+	return result, resp, err
```
edlerd: If we reach this, we will return an empty result and response. I think we should throw an error here instead.

internal/oidc/oidc.go
```diff
@@ -0,0 +1,582 @@
...
+		err := o.WriteTokenToCookies(w, tokens.IDToken, tokens.RefreshToken)
+
+		if err != nil {
+			_ = response.ErrorResponse(http.StatusInternalServerError, err.Error()).Render(w)
+		}
+
+		if redirect {
+			// Send to the UI.
+			// NOTE: Once the UI does the redirection on its own, we may be able to use the referer here instead.
+			http.Redirect(w, r, "/ui", http.StatusMovedPermanently)
```
edlerd: It would be nice if the ui sets a `next` param on starting the flow. We should store it in the state token and unpack it here to forward to the next url. That way in case a user accesses an internal url, they will land on that url after login.

internal/oidc/oidc.go
```diff
@@ -0,0 +1,582 @@
...
+	o.accessTokenVerifier = op.NewAccessTokenVerifier(o.issuer, keySet)
+	return nil
+}
+
+// getCookies gets the sessionID, identity and refresh tokens from the request cookies and decrypts them.
+func (o *Verifier) getCookies(r *http.Request) (sessionIDPtr *uuid.UUID, idToken string, refreshToken string, err error) {
+	sessionIDCookie, err := r.Cookie(cookieNameSessionID)
+	if err != nil && !errors.Is(err, http.ErrNoCookie) {
+		return nil, "", "", fmt.Errorf("Failed to get session ID cookie from request: %w", err)
+	} else if sessionIDCookie == nil {
```
edlerd:
```suggestion
	}
	if sessionIDCookie == nil {
```

edlerd (changes requested):
Best to factor out the forwarding logic to keep this one concise.

ui/src/App.tsx
mas-who: quick fix is to set `const isAuthError = isLoading || error?.message === "not authorized";`. In general I think the auth logic in the UI will need some refactoring once we have the `/1.0` endpoint.

ui/src/context/auth.tsx
mas-who: Thanks I've moved the query from `App.tsx` to `auth.tsx` and also added a `FIXME` comment to say the query should reach out to `/1.0` once the endpoint is ready. I've also moved the `logout` helper from lxd-ui. We don't need to do a query invalidation since we are setting `window.location.href` when logging out which triggers a page reload.

internal/oidc/oidc.go
mas-who: I kept it previously because the `identityCache` will be needed if we perform access token verification which is used for device grant flow in lxd. I think it's easy enough to add it back later if we need it, so removing this now.

internal/oidc/oidc.go
mas-who: related to the earlier comment about dependency for device grant flow. Will remove for now as it's not needed.

internal/oidc/oidc.go
mas-who: The `StateToken` gets base64 encoded and will be passed to the idp as the `state` query param in the callback url. The idea here is that the id field in the token would uniquely identify the current login flow otherwise multiple users triggering the login flow would have the exact same `state`. The query param is used by zitadel internally to verify the login session.

That being said, since we are going to strip out the logic for forwarding idp callback request. `StateToken` will no longer be relevant in the context of this PR. So going to resolve this convo for now.

internal/oidc/oidc.go
mas-who: sorry I'm not sure if I follow, if we reach this point we would have valid result and response. Although we should return `nil` for error.

internal/api/auth.go
mas-who: as discussed, will strip out the logic for forwarding request into a separate PR for future reference.

internal/oidc/oidc.go
edlerd: I was thinking we can reach here in case id-token and refresh-token are both empty.

mas-who: aah I see, this is taken care of in `authenticateIDToken` since if both tokens are empty, then it would result in a token refresh error.

internal/oidc/oidc.go
mas-who: Yeah sounds good. Will implement this logic once I have moved out the forwarding request logic shortly.

mas-who: @edlerd implemented this. It would be easy to check the changes in the last commit.

internal/oidc/oidc.go
edlerd: Problem is we won't enter that branch as it is guarded by `if idToken != "" || refreshToken != "" {`

edlerd: Discussed in person, there is a check upstream in `!s.OIDCVerifier.IsRequest(r)` guarding from this case.

ui/src/context/auth.tsx
```diff
@@ -0,0 +1,46 @@
...
+    queryFn: fetchClusters,
+    retry: false,
+  });
+
+  const isAuthError = isLoading || error?.message === "not authorized";
+
+  return (
+    <AuthContext.Provider
+      value={{
+        isAuthenticated: !isAuthError,
```
edlerd:
```suggestion
  const isAuthenticated = !isLoading && error?.message === "not authorized";

  return (
    <AuthContext.Provider
      value={{
        isAuthenticated,
```
Seems easier to reason about?

edlerd (changes requested):

edlerd (approved):
LGTM :)

ui/src/context/auth.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+export const AuthProvider: FC<ProviderProps> = ({ children }) => {
+  // FIXME: this should query /1.0 when the endpoint is ready to check if the user is authenticated
+  const { error, isLoading } = useQuery({
+    queryKey: [queryKeys.clusters],
+    queryFn: fetchClusters,
+    retry: false,
+  });
+
+  const isAuthenticated = !isLoading && error?.message === "not authorized";
```
edlerd:
```suggestion
  const isAuthenticated = !isLoading && error?.message !== "not authorized";
```

edlerd (changes requested):

edlerd (approved):
Good work, thanks for adding login :)

PR #12, opened 2024-06-05
Add make clean target

edlerd:
Add a make target to clean all temp files

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,6 +34,13 @@ endif
 	golangci-lint run --timeout 5m
 	revive -set_exit_status ./...
 
+# Cleanup temp and build artifacts.
+.PHONY: clean
+clean:
+	rm -rf api/static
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -34,6 +34,13 @@ endif
 	golangci-lint run --timeout 5m
 	revive -set_exit_status ./...
 
+# Cleanup temp and build artifacts.
+.PHONY: clean
+clean:
+	rm -rf api/static
```
mas-who, line 40: I've been using `ps aux | grep "lxd-site-mgrd" | awk '{print $2}' | xargs kill -9` to stop the daemon process, should we also add it to this target?

edlerd: I think make should focus on building, not managing the processes.

mas-who (approved):

PR #13, opened 2024-06-05
Setup CI for backend and UI tests [WD-11367]

mas-who:
## Done
 - Added static code analysis for backend
 - Added unit test for backend
 - Added e2e test for UI (using compiled go binary)

## Notes
 - Left out playwright publishing workflow for now since github pages does not work for private repos without a paid plan

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -44,9 +44,7 @@ jobs:
           set -eux
           sudo add-apt-repository ppa:dqlite/dev -y --no-update
           sudo apt-get update
-          sudo apt-get install --no-install-recommends -y \
-            libdqlite-dev \
-            shellcheck
+          sudo apt-get install --no-install-recommends -y libdqlite-dev
           cd ui && yarn install --immutable
 
       - name: Build LXD Site Manager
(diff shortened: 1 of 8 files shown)


ui/blob-report/report.zip
edlerd, line 1: This should be ignored and not committed, right?

.github/workflows/pr.yaml
```diff
@@ -21,34 +24,145 @@ jobs:
...
+      - name: Make LXD Site Manager Binaries
+        run: make compile
+
+      - name: Run site manager daemon
+        run: lxd-site-mgrd --state-dir ./state/dir1/ &
+
+      - name: Initialise site manager
+        run: |
+          lxd-site-mgr --state-dir ./state/dir1/ waitready
+          lxd-site-mgr --state-dir ./state/dir1/ init "member1" 0.0.0.0:9001 --bootstrap
```
edlerd, line 126: Shall we rely on `make` and then `yarn backend-run` as well as `yarn backend-init` here? Then we would cover checking our dev setup works as described on the side.

ui/blob-report/report.zip
mas-who: oops that slipped through the cracks

.github/workflows/pr.yaml
mas-who: hmm I was hoping to use the CI to test the compiled binaries, so it's almost like testing the snaps, later on we could also potentially compile the go binaries across different os + arch and run those tests in a matrix, but maybe that's too much and we can instead cross check our dev scripts?

edlerd: Right, probably better to run tests closer to the real world than with the dev builds.

mas-who: So keeping it as is?

edlerd: Yeah.

edlerd (approved):
LGTM

PR #15, opened 2024-06-07
Setup management listener using `extensionServers` [WD-11937]

mas-who:
## Done
 - removed the old way of registering endpoints using `extendedEnpoints`
 - added `server.go` which has `Servers` slice. The network listeners are defined in this slice.
 - added all existing endpoints to the site management listener which will be grouped with the core listener
 - server static assets for the frontend via `/ui` instead of `/1.0/ui`
 
## Note
 - Originally enabled by this [PR](https://github.com/canonical/microcluster/pull/130).
 - This PR is dependent on the upstream microcluster [PR](https://github.com/canonical/microcluster/pull/142) getting merged.
 - relating to this [issue](https://github.com/canonical/microcluster/issues/139), it seems like currently there is no way to pass the server structs to microcluster during cluster initialisation. This is not a problem for the site management listener since its endpoints resides with the core listener. However, the status listener will need to have its address passed during initialisation.


go.mod
```diff
@@ -2,9 +2,12 @@ module github.com/canonical/lxd-site-manager
 
 go 1.22.3
 
+// TODO: remove after testing
+replace github.com/canonical/microcluster => ../microcluster
```
mas-who: We can use this to test microcluster dependency locally. This will be removed after upstream PR gets merged.

internal/api/server.go
```diff
@@ -0,0 +1,30 @@
...
+	// site management listener (same network listener as core)
+	{
+		CoreAPI: true,
+		Resources: []rest.Resources{
+			{
+				Path: "",
+				Endpoints: []rest.Endpoint{
+					uiCmd,
+					uiAssetsCmd,
+					uiImgCmd,
```
edlerd: I think we should add a handler for the `/` route, to redirect to `/ui` from the go side. We do have it in the App.tsx for the dotrun/ dev server side. But in production we'd need it as well.

mas-who: That makes sense, although currently it's not possible to do that with microcluster because there is a handler for `/` that's registered before the user defined endpoints. See [here](https://github.com/canonical/microcluster/blob/c2ff70b3b3e39b601727ebda4a4a43c9c0f0de57/internal/daemon/daemon.go#L321).

I see lxd has a pretty reasonable [approach](https://github.com/canonical/lxd/blob/45f8fa530224da96acd25ac5e450f8d851be2022/lxd/api.go#L166) for the redirect, I'm gonna see if I can get that into the upstream PR.

Edit: might be tricky to add it to microcluster exactly how lxd does it because we can't assume in microcluster that there will be a `/ui` path.

mas-who: So I got a [workaround](https://github.com/canonical/microcluster/pull/142/commits/817979f5a17227cdbed5a40dc19885cd8eeb0aef) in the same upstream PR. Idea is just to change the ordering of route definition. If a handler for `/` is defined in a downstream app, then that handler will be used over the one defined in microcluster.

Makefile
```diff
@@ -13,6 +13,7 @@ compile-binary:
 .PHONY: build
 build:
 	cd ui && yarn install && yarn build
+	rm -rf internal/api/static &>/dev/null
```
edlerd, line 16: I think this is still needed, otherwise we might leave orphaned files in there from previous runs.

internal/api/server.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"github.com/canonical/microcluster/rest"
+
+	"github.com/canonical/lxd-site-manager/internal/api/types"
+)
+
+// Servers contains all the network listeners for site manager.
+var Servers = []rest.Server{
+	// site management listener (same network listener as core)
+	{
```
edlerd: Should we assign this struct to a variable named `SiteManagementListener` and use it here in favour of the comment? I think that might enhance structure and readability.

internal/api/types/server.go
```diff
@@ -0,0 +1,10 @@
+package types
+
+import "github.com/canonical/microcluster/rest/types"
+
+const (
+	// CoreEndpointPrefix is the prefix for the core API.
+	CoreEndpointPrefix types.EndpointPrefix = "1.0"
```
edlerd: Maybe better `ApiVersionPrefix`?

internal/api/types/server.go
```diff
@@ -0,0 +1,10 @@
...
+	// PublicEndpointPrefix is the prefix for the public API.
+	PublicEndpointPrefix types.EndpointPrefix = ""
```
edlerd: This is not really public, is it? Maybe better `NoPrefix` or `RootPrefix`?

ui/vite.config.ts
```diff
@@ -7,21 +7,14 @@ export default defineConfig({
   base: process.env.VITE_BASE_URL || "/",
   server: {
     port: 3000,
-    proxy: {
-      "/ui/assets": {
-        target: "https://localhost:8414/",
-        rewrite: (path) => path.replace(/^\/ui/, ""),
-        secure: false,
-      },
-    },
```
edlerd: Why remove this? I think this is still needed.

edlerd (changes requested):

Makefile
mas-who: The intention here is to cleanup the static directory so it's a clean copy from `/ui/static` to `/internal/api/static` every time?

edlerd: Sorry, I was confused and thought you removed this line, but actually are adding it.

internal/api/server.go
mas-who: good idea! :)

internal/api/types/server.go
mas-who: Agreed, currently it's all public endpoints e.g. /ui, /oidc etc. But I think this is not guaranteed. Let's use `NoPrefix` since it's explicit

ui/vite.config.ts
mas-who: It was causing proxy errors in the dotrun log output. The fix is actually to change the `target`, I've made the update

edlerd (approved):
Code and QA LGTM

PR #20, opened 2024-06-19
[WD-12166] feat: Added Login/Logout button (ui) for LXD-SM.

Kxiru:
Tasks:
* Added a route to the Login Page
* Added a Login page using LoginPageLayout
* Added a Login button to the login page and a logout button to the SiteList page.

Login Page
image

Logout Button
image

### DIFF
--- a/ui/src/App.tsx
+++ b/ui/src/App.tsx
@@ -2,13 +2,15 @@ import { FC, lazy, Suspense } from "react";
 import { Navigate, Route, Routes } from "react-router-dom";
 
 const SiteList = lazy(() => import("pages/sites/SiteList"));
+const Login = lazy(() => import("pages/Login"));
 
 const App: FC = () => {
   return (
     <Suspense fallback={<div>Loading</div>}>
       <Routes>
         <Route path="/" element={<Navigate to="/ui" replace={true} />} />
-        <Route path="/ui" element={<SiteList />} />
+        <Route path="/ui" element={<Login />} />
+        <Route path="/ui/sites" element={<SiteList />} />
       </Routes>
     </Suspense>
   );
--- a/ui/src/pages/Login.tsx
+++ b/ui/src/pages/Login.tsx
@@ -1,0 +1,15 @@
+import { FC } from "react";
+import { Button, LoginPageLayout } from "@canonical/react-components";
+import { useNavigate } from "react-router-dom";
+
+const Login: FC = () => {
+  const navigate = useNavigate();
+
+  return (
+    <LoginPageLayout title="Login to LXD site manager">
+      <Button onClick={() => navigate("/oidc/login")}>Login</Button>
+    </LoginPageLayout>
+  );
+};
+
+export default Login;
--- a/ui/src/pages/sites/SiteList.tsx
+++ b/ui/src/pages/sites/SiteList.tsx
@@ -2,9 +2,11 @@ import { FC } from "react";
 import { queryKeys } from "util/queryKeys";
 import { fetchSites } from "api/sites";
 import { useQuery } from "@tanstack/react-query";
-import { MainTable } from "@canonical/react-components";
+import { Button, MainTable } from "@canonical/react-components";
+import { useNavigate } from "react-router-dom";
 
 const SiteList: FC = () => {
+  const navigate = useNavigate();
   const { data: sites = [] } = useQuery({
     queryKey: [queryKeys.sites],
     queryFn: fetchSites,
@@ -29,6 +31,8 @@ const SiteList: FC = () => {
           };
         })}
       />
+
+      <Button onClick={() => navigate("/oidc/logout")}>Logout</Button>
     </div>
   );
 };
--- a/ui/tests/site-list.spec.ts
+++ b/ui/tests/site-list.spec.ts
@@ -1,6 +1,6 @@
 import { expect, test } from "@playwright/test";
 
 test("site list", async ({ page }) => {
-  await page.goto("/ui");
+  await page.goto("/ui/sites");
   expect(await page.title()).toBe("LXD Site Manager");
 });


ui/src/App.tsx
```diff
@@ -2,12 +2,14 @@ import { FC, lazy, Suspense } from "react";
...
 
 const SiteList = lazy(() => import("pages/sites/SiteList"));
+const Login = lazy(() => import("pages/Login"));
 
 const App: FC = () => {
   return (
     <Suspense fallback={<div>Loading</div>}>
       <Routes>
         <Route path="/" element={<Navigate to="/1.0/ui" replace={true} />} />
+        <Route path="/ui/login" element={<Login />} />
```
edlerd:
```suggestion
        <Route path="/ui" element={<Login />} />
```
I think this should be the default route, the one below can be /ui/sites

ui/src/pages/Login.tsx
```diff
@@ -0,0 +1,15 @@
+import { FC } from "react";
+import { Button, LoginPageLayout } from "@canonical/react-components";
+import { Link } from "react-router-dom";
+
+const Login: FC = () => {
+  return (
+    <LoginPageLayout title="Login">
+      <Button>
+        <Link to={"/1.0/oidc/login"}>Login (OIDC)</Link>
```
edlerd: There shouldn't be a link inside a button, you can set the href param on the button instead. The caption should be "Login", no need to point out we are using oidc to the user.

ui/src/pages/Login.tsx
```diff
@@ -0,0 +1,15 @@
+import { FC } from "react";
+import { Button, LoginPageLayout } from "@canonical/react-components";
+import { Link } from "react-router-dom";
+
+const Login: FC = () => {
+  return (
+    <LoginPageLayout title="Login">
```
edlerd:
```suggestion
    <LoginPageLayout title="Login to LXD site manager">
```
Maybe better?

ui/src/pages/sites/SiteList.tsx
```diff
@@ -29,6 +30,10 @@ const SiteList: FC = () => {
           };
         })}
       />
+
+      <Button>
+        <Link to={"/1.0/oidc/logout"}>Login (OIDC)</Link>
```
edlerd:
```suggestion
        <Link to="/1.0/oidc/logout">Logout</Link>
```

edlerd (changes requested):

ui/src/pages/Login.tsx
mas-who: Please also change the link url to `/oidc/login` instead, we planning on getting rid of the `/1.0` prefix for login endpoints

ui/src/pages/sites/SiteList.tsx
mas-who:
```suggestion
        <Link to={"/oidc/logout"}>Login (OIDC)</Link>
```

edlerd (changes requested):
Code and QA LGTM, can you ensure the commit message is reflecting the changes?

For the future, you don't need to use the project name in the branch name, as the branches are namespaced per project.

edlerd:
You can use `git commit --amend -a -S -s` to change the commit message.

edlerd (approved):
All good now, thanks for adding it and providing feedback to improve the setup / readme when starting.

Just one issue with the failing e2e test, I think we need to update the test with the new home page title.

Kxiru:
> All good now, thanks for adding it and providing feedback to improve the setup / readme when starting.
> 
> Just one issue with the failing e2e test, I think we need to update the test with the new home page title.

All fixed!

Kxiru:
Although I believe the Chromium e2e test is failing incorrectly. As the information is it pulling is outdated. The title of the Login.tsx page has definitely been changed, and it is aligned with the correct navigation url.

edlerd:
> Although I believe the Chromium e2e test is failing incorrectly. As the information is it pulling is outdated. The title of the Login.tsx page has definitely been changed, and it is aligned with the correct navigation url.

This is probably a timing issue. Because the login page is changing the title after the initial loading. You could also change the test to rely on the /ui/sites page to have the previous title. After all that is the site list spec test.

Kxiru:
@edlerd After the alterations we made yesterday, the tests are still failing. How would you like me to approach this?

edlerd:
> @edlerd After the alterations we made yesterday, the tests are still failing. How would you like me to approach this?

I think we are missing the latest changes from the main branch and that's why the request to /ui/sites is 404. Can you rebase on main and force-push the update? That should fix it.

Kxiru:
I have already rebased and forced pushed here :(. Which changes are missing?

edlerd:
> I have already rebased and forced pushed here :(. Which changes are missing?

You are right indeed. I think this is because /ui/sites is a new path that needs to be registered in go explicitly. @mas-who can we make it like in LXD where all paths under /ui/* serve the html/js frontend files? If not, can you help to register the new path in go?

mas-who:
> > I have already rebased and forced pushed here :(. Which changes are missing?
> 
> You are right indeed. I think this is because /ui/sites is a new path that needs to be registered in go explicitly. @mas-who can we make it like in LXD where all paths under /ui/* serve the html/js frontend files? If not, can you help to register the new path in go?

I've created a PR #22 to resolve the issue here. Left some comments on that PR for context and I think there is still a bit of work to be done to streamline how the ui should be served.

Kxiru:
@edlerd , @mas-who , thanks for making the fixing PR! What does this mean for this PR though? :)

mas-who:
> @edlerd , @mas-who , thanks for making the fixing PR! What does this mean for this PR though? :)

The other PR will need to be merged first, then the failing CI here should pass since I've tested the changes with your PR.

Kxiru:
> > @edlerd , @mas-who , thanks for making the fixing PR! What does this mean for this PR though? :)
> 
> The other PR will need to be merged first, then the failing CI here should pass since I've tested the changes with your PR.

As vigilant as ever. Thanks a bunch!

mas-who:
@Kxiru the #22 just got merged. Can you rebase the changes and push again

PR #22, opened 2024-06-22
serve index for component paths

mas-who:
## Done
- Since the ui implements lazy loading, this means components will be their own static JS files in the `static` directory. When navigating to a specific ui path, a request is sent to the same backend server path e.g. `/ui/sites` for the component. Since this route is not a valid file path in the `static` directory, we need to open `index.html` instead. This will load the built `index.js` file which then will load the correct `[Component].js` file. This is fixed by setting up a `uiHTTPDir` struct that implements the `http.FileSystem` interface where we define a custom `Open` method that will open the `index.html` file if a file path is not found.
- Created a slice `uiServeRoutes` which is used to generate all the endpoints + handlers for the ui routes.
- Added `ui/sites` route

## Context
 - Microcluster currently does not expose the `mux` object on which we can use the `mux.PathPrefix` method to define generic path prefix matching. Each endpoint must be defined individually which is not ideal, but an upstream update for this may be a larger task than it seems since these endpoints will need to be defined at the top level of the server and that's a paradigm change. For now, we will need to add each frontend route to the `uiServeRoutes` slice which should solve the issue in #20.
 - The above challenge could be improved by some code generation magic. If we can somehow export all the routes into a json file that will live in the `static` directory, then we can read that list out in the go server and generate all the endpoints for those ui routes. At least we can have a single source of truth that way. I wonder if we can customise the vite build step to achieve that somehow?

### DIFF
--- a/internal/api/server.go
+++ b/internal/api/server.go
@@ -11,12 +11,12 @@ var siteManagementListener = rest.Server{
 	Resources: []rest.Resources{
 		{
 			PathPrefix: types.NoPrefix,
-			Endpoints: []rest.Endpoint{
-				uiRootCmd,
-				uiCmd,
-				uiAssetsCmd,
-				uiImgCmd,
-			},
+			Endpoints: append(
+				[]rest.Endpoint{
+					uiRootCmd,
+				},
+				generateUIEndpoints()...,
+			),
 		},
 		{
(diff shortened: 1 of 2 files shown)


edlerd (approved):
LGTM :+1: 

Thanks for adding this so quickly!

PR #23, opened 2024-06-25
fix: ui auth should redirect to login on auth errors 

edlerd:
# Done
- move login page to dedicated route
- redirect to login page on auth errors
- redirect to site list after user authenticated.
- add a 404 page.


mas-who:
LGTM :+1: This will help with UI testing for #11

mas-who (approved):

PR #24, opened 2024-06-25
Setup DB schema [WD-12171]

mas-who:
## Done 
 - setup database schema as per the site manager [spec](https://docs.google.com/document/d/1q0mlkekLO7iTVnktHFJU2VTGAE9ojtR7av1N3QHZGo8/edit)
 - used lxd-generate to setup all the database helper functions
 - update example `/sites` api usage of database methods
 - adjusted ui to reflect new sites schema

## Context
 To setup `lxd-generate`, I did it with the below steps:
 1. Clone lxd repo
 2. Inside `/lxd/db/generate`, run `go build` to compile the binary `generate`
 3. move the binary to a folder somewhere and then add it to `PATH`
 4. Also had to install goimports `go install golang.org/x/tools/cmd/goimports@latest`
 5. Once the above steps are done, then running `make update-schema` in the root project folder will generate all the db helper functions inside `/internal/database` folder


internal/database/site_member_statuses.go
```diff
@@ -0,0 +1,47 @@
...
+//go:generate mapper method -i -d github.com/canonical/microcluster/cluster -e site_member GetMany
+//go:generate mapper method -i -d github.com/canonical/microcluster/cluster -e site_member GetOne
+//go:generate mapper method -i -d github.com/canonical/microcluster/cluster -e site_member ID
+//go:generate mapper method -i -d github.com/canonical/microcluster/cluster -e site_member Exists
+//go:generate mapper method -i -d github.com/canonical/microcluster/cluster -e site_member Create
+//go:generate mapper method -i -d github.com/canonical/microcluster/cluster -e site_member DeleteOne-by-MemberName-and-CoreSiteID
+//go:generate mapper method -i -d github.com/canonical/microcluster/cluster -e site_member Update
+
+// SiteMember represents all site member level data.
+type SiteMember struct {
```
edlerd: Should the name match the filename?

internal/api/sites.go
```diff
@@ -28,26 +29,52 @@ var siteCmd = rest.Endpoint{
...
+				Name:             coreSite.Name,
+				SiteCertificate:  coreSite.SiteCertificate,
+				Status:           siteDetail.Status,
+				InstanceCount:    siteDetail.InstanceCount,
+				InstanceStatuses: siteDetail.InstanceStatuses,
+				JoinedAt:         siteDetail.JoinedAt,
+				CreatedAt:        coreSite.CreatedAt,
+				UpdatedAt:        siteDetail.UpdatedAt,
+			})
+		}
```
edlerd: I think we should change this to use a join in the database.
This should also include the member statuses.
Same for the single get below.

internal/database/site_details.go
```diff
@@ -0,0 +1,39 @@
...
+
+// SiteDetail represents all site level data.
+type SiteDetail struct {
+	ID               int
+	CoreSiteID       int `db:"primary=true"`
+	JoinedAt         time.Time
+	Status           string `db:"primary=true"`
+	InstanceCount    int
+	InstanceStatuses string
+	UpdatedAt        time.Time
```
edlerd: I think we should keep the indexed fields on top and the time structs on the bottom.
Same for other entities.

internal/api/types/site.go
```diff
@@ -1,9 +1,16 @@
...
-	Addresses []string `json:"addresses"`
-	Status    string   `json:"status"`
+	Name             string    `json:"name"`
+	SiteCertificate  string    `json:"site_certificate"`
+	JoinedAt         time.Time `json:"joined_at"`
+	Status           string    `json:"status"`
+	InstanceCount    int       `json:"instance_count"`
+	InstanceStatuses string    `json:"instance_statuses"`
+	CreatedAt        time.Time `json:"created_at"`
+	UpdatedAt        time.Time `json:"updated_at"`
```
edlerd: We should add the member statuses here as well.

cmd/lxd-site-mgrd/main.go
```diff
@@ -80,16 +80,15 @@ func (c *cmdDaemon) run(cmd *cobra.Command, args []string) error {
...
-`)
+				testStmt := `
+					INSERT INTO core_sites (name, site_certificate) VALUES ('site1', 'a');
+					INSERT INTO core_sites (name, site_certificate) VALUES ('site2', 'as');
+					INSERT INTO core_sites (name, site_certificate) VALUES ('site3', 'asd');
+					INSERT INTO site_details (core_site_id, status, instance_statuses) VALUES (1, 'PENDING_APPROVAL', '{}');
+					INSERT INTO site_details (core_site_id, status, instance_statuses) VALUES (2, 'ACTIVE', '{}');
+					INSERT INTO site_details (core_site_id, status, instance_statuses) VALUES (3, 'ACTIVE', '{}');
+				`
+				_, err := tx.ExecContext(ctx, testStmt)
```
edlerd: Shall we remove this here and extract it in a dedicated script, that can be triggered to add the dummy values?

edlerd (commented):
This looks good, some observations below.

internal/database/site_member_statuses.go
mas-who: These structs should represent a single row entry in the table, the file name represents the table names

internal/api/sites.go
mas-who: Yes agreed, I couldn't quite find an example in `lxd-generate` that supports creating helper functions that require multiple table joins (single table join is supported). Also double checked this with @roosterfish , seems like for multiple table joins in lxd, the db functions are manually created. This is relevant here since we need to join `core_sites`, `site_details` and `site_member_statuses` tables to complete the definition of a `Site` that gets returned by the API.

So I think I'm going to do some re-shuffling in the `database` folder. We will have `lxd-generate` generate all the single table helper functions, then in a separate package folder `relations` (naming suggestions welcomed), I will have composite db functions defined. wdyt?

internal/database/site_member_statuses.go
edlerd: Shouldn't this struct be called SiteMemberStatus?

internal/api/types/site.go
mas-who: Yes definitely. I didn't add it here yet since this PR is related to the database layer. There will be a followup PR that's dedicated for setting up all the API endpoints as well as there data structs

cmd/lxd-site-mgrd/main.go
mas-who: Good idea :+1:

internal/api/sites.go
edlerd: I think it is fine to keep some manual joins next to the generated "simple" db operations in the generated `*_mapper` files. As to the structure, I have no strong opinion how to organize the two. Maybe just another file `*_relation` for hand crafted joins instead of a dedicated package?

internal/database/site_member_statuses.go
mas-who: Ideally yes. But `lxd-generate` will generate two db functions with the same name `GetSiteMemberStatus` since it does not know how to handle the plural case. I think we actually don't need a helper function that only gets one entry really, so we can overcome this by getting rid of `//go:generate mapper method -i -d github.com/canonical/microcluster/cluster -e site_member GetOne`

internal/api/sites.go
mas-who: yeah sounds good :+1: I just wanted to make sure that if the functionality for multiple table joins exist in `lxd-generate`, we might as well use it haha

mas-who: I've added a `site_member_statuses.relation.go` file to hold the joint request.

internal/api/types/site.go
mas-who: I've added member statuses here now since backend is returning it

internal/api/sites.go
edlerd: perfect, thanks.

cmd/lxd-site-mgrd/main.go
mas-who: Added the dummy site data generator

edlerd: Thanks, this is a good improvement.

I think there was a misunderstanding. I meant not running this automatically post bootstrap, so removing this hook. We can have a non-hooked script, that can be manually triggered (or run as part of our CI and local dev setup) to have the dummy data imported. Because thinking far ahead on release, we wouldn't want the hook to be in place. It was good to have it on the start to have some data to play with the api.

internal/database/site_member_statuses.relation.go
```diff
@@ -0,0 +1,148 @@
...
+type MemberStatusWithSiteInfo struct {
+	ID               int       `db:"id"`
+	Name             string    `db:"name"`
+	SiteCertificate  string    `db:"site_certificate"`
+	CreatedAt        time.Time `db:"created_at"`
+	Status           string    `db:"site_status"`
+	InstanceCount    int       `db:"instance_count"`
+	InstanceStatuses string    `db:"instance_statuses"`
+	JoinedAt         time.Time `db:"joined_at"`
+	UpdatedAt        time.Time `db:"updated_at"`
```
edlerd: created and updated at are ambiguous, they could relate to the core_sites or site_details tables. Maybe avoid having them here to resolve it? Or we can have multiple createdAt and updatedAt fields with a prefix to clarify.

internal/database/site_member_statuses.relation.go
```diff
@@ -0,0 +1,148 @@
...
+	MemberName       string    `db:"member_name"`
+	Address          string    `db:"address"`
+	Architecture     string    `db:"architecture"`
+	Role             string    `db:"role"`
+	UsageCPU         float64   `db:"usage_cpu"`
+	UsageMemory      float64   `db:"usage_memory"`
+	UsageDisk        float64   `db:"usage_disk"`
+	MemberStatus     string    `db:"member_status"`
```
edlerd, line 31: This should be an array, I think you can use `json_group_array` for this. Currently, you would get multiple rows with the same site if a site has multiple members.

edlerd (commented):
Thanks, this is a perfect improvement.

cmd/lxd-site-mgrd/main.go
mas-who: Aaah got it. Maybe we can have a dedicated `/scripts` folder for this?

internal/database/site_member_statuses.relation.go
mas-who: I handle that at the API level. I think the data returned by these db level functions should be representative of the query statement and not do too much transformation?

cmd/lxd-site-mgrd/main.go
edlerd: Sounds good to me :+1:

internal/database/site_member_statuses.relation.go
mas-who: Yeah true, it is a little confusing since we have member details here as well. We can change the fields to be `SiteCreateAt`, `SiteUpdatedAt` (comes from `site_details` table btw) and `SiteJoinedAt`. I don't think we care too much about creation time for `site_details` entries and `site_member_statuses` entries.

internal/database/site_member_statuses.relation.go
edlerd: Right, I see. Was wondering if we could rely on dqlite for this with the json_group_array function. But then I see now there are some performance concerns with it and lxd never uses it. But maybe they have a solution for grouping in the database that we could adopt? Otherwise sounds good to transform in the api layer.

edlerd: I think lxd is doing individual queries for the grouped entity, also in the api layer i.e. with the `ClusterGroupToAPI` in db/cluster.go. This (our) approach seems to be fine then as well.

cmd/lxd-site-mgrd/main.go
mas-who: I've added a script for generating data and also combined the `package.json` scripts into `backend-dev` which injects the data post cluster initialisation. Please give it a try :)

internal/database/site_member_statuses.relation.go
mas-who: Yeah I mostly referenced examples from lxd or microcluster where output of join queries are transformed in go and not grouped in the database. I see `json_group_array` can work, but we would need to use it with the `json_object` function which I think would make the query difficult to understand. On the contrary doing the grouping and transforming the data in code is a lot easier to reason with. In any case, once we get the data out of the database, we'd probably still need to do some transformation in code anyway so I don't think grouping in database would be a huge performance boost.

internal/database/site_member_statuses.relation.go
mas-who: I've updated the struct time related fields

ui/package.json
```diff
@@ -22,6 +22,7 @@
     "start": "./entrypoint 0.0.0.0:${PORT}",
     "backend-run": "cd .. && make && go run ./cmd/lxd-site-mgrd --state-dir ./state/dir1/",
     "backend-init": "cd .. &&  go run ./cmd/lxd-site-mgr --state-dir ./state/dir1/ waitready && go run ./cmd/lxd-site-mgr --state-dir ./state/dir1/ init \"member1\" 0.0.0.0:9001 --bootstrap",
+    "backend-dev": "yarn backend-run & yarn backend-init && cd .. && scripts/generate_dummy_data.sh",
```
edlerd: Shall we make the generate dummy data call part of backend-init? I don't think we need an additional script entry.

scripts/generate_dummy_data.sh
```diff
@@ -0,0 +1,25 @@
...
+        VALUES (1, 'member2', '127.0.0.1:9002', 'x86_64', 'controller', 0.0, 0.0, 0.0, 'ACTIVE');
+    INSERT INTO site_member_statuses (core_site_id, member_name, address, architecture, role, usage_cpu, usage_memory, usage_disk, status) 
+        VALUES (2, 'member1', '127.0.0.1:9001', 'x86_64', 'controller', 0.0, 0.0, 0.0, 'ACTIVE');
+    INSERT INTO site_member_statuses (core_site_id, member_name, address, architecture, role, usage_cpu, usage_memory, usage_disk, status) 
+        VALUES (2, 'member2', '127.0.0.1:9002', 'x86_64', 'controller', 0.0, 0.0, 0.0, 'ACTIVE');
+    INSERT INTO site_member_statuses (core_site_id, member_name, address, architecture, role, usage_cpu, usage_memory, usage_disk, status) 
+        VALUES (3, 'member1', '127.0.0.1:9001', 'x86_64', 'controller', 0.0, 0.0, 0.0, 'ACTIVE');
+    INSERT INTO site_member_statuses (core_site_id, member_name, address, architecture, role, usage_cpu, usage_memory, usage_disk, status) 
+        VALUES (3, 'member2', '127.0.0.1:9002', 'x86_64', 'controller', 0.0, 0.0, 0.0, 'ACTIVE');
+"
```
edlerd:
```suggestion
"

```

edlerd (commented):
This looks good to merge. Tiny nit and an idea to simplify calling the generate script.

ui/package.json
mas-who: Yeah that sounds good, was thinking about updating docs for this but now we won't need to :+1:

mas-who: done

edlerd (approved):
LGTM, thanks for improving this further.

PR #25, opened 2024-06-26
add DELETE endpoint for `/sites/:name` [WD-11657]

mas-who:
## Done
 - Added a DELETE endpoint for `1.0/sites/:name`
 - Added some buttons in the UI for testing the delete endpoint

## Context
 - With #24 merged, this has become quite easy since we have already implemented the db methods.

### DIFF
--- a/ui/src/pages/sites/SiteList.tsx
+++ b/ui/src/pages/sites/SiteList.tsx
@@ -3,6 +3,7 @@ import { queryKeys } from "util/queryKeys";
 import { fetchSites } from "api/sites";
 import { useQuery } from "@tanstack/react-query";
 import { Link, MainTable } from "@canonical/react-components";
+import DeleteSiteButton from "./DeleteSiteButton";
 
 const SiteList: FC = () => {
   const { data: sites = [] } = useQuery({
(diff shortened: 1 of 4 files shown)


mas-who:
As a side note, I checked the delete on cascade on `site_details` and `site_member_statuses`, works nicely.

ui/src/pages/sites/SiteList.tsx
```diff
@@ -1,15 +1,24 @@
...
   });
 
+  const handleDeleteSite = async (siteName: string) => {
+    try {
+      await deleteSite(siteName);
+      await refetch();
+    } catch (error) {
+      console.error(error);
+    }
+  };
```
edlerd: This should be in a dedicated DeleteSiteButton component together with the button rendering. We should rely on cache invalidation instead of hooking into refetch to reduce coupling and to keep the delete function reusable.

edlerd (changes requested):
Shall we keep this as draft until the oidc PR is merged and include the permission checks here directly?

mas-who:
> Shall we keep this as draft until the oidc PR is merged and include the permission checks here directly?

Yeah sure, we can hold off on this PR until the authentication PR is merged.

ui/src/pages/sites/SiteList.tsx
mas-who: Made the suggested change :slightly_smiling_face:

mas-who:
@edlerd as discussed yesterday, I think we can merge this one if it's ready. I will apply the access handler on the oidc PR to this endpoint

edlerd (approved):
LGTM :+1:

PR #28, opened 2024-07-01
feat: LXD-SM setup application layout for the UI [WD-12637]

Kxiru:
- Added Navigation.tsx in it's own Components folder, as well as Logo and NavLink.
- Implemented the ApplicationLayout structure from react-components.
- Added a BaseLayout component as in LXD-UI.

image

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -25,8 +25,9 @@
     "test-e2e": "npx playwright test"
   },
   "dependencies": {
-    "@canonical/react-components": "0.58.0",
+    "@canonical/react-components": "0.59.0",
     "@tanstack/react-query": "5.40.0",
+    "@use-it/event-listener": "0.1.7",
     "formik": "2.4.6",
     "react": "18.3.1",
     "react-dom": "18.3.1",
--- a/ui/src/Root.tsx
+++ b/ui/src/Root.tsx
@@ -1,13 +1,19 @@
 import { FC } from "react";
 import App from "./App";
 import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { Application } from "@canonical/react-components";
+
+import Navigation from "./components/Navigation";
 
 const queryClient = new QueryClient();
 
 const Root: FC = () => {
   return (
     <QueryClientProvider client={queryClient}>
-      <App />
+      <Application>
+        <Navigation />
+        <App />
+      </Application>
     </QueryClientProvider>
   );
 };
--- a/ui/src/components/BaseLayout.tsx
+++ b/ui/src/components/BaseLayout.tsx
@@ -1,0 +1,27 @@
+import { FC, ReactNode } from "react";
+import { AppMain, Panel } from "@canonical/react-components";
+
+interface Props {
+  title: string | ReactNode;
+  controls?: ReactNode;
+  children: ReactNode;
+  mainClassName?: string;
+  contentClassName?: string;
+}
+
+const BaseLayout: FC<Props> = ({
+  title,
+  controls,
+  children,
+  mainClassName,
+}: Props) => {
+  return (
+    <AppMain className={mainClassName}>
+      <Panel title={title} controls={controls} wrapContent={true}>
+        {children}
+      </Panel>
+    </AppMain>
+  );
+};
+
+export default BaseLayout;
--- a/ui/src/components/Logo.tsx
+++ b/ui/src/components/Logo.tsx
@@ -1,0 +1,17 @@
+import { FC } from "react";
+import { NavLink } from "react-router-dom";
+
+const Logo: FC = () => {
+  return (
+    <NavLink className="p-panel__logo" to={`/`}>
+      <img
+        src="/ui/assets/img/lxd-logo.svg"
+        alt="LXD-UI logo"
+        className="p-panel__logo-image"
+      />
+      <div className="logo-text p-heading--4">LXD Site Manager</div>
+    </NavLink>
+  );
+};
+
+export default Logo;
--- a/ui/src/components/NavLink.tsx
+++ b/ui/src/components/NavLink.tsx
@@ -1,0 +1,47 @@
+import { FC, LinkHTMLAttributes, ReactNode } from "react";
+import { Link, useLocation } from "react-router-dom";
+import classnames from "classnames";
+
+interface Props {
+  to: string;
+  title: string;
+  children: ReactNode;
+  className?: string;
+  activeUrlMatches?: string[];
+}
+
+const NavLink: FC<Props & LinkHTMLAttributes<HTMLElement>> = ({
+  to,
+  title,
+  children,
+  className,
+  activeUrlMatches = [],
+  ...linkProps
+}) => {
+  const location = useLocation();
+
+  // ignore last char to consider plural form of URL Paths
+  const matchPart = to.substring(0, to.length - 1);
+  let isActive = location.pathname.startsWith(matchPart);
+
+  // this caters for more custom url matches for link to be active.
+  for (const match of activeUrlMatches) {
+    if (location.pathname.includes(match)) {
+      isActive = true;
+    }
+  }
+
+  return (
+    <Link
+      title={title}
+      to={to}
+      aria-current={isActive ? "page" : undefined}
+      className={classnames("p-side-navigation__link", className)}
+      {...linkProps}
+    >
+      {children}
+    </Link>
+  );
+};
+
+export default NavLink;
--- a/ui/src/components/Navigation.tsx
+++ b/ui/src/components/Navigation.tsx
@@ -1,0 +1,158 @@
+import { FC, MouseEvent, useState } from "react";
+import {
+  AppNavigation,
+  AppNavigationBar,
+  Button,
+  Icon,
+  Panel,
+} from "@canonical/react-components";
+import classnames from "classnames";
+import Logo from "./Logo";
+import NavLink from "components/NavLink";
+import { useMenuCollapsed } from "context/menuCollapsed";
+import { isWidthBelow } from "util/helpers";
+
+const Navigation: FC = () => {
+  const isSmallScreen = () => isWidthBelow(620);
+  const [isAuthenticated, setIsAuthenticated] = useState(true);
+  const { menuCollapsed, setMenuCollapsed } = useMenuCollapsed();
+
+  const handleLogout = () => {
+    setIsAuthenticated(false);
+    softToggleMenu();
+  };
+
+  const handleLogin = () => {
+    setIsAuthenticated(true);
+    softToggleMenu;
+  };
+
+  const softToggleMenu = () => {
+    if (isSmallScreen()) {
+      setMenuCollapsed((prev) => !prev);
+    }
+  };
+
+  const hardToggleMenu = (e: MouseEvent<HTMLElement>) => {
+    setMenuCollapsed((prev) => !prev);
+    e.stopPropagation();
+  };
+
+  return (
+    <>
+      <AppNavigationBar>
+        <Panel
+          className="is-dark"
+          stickyHeader={true}
+          logo={<Logo />}
+          controls={
+            <Button dense className="p-panel__toggle" onClick={hardToggleMenu}>
+              Menu
+            </Button>
+          }
+        />
+      </AppNavigationBar>
+      <AppNavigation
+        aria-label="main navigation"
+        className={classnames({
+          "is-collapsed": menuCollapsed,
+          "is-pinned": !menuCollapsed,
+        })}
+      >
+        <Panel
+          logo={<Logo />}
+          dark={true}
+          controls={
+            <Button
+              appearance="base"
+              hasIcon
+              className="u-no-margin"
+              aria-label="close navigation"
+              onClick={hardToggleMenu}
+            >
+              <Icon name="close" />
+            </Button>
+          }
+          controlsClassName="u-hide--medium u-hide--large"
+        >
+          <div className="p-side-navigation--icons is-dark">
+            <ul className="p-side-navigation__list sidenav-top-ul">
+              {isAuthenticated && (
+                <li>
+                  <NavLink
+                    to={`/ui/sites`}
+                    title={`Sites List`}
+                    onClick={softToggleMenu}
+                  >
+                    <Icon
+                      className="is-light p-side-navigation__icon"
+                      name="units"
+                    />
+                    Sites
+                  </NavLink>
+                </li>
+              )}
+              {!isAuthenticated && (
+                <li>
+                  <NavLink to="/ui/login" title="Login" onClick={handleLogin}>
+                    <Icon
+                      className="is-light p-side-navigation__icon"
+                      name="profile"
+                    />
+                    Login
+                  </NavLink>
+                </li>
+              )}
+            </ul>
+            <ul
+              className={classnames(
+                "p-side-navigation__list sidenav-bottom-ul",
+                {
+                  "authenticated-nav": isAuthenticated,
+                },
+              )}
+            >
+              <hr className="is-dark navigation-hr" />
+              {isAuthenticated && (
+                <li className="p-side-navigation__item">
+                  <a
+                    className="p-side-navigation__link"
+                    title="Log out"
+                    onClick={handleLogout}
+                  >
+                    <Icon
+                      className="is-light p-side-navigation__icon p-side-logout"
+                      name="export"
+                    />
+                    Log out
+                  </a>
+                </li>
+              )}
+            </ul>
+          </div>
+        </Panel>
+
+        <div
+          className={classnames("sidenav-toggle-wrapper", {
+            "authenticated-nav": isAuthenticated,
+          })}
+        >
+          <Button
+            appearance="base"
+            aria-label={`${
+              menuCollapsed ? "expand" : "collapse"
+            } main navigation`}
+            hasIcon
+            dense
+            className="sidenav-toggle is-dark u-no-margin l-navigation-collapse-toggle u-hide--small"
+            onClick={hardToggleMenu}
+          >
+            <Icon light name="sidebar-toggle" />
+          </Button>
+        </div>
+      </AppNavigation>
+    </>
+  );
+};
+
+export default Navigation;
--- a/ui/src/context/menuCollapsed.tsx
+++ b/ui/src/context/menuCollapsed.tsx
@@ -1,0 +1,20 @@
+import { useState } from "react";
+import useEventListener from "@use-it/event-listener";
+import { isWidthBelow } from "util/helpers";
+
+const isSmallScreen = () => isWidthBelow(620);
+const isMediumScreen = () => isWidthBelow(820);
+
+export const useMenuCollapsed = () => {
+  const [menuCollapsed, setMenuCollapsed] = useState(isMediumScreen());
+
+  const collapseOnMediumScreen = () => {
+    if (isSmallScreen()) {
+      return;
+    }
+  };
+
+  useEventListener("resize", collapseOnMediumScreen);
+
+  return { menuCollapsed, setMenuCollapsed };
+};
--- a/ui/src/pages/Login.tsx
+++ b/ui/src/pages/Login.tsx
@@ -4,7 +4,7 @@ import { Link, LoginPageLayout } from "@canonical/react-components";
 const Login: FC = () => {
   return (
     <LoginPageLayout title="Login to LXD site manager">
-      <Link href="/oidc/login" className="p-button">
+      <Link href="/oidc/sites" className="p-button">
         Login
       </Link>
     </LoginPageLayout>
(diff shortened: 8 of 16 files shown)


edlerd (commented):
As discussed, let's add the navigation in this PR as well.

ui/src/Components/Navigation.tsx
```diff
@@ -0,0 +1,7 @@
+import { FC } from "react";
```
mas-who: can we have the `Components` folder not capitalised?

ui/src/Root.tsx
```diff
@@ -1,13 +1,17 @@
...
 import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import Navigation from "./Components/Navigation";
 
 const queryClient = new QueryClient();
 
 const Root: FC = () => {
   return (
     <QueryClientProvider client={queryClient}>
-      <App />
+      <div id="l-application" className="l-application" role="presentation">
```
mas-who: @edlerd @Kxiru , I think since we are adding the application layout, we should probably also create a layout component similar to the `BaseLayout` component in lxd-ui? Maybe we should use composite components pattern to avoid passing down too many props

ui/package.json
```diff
@@ -25,8 +25,9 @@
     "test-e2e": "npx playwright test"
   },
   "dependencies": {
-    "@canonical/react-components": "0.58.0",
+    "@canonical/react-components": "0.59.0",
     "@tanstack/react-query": "5.40.0",
+    "@use-it/event-listener": "^0.1.7",
```
mas-who: please remove `^`

ui/src/Root.tsx
```diff
@@ -1,13 +1,20 @@
...
 
 const queryClient = new QueryClient();
 
 const Root: FC = () => {
   return (
     <QueryClientProvider client={queryClient}>
-      <App />
+      <Application>
+        <Navigation />
+        <AppStatus />
```
mas-who: we can leave out `AppStatus` for now I rate. AFAIK we don't have a status bar in the design yet correct?

ui/src/components/BaseLayout.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+const BaseLayout: FC<Props> = ({
+  title,
+  controls,
+  children,
+  mainClassName,
+  contentClassName,
+}: Props) => {
+  return (
+    <main className={classnames("l-main", mainClassName)}>
```
mas-who: `AppMain` already has the `l-main` class, so you can remove this `main` element. You can also just pass `mainClassName` to `AppMain`. See [AppMain](https://github.com/canonical/react-components/blob/main/src/components/ApplicationLayout/AppMain/AppMain.tsx) in react-components. It has `classnames` implemented so your custom classes will be appended to `l-main`.

ui/src/components/BaseLayout.tsx
```diff
@@ -0,0 +1,37 @@
...
+      <AppMain>
+        <Panel>
```
mas-who: The `Panel` component already has the div with `p-panel__header`, `p-panel__title`, `p-panel__controls` and `p-panel__content` classes. You should use the relevant props on the `Panel` component to modify relevant aspects of the panel section. see [docs](https://canonical.github.io/react-components/?path=/docs/components-panel--docs)

ui/src/components/Logo.tsx
```diff
@@ -0,0 +1,24 @@
+import { FC } from "react";
+// import { useProject } from "context/project";
```
mas-who: remove comments through out please

ui/src/components/Logo.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { NavLink } from "react-router-dom";
+
+const Logo: FC = () => {
+  // const { project, isLoading } = useProject();
+
+  const getLogoLink = () => {
```
mas-who: we can remove this function for now

ui/src/components/NavLink.tsx
```diff
@@ -0,0 +1,47 @@
...
+  to,
+  title,
+  children,
+  className,
+  activeUrlMatches = [],
+  ...linkProps
+}) => {
+  const location = useLocation();
+
+  // ignore last char to consider /instances and /instance as active
```
mas-who: these comments are specific to lxd-ui. The logic is still relevant though. I suggest we change to to something a bit more generic like "ignore last char to consider plural form of url paths".

ui/src/components/NavLink.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const matchPart = to.substring(0, to.length - 1);
+  let isActive = location.pathname.startsWith(matchPart);
+
+  // this caters for more custom url matches for link to be active e.g. /identities & /identity/oidc etc
```
mas-who: same thing here, maybe just remove the example part of the comment

ui/src/components/Navigation.tsx
```diff
@@ -0,0 +1,187 @@
...
+import { useMenuCollapsed } from "context/menuCollapsed";
+import { isWidthBelow } from "util/helpers";
+
+const Navigation: FC = () => {
+  const isSmallScreen = () => isWidthBelow(620);
+  const [isAuthenticated, setIsAuthenticated] = useState(true);
+  const { menuCollapsed, setMenuCollapsed } = useMenuCollapsed();
+  const [scroll, setScroll] = useState(false);
+
+  // const { isAuthenticated } = useAuth();
```
mas-who: remove

ui/src/components/Navigation.tsx
```diff
@@ -0,0 +1,187 @@
...
+    e.stopPropagation();
+  };
+
+  return (
+    <>
+      <AppNavigationBar>
+        <Panel
+          className="is-dark"
+          logo={<Logo />}
+          controls={
```
mas-who, line 48: `controls` is already a child of `p-panel__header`. So this is incorrect. You should use `controls`, `logo` and `toggle` props from the `Panel` component to replicate this section (relative to lxd-ui). See the [Panel](https://github.com/canonical/react-components/blob/main/src/components/Panel/Panel.tsx) component.

ui/src/components/Navigation.tsx
```diff
@@ -0,0 +1,187 @@
...
+                  dense
+                  className="p-panel__toggle"
+                  onClick={hardToggleMenu}
+                >
+                  Menu
+                </Button>
+              </div>
+            </div>
+          }
+        ></Panel>
```
mas-who: If there is no child, prefer self-closing tags e.g. `<Panel />` instead of `<Panel></Panel>`

ui/src/components/Navigation.tsx
```diff
@@ -0,0 +1,187 @@
...
+      </AppNavigationBar>
+      <AppNavigation
+        aria-label="main navigation"
+        className={classnames("l-navigation", {
```
mas-who: no need to have `l-navigation` here, it's already in `AppNavigation`

ui/src/components/Navigation.tsx
```diff
@@ -0,0 +1,187 @@
...
+          "is-collapsed": menuCollapsed,
+          "is-pinned": !menuCollapsed,
+          "is-scroll": scroll,
```
mas-who: we don't need `is-scroll` in site manager yet. We can remove this for now. Please remove the relevant state, effects and sass styles as well.

ui/src/components/Navigation.tsx
```diff
@@ -0,0 +1,187 @@
...
+        })}
+      >
+        <SideNavigation>
```
mas-who: Why are we using `SideNavigation` here. To replicate lxd-ui, `AppNavigation` already has the `l-navigation__drawer` class inside the component

ui/src/components/Navigation.tsx
```diff
@@ -0,0 +1,187 @@
...
+          <div className="p-panel is-dark">
```
mas-who: We should use the `Panel` component here.

ui/src/components/Navigation.tsx
```diff
@@ -0,0 +1,187 @@
...
+                  <Icon name="close" />
+                </Button>
+              </div>
+            </div>
+            <div className="p-panel__content">
+              <div className="p-side-navigation--icons is-dark">
+                <ul className="p-side-navigation__list sidenav-top-ul">
+                  {isAuthenticated && (
+                    <>
+                      <SideNavigationItem href={`/ui/sites`}>
```
mas-who: same comment as above, why are we using `SideNavigationItem` here?

ui/src/components/Navigation.tsx
```diff
@@ -0,0 +1,187 @@
...
+                  aria-label="close navigation"
+                  onClick={hardToggleMenu}
+                >
+                  <Icon name="close" />
+                </Button>
+              </div>
+            </div>
+            <div className="p-panel__content">
+              <div className="p-side-navigation--icons is-dark">
+                <ul className="p-side-navigation__list sidenav-top-ul">
```
mas-who: @edlerd are we going to have the same navigation structure as lxd-ui? i.e. bottom half of the nav has fixed height and sticks to the bottom of the viewport. If so, we should probably bring in the vertical adjustment logic now.

ui/src/components/Navigation.tsx
```diff
@@ -0,0 +1,187 @@
...
+                      <hr className="is-dark navigation-hr" />
+                    </>
+                  )}
+                  {!isAuthenticated && (
+                    <>
+                      <SideNavigationItem href={`/ui/login`}>
+                        <NavLink
+                          to="/ui/login"
+                          title="Login"
+                          onClick={() => {
```
mas-who: let's put this callback into a `handleLogin` function

ui/src/components/Navigation.tsx
```diff
@@ -0,0 +1,187 @@
...
+                      "authenticated-nav": isAuthenticated,
+                    },
+                  )}
+                >
+                  {isAuthenticated && (
+                    <li className="p-side-navigation__item">
+                      <a
+                        className="p-side-navigation__link"
+                        title="Log out"
+                        onClick={() => {
```
mas-who: let's put this callback into a `handleLogout` function. Move logic from `logoutFunc` into `handleLogout`

mas-who (changes requested):
Providing some feedback so far, I haven't completely reviewed everything yet. Will continue tomorrow.

Screenshot from 2024-07-03 19-16-49

ui/src/Root.tsx
Kxiru: Not yet, I suppose. Done!

ui/src/components/BaseLayout.tsx
Kxiru: I believe I forgot to take out the main element. Thanks for spotting this!

ui/package.json
Kxiru: We need this for context/Menucollapsed.tsx. Unless you have a workaround?

mas-who: I meant remove the `^` symbol to keep version fixed

ui/src/context/menuCollapsed.tsx
```diff
@@ -0,0 +1,33 @@
+import { useState } from "react";
```
mas-who, line 1: I think this file can be simplified for now. For `collapseOnMediumScreen`, you can remove the if statement that's checking the event detail since it's related to `spice` in lxd-ui and `sfp-toggle` event which currently we don't have in site manager. You can also remove all logic related to search filter here. We should start off with the minimal amount of logic necessary when copying over existing code, so we don't add complexity to this project if not needed.

ui/src/pages/Login.tsx
```diff
@@ -4,7 +4,7 @@ import { Link, LoginPageLayout } from "@canonical/react-components";
 const Login: FC = () => {
   return (
     <LoginPageLayout title="Login to LXD site manager">
-      <Link href="/oidc/login" className="p-button">
+      <Link href="/ui/sites" className="p-button">
```
mas-who: why is this changed? We need to have the `/oidc/login` path so that a request will be sent to the backend.

ui/src/pages/sites/SiteList.tsx
```diff
@@ -12,35 +13,36 @@ const SiteList: FC = () => {
...
-              { content: site.instance_count },
-              {
-                content: <DeleteSiteButton siteName={site.name} />,
-              },
-            ],
-          };
-        })}
-      />
+    <BaseLayout title="Sites">
+      <div>
```
mas-who: we should apply the [vanilla grid layout](https://vanillaframework.io/docs/patterns/grid) here. The equivalent components in react-components are [Row](https://github.com/canonical/react-components/tree/main/src/components/Row) and [Col](https://github.com/canonical/react-components/tree/main/src/components/Col). The grid pattern in vanilla aligns elements within a 12 column grid, which is used for all canonical apps/websites, so it's a good opportunity to familiarise yourself with it.

ui/src/pages/sites/SiteList.tsx
```diff
@@ -12,35 +13,36 @@ const SiteList: FC = () => {
...
+              ],
+            };
+          })}
+        />
 
-      <Link href="/oidc/logout" className="p-button">
-        Logout
-      </Link>
-    </div>
+        <Link href="/oidc/logout" className="p-button">
```
mas-who: we can remove this logout button here.

ui/src/sass/_pattern_icon.scss
```diff
@@ -0,0 +1,15 @@
+@mixin lxdui-icon-sidebar-collapse() {
```
mas-who: we should replace `lxdui` in these mixin names. perhaps something like `sidebar-collapse-icon` and `custom-p-icon`. Prefer having more generic naming here instead of adding a prefix of `lxdsm` or something alike since we might change `site manager` to `cluster manager` later, so let's try avoid having to come back for updates where possible.

ui/src/sass/styles.scss
```diff
@@ -1 +1,34 @@
 @import "vanilla-framework/scss/build";
+@import "settings";
+@import "pattern_navigation";
+@import "./breakpoints";
+@import "pattern_icon";
+@include lxdui-p-icon;
```
mas-who: change mixin name

ui/src/sass/_pattern_navigation.scss
```diff
@@ -0,0 +1,194 @@
...
+}
+
+@include mobile {
+  .l-navigation {
+    transition-duration: 0s !important;
+  }
+}
+
+@media screen and (min-width: $breakpoint-x-small) and (max-width: $breakpoint-small - 1px) {
+  .l-navigation__drawer {
```
mas-who: this class isn't used anywhere, but we probably should remove it since it was definitely needed in lxd-ui. This is probably related to my previous comment about using the `SideNavigation` component from react-components. Would you please investigate.

ui/src/sass/_pattern_navigation.scss
```diff
@@ -0,0 +1,194 @@
...
+      left: 2.5rem;
+    }
+  }
+}
+
+.p-side-logout {
+  rotate: 90deg;
+}
+
+.accordion-nav-menu {
```
mas-who: this class is not used, please remove, if we need it later we can copy it over from lxd-ui

ui/src/sass/_pattern_navigation.scss
```diff
@@ -0,0 +1,194 @@
...
+  .open {
+    rotate: 180deg;
+  }
+
+  &:hover {
+    cursor: pointer;
+  }
+}
+
+.accordion-nav-secondary {
```
mas-who: we can remove this

mas-who (commented):
some more comments for code changes.

ui/src/components/Navigation.tsx
Kxiru: Thank you for the link, however I still do not understand what toggle is supposed to be.  By name it is implied to be a function but by the link you sent me, it is implied to be a component. However if the Button component is passed through to toggle instead, then there are no longer any controls. Please could you clarify what you mean, here?
image

mas-who:
Currently the `hr` element sit right below the top half of the navigation.
Screenshot from 2024-07-04 10-36-37

I think it should sit above the bottom half of the navigation similar to lxd-ui, no?
Screenshot from 2024-07-04 10-37-55

ui/src/components/Navigation.tsx
Kxiru: The transfer was well and good until Toggle got in the way again. This is all that can be passed through to it (see screenshot), I can't find much documentation on it, and when used as such the toggle does not appear in LXDSM. I am open to further investigation.
image

ui/src/pages/sites/SiteList.tsx
Kxiru: Sorry, I don't understand, you would like me to implement a grid layout in Base Layout? Could you please provide an example of this?

ui/src/sass/_pattern_navigation.scss
Kxiru: I believe we are good to remove this.

ui/src/components/Navigation.tsx
edlerd: That is not decided yet, I'd keep the nav simple for now.

ui/src/components/Navigation.tsx
mas-who: Yeah this looks good now. I think since you are using `controls` it is fine. `toggle` would be a replacement for your current implementation which I don't think is necessary.

ui/src/components/Navigation.tsx
mas-who: can't find the original code so it's a bit confusing to review now since the context is lost, but I think your current implementation using `controls` is fine. Related to a previous comment, I think `toggle` is just a convenient way to add a toggle button for the panel, you can achieve the same using the `controls` prop.

ui/src/pages/sites/SiteList.tsx
mas-who: see [ClusterList.tsx](https://github.com/canonical/lxd-ui/blob/main/src/pages/cluster/ClusterList.tsx) from lxd-ui. The `Row` component wraps around children components to establish the 12 column grid. You can achieve variations of the grid pattern using combination of the `Row` and `Col` components which relates to the `row` and `col` classes in vanilla.

Kxiru: Implemented. Looking much better!

ui/src/components/Navigation.tsx
```diff
@@ -0,0 +1,158 @@
...
+      </AppNavigationBar>
+      <AppNavigation
+        aria-label="main navigation"
+        className={classnames({
+          "is-collapsed": menuCollapsed,
+          "is-pinned": !menuCollapsed,
+        })}
+      >
+        <Panel
+          logo={<Logo />}
```
mas-who, line 63: missing a prop here, `stickyHeader` on `Panel` should be set

ui/src/components/Navigation.tsx
```diff
@@ -0,0 +1,158 @@
...
+              className="u-no-margin"
+              aria-label="close navigation"
+              onClick={hardToggleMenu}
+            >
+              <Icon name="close" />
+            </Button>
+          }
+          controlsClassName="u-hide--medium u-hide--large"
+        >
+          {" "}
```
mas-who: this empty space shouldn't be here right?

ui/src/components/Navigation.tsx
```diff
@@ -0,0 +1,158 @@
...
+          <div className="p-side-navigation--icons is-dark">
+            <ul className="p-side-navigation__list sidenav-top-ul">
+              {isAuthenticated && (
+                <>
+                  <NavLink
```
mas-who, line 82: each `NavLink` should be inside an `li`

ui/src/pages/sites/SiteList.tsx
```diff
@@ -12,35 +13,34 @@ const SiteList: FC = () => {
...
-        })}
-      />
-
-      <Link href="/oidc/logout" className="p-button">
-        Logout
-      </Link>
-    </div>
+    <BaseLayout title="Sites">
+      <Row>
+        <div>
```
mas-who: this `div` won't be needed since `Row` is a div already

mas-who (commented):
almost there, few minor comments :slightly_smiling_face:

mas-who:
great work @Kxiru it's looking like a proper app now :smile: 

one query from me. I think the app title next to the logo is way too close to the edge of the navigation and it looks cramped. Not sure how we deal with this though since there isn't enough space. @piperdeck any ideas?

Screenshot from 2024-07-04 22-55-43

ui/src/sass/styles.scss
```diff
@@ -1 +1,34 @@
 @import "vanilla-framework/scss/build";
+@import "settings";
```
mas-who, line 2: final comment here. Can you add the below style to this file? This will remove the max-width constraint allowing the `row` grid to not have margin for large screen sizes. This should align the header and body of the page nicely.

```
.l-main .row {
  max-width: none;
}
```

Kxiru: All done! If that's all, then I'm just awaiting an approval now :)

mas-who (approved):

mas-who:
LGTM :+1: Thanks for this!

PR #29, opened 2024-07-01
add POST `/1.0/external-site-tokens` endpoint [WD-12698]

mas-who:
## Done

- [x] setup `POST /1.0/external-site-join-token` 
- [x] setup utility function to calculate site manager addresses
- [x] setup `GET /1.0/external-site-join-token` 
- [x] setup `DELETE /1.0/external-site-join-token/:site_name`  

## Note
 - I have included a commit here that introduced experimental hooks on cluster bootsrap and member join. This is for testing the site manager address generation logic. I will drop the commit once the PR is approved.
 - The manager address logic will likely fail before we get the ability to provide network listener with custom addresses.


internal/api/tokens.go
```diff
@@ -0,0 +1,90 @@
...
+func tokenPost(s *state.State, r *http.Request) response.Response {
+	payload := types.ExternalSiteTokenPost{}
+
+	err := json.NewDecoder(r.Body).Decode(&payload)
+	if err != nil {
+		return response.BadRequest(err)
+	}
+
+	// validate expiry is in the future
+	if payload.Expiry != (time.Time{}) && payload.Expiry.Before(time.Now()) {
```
edlerd: Expiry should also be set, it is not allowed to be empty.

internal/api/tokens.go
```diff
@@ -0,0 +1,90 @@
...
+	if err != nil {
+		return response.SmartError(err)
+	}
+
+	clusterCert, err := s.ClusterCert().PublicKeyX509()
+	if err != nil {
+		return response.InternalError(err)
+	}
+
+	// TODO: these addresses should be calculated based on configs, extract to util function
```
edlerd: We have defined a 3 way options:
1. config key if it is present with the array of members, unsorted just as is.
3. the external ip on the members table, this should be sorted so db-leader is in 1st place.
4. the default address on the members table, this should be sorted so db-leader is in 1st place.

We should check from 1 and fall back successively until we reach 3. This should be in a util function as the comment suggests.

internal/api/types/tokens.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+// ExternalSiteToken is the response body for an external site token.
+type ExternalSiteToken struct {
+	ExternalSiteTokenPost
+	CreateAt time.Time `json:"created_at"`
+}
+
+// ExternalSiteTokenBody is the body of the external site token.
+type ExternalSiteTokenBody struct {
```
edlerd, line 29: How do we share this and other structs with LXD source that is receiving and unpacking them?

edlerd (changes requested):

internal/api/tokens.go
mas-who: I don't recall we have this requirement? But sure, we can enforce that expiry should always be set for the token.

internal/api/tokens.go
mas-who: Yeah that will be the logic. For point 1, I don't think we will return any member address and just return an array of a single address that is the `global.address` config value.

internal/api/types/tokens.go
mas-who: We can put these outside of the `internal` directory. Although, a project that's importing dependencies from this project will need to do some extra setup since this is a private repo.

edlerd: Let's ignore the problem for now, I'll put it on our next sync agenda.

internal/api/tokens.go
edlerd: I think we do not want tokens without an expiry date.

internal/api/tokens.go
edlerd: The configured value in 1 should be an array as well, to allow multiple addresses. I agree, it can be an array of 1. But we shouldn't limit it to that.

cmd/lxd-site-mgrd/main.go
```diff
@@ -73,9 +76,42 @@ func (c *cmdDaemon) run(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	hooks := &config.Hooks{
```
mas-who: don't worry about this, I will drop the commit once PR is approved. Keeping this here for now in case if anyone wants to checkout the branch and do some testing

scripts/create_join_token.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+set -e
+
+# Execute the curl command with the provided values
+curl --insecure -H 'Content-Type: application/json' -d '{
+  "expiry": "2025-01-01T00:00:00Z",
+  "site_name": "site1"
+}' -X POST https://0.0.0.0:9001/1.0/external-site-join-token
```
edlerd:
```suggestion
}' -X POST https://0.0.0.0:9001/1.0/external-site-join-token
```

cmd/lxd-site-mgrd/main.go
edlerd: I think we need to keep this -- at least the post join hook, no?

cmd/lxd-site-mgrd/main.go
```diff
@@ -73,9 +76,42 @@ func (c *cmdDaemon) run(cmd *cobra.Command, args []string) error {
...
+		PostBootstrap: func(s *state.State, initConfig map[string]string) error {
+			return s.Database.Transaction(s.Context, func(ctx context.Context, tx *sql.Tx) error {
+				_, err := database.CreateManagerConfig(ctx, tx, database.ManagerConfig{
+					Key:   "global.address",
+					Value: "172.168.10.01",
+				})
+				if err != nil {
+					return err
+				}
```
edlerd: This is better in a script to bootstrap data?

internal/api/tokens.go
```diff
@@ -0,0 +1,254 @@
...
+		}
+
+		// check if any external addresses are set
+		for _, memberConfig := range memberConfigs {
+			if memberConfig.ExternalAddress != "" {
+				hasExternalAddresses = true
+				break
+			}
+		}
+
```
edlerd: I think this should be moved further down. We don't need to read manager configs in case a globalAddress is found.

internal/api/tokens.go
```diff
@@ -0,0 +1,254 @@
...
+	return response.SyncResponse(true, nil)
+}
+
+// getSiteManagerAddresses returns the addresses of the site managers that are online.
+// 1. If the global address is set, it is returned as the first address.
+// 2. If any external addresses are set, external addresses are returned.
+// 3. If none of the above conditions are met, the HTTPS address is returned.
+// For 2. and 3., the leader address is returned as the first address always.
+func getSiteManagerAddresses(s *state.State) ([]string, error) {
```
edlerd: I don't think we should filter offline members. We should sort them to the end of the array instead, and put the leader first as now.

Also, can we break this function down so the 1,2,3 steps are obvious by reading the code instead of needing this comment here explaining what the code is doing below? Ideally, the code tells you what it is doing, and the comments tell you why something is done -- which can't be expressed in the code itself.

cmd/lxd-site-mgrd/main.go
mas-who: Yes we do need this hook when the ability to provide configs for network listeners during initialisation is enabled in microcluster. For now, I wouldn't really want to keep this commit for this PR. Soon I will have a PR up for configs CLI, and I plan to write a script using the CLI to initialise the configs just so that we have some dummy data so that the site manager addresses calculation won't break.

cmd/lxd-site-mgrd/main.go
mas-who: yes I agree, as mentioned in the previous comment, I will setup a script in the configs CLI PR to populate some data.

internal/api/tokens.go
mas-who: I don't think there is any value in sending offline member addresses to LXD? This calculation would be performed for every status update eventually so LXD would always get the latest valid addresses. Including offline member addresses means we would need to setup a separate slice to keep track of those addresses then combine them with the valid addresses. This only slightly complicates the logic, but I don't see it being necessary?

I've split out the logic into smaller functions and removed the comments :+1:

cmd/lxd-site-mgrd/main.go
edlerd: I think the configs already merged, so we can have the script here with some api calls?

internal/api/tokens.go
edlerd: The reasoning to include offline members is that they might become online and all other members offline. So it will be beneficial for a cluster to know all possible members, even if they are offline during the time of a status update.

cmd/lxd-site-mgrd/main.go
mas-who: I was thinking since we will have the CLI ready, might as well write the shell script with the CLI instead of using `curl`?

edlerd: Sounds good to me. Then let's keep this on in draft until the CLI is merged and keep the CLI PR focused on adding the feature.

internal/api/tokens.go
mas-who: Yeah that's a good point, I've made the change to append offline member address to the end of the site manager addresses slice :+1:

cmd/lxd-site-mgrd/main.go
mas-who: Here is the CLI [PR](https://github.com/canonical/lxd-site-manager/pull/38) with the config generation script

mas-who:
@edlerd I have rebased main now that the configs generation script is there. I've marked the PR as non-draft now

internal/api/tokens.go
```diff
@@ -0,0 +1,285 @@
...
+		SiteName:             payload.SiteName,
+		SiteManagerCert:      shared.CertFingerprint(clusterCert),
+	}
+
+	encodedToken, err := token.Encode()
+	if err != nil {
+		return response.InternalError(err)
+	}
+
+	return response.SyncResponse(true, encodedToken)
```
edlerd: I think this needs to be an object, so we don't send the token directly under the metadata field, but wrapped as a named field. That way, we ensure that we can extend the object later.

internal/api/tokens.go
```diff
@@ -0,0 +1,285 @@
...
+
+	var responseTokens []types.ExternalSiteToken
+	for _, token := range tokens {
+		responseTokens = append(responseTokens, types.ExternalSiteToken{
+			ExternalSiteTokenPost: types.ExternalSiteTokenPost{
+				Expiry:   token.Expiry,
+				SiteName: token.SiteName,
+			},
+			CreateAt: token.CreatedAt,
+		})
```
edlerd:
```suggestion
		responseTokens = append(responseTokens, types.ExternalSiteToken{
			Expiry:   token.Expiry,
			SiteName: token.SiteName,
			CreateAt: token.CreatedAt,
		})
```
Can't this be flattened?

internal/api/tokens.go
```diff
@@ -0,0 +1,285 @@
...
+// getLeaderPrioritisedAddresses returns the addresses of site manager members (external if any is set) with the leader as the first address.
+func getLeaderPrioritisedAddresses(s *state.State, memberConfigs []database.ManagerMemberConfig) ([]string, error) {
+	hasExternalAddresses := checkExternalAddresses(memberConfigs)
+	leaderName, onlineMembers, err := getLeaderAndOnlineMembers(s)
+	if err != nil {
+		return nil, err
+	}
+
+	siteManagerAddresses := make([]string, 0, len(memberConfigs))
+	offlineMemberAddresses := make([]string, 0, len(memberConfigs))
```
edlerd: There are more states than just online and offline. These are non-online members, not necessarily offline members.

internal/api/types/tokens.go
```diff
@@ -0,0 +1,39 @@
...
+
+// ExternalSiteTokenPost is the request body for creating a new external site token.
+type ExternalSiteTokenPost struct {
+	Expiry   time.Time `json:"expiry"`
+	SiteName string    `json:"site_name"`
+}
+
+// ExternalSiteToken is the response body for an external site token.
+type ExternalSiteToken struct {
+	ExternalSiteTokenPost
```
edlerd: A bit surprised to see the post type included here. I see there is an overlap, though not necessarily a semantic connection.

internal/api/tokens.go
mas-who: The `ExternalSiteToken` struct type embeds the `ExternalSiteTokenPost` struct type. This is how the syntax works to construct `ExternalSiteToken`, but it does not affect the end result i.e. the json content of the decoded token would be flattened

internal/api/types/tokens.go
mas-who: The intended connection here is that `ExternalSiteToken` represent the data that would be returned from the `GET` endpoint, and it would contain the data from `ExternalSiteTokenPost`

internal/api/tokens.go
edlerd: This code is a bit unclear, it seems there is a post involved here, but the data is coming entirely from the database. I'd suggest keeping the types independent of each other to avoid making a connection here, where there is none.

edlerd (approved):
LGTM, thanks for the improvements.

PR #30, opened 2024-07-02
add site management listener `PATCH` and `GET` endpoints for `/1.0/configs` [WD-12693]

mas-who:
## Done
- [x]  ~~Add `PUT /1.0/configs` which will replace all existing configs with payload~~
- [X]  Add `PATCH /1.0/configs` which will upsert config key/values
- [X]  Add `GET /1.0/configs` return all site manager configs

### DIFF
--- a/internal/api/configs.go
+++ b/internal/api/configs.go
@@ -1,0 +1,80 @@
+package api
+
+import (
+	"context"
+	"database/sql"
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	"github.com/canonical/lxd/lxd/db/query"
+	"github.com/canonical/lxd/lxd/response"
+	"github.com/canonical/microcluster/rest"
+	"github.com/canonical/microcluster/state"
+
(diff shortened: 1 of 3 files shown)


internal/api/configs.go
```diff
@@ -0,0 +1,118 @@
...
+
+var configsCmd = rest.Endpoint{
+	Path:  "config",
+	Put:   rest.EndpointAction{Handler: configsPut, AllowUntrusted: true},
+	Patch: rest.EndpointAction{Handler: configsPatch, AllowUntrusted: true},
+	Get:   rest.EndpointAction{Handler: configsGet, AllowUntrusted: true},
+}
+
+// completely replace all manager configs with payload.
+func configsPut(s *state.State, r *http.Request) response.Response {
```
edlerd: As discussed, we don't really need the put endpoint, and it introduces a requirement for ETAG or alike. Let's remove it, so we can keep things simple.

edlerd (changes requested):

edlerd (approved):
LGTM :+1:

PR #32, opened 2024-07-03
removed site_member_statuses from db schema and updated sites api

mas-who:
## Done
- [x] removed `site_member_statuses` from db schema
- [x] removed `site_member_statuses` related helper functions
- [x] updated helper functions for `site_details` to join with `core_sites`
- [x] updated sites api types and endpoints
- [x] updated dummy data generation for sites


internal/api/sites.go
```diff
@@ -29,19 +29,24 @@ var siteCmd = rest.Endpoint{
 }
 
 func sitesGet(s *state.State, r *http.Request) response.Response {
-	var dbMemberStatusesForAllSites []database.MemberStatusWithSiteInfo
+	var dbMemberStatusesForAllSites []database.CoreSiteDetail
```
edlerd: dbMemberStatusesForAllSites seems wrong naming?

internal/api/sites.go
```diff
@@ -50,10 +55,10 @@ func siteGet(s *state.State, r *http.Request) response.Response {
 		return response.SmartError(err)
 	}
 
-	var dbSiteMemberStatuses []database.MemberStatusWithSiteInfo
+	var dbSiteMemberStatuses []database.CoreSiteDetail
```
edlerd: dbSiteMemberStatuses naming

internal/database/site_details.relation.go
```diff
@@ -0,0 +1,137 @@
...
+// CoreSiteDetail is a struct that contains all the information about a site directly queried from the database.
+type CoreSiteDetail struct {
+	ID               int       `db:"id"`
+	Name             string    `db:"name"`
+	SiteCertificate  string    `db:"site_certificate"`
+	SiteCreatedAt    time.Time `db:"created_at"`
+	Status           string    `db:"status"`
+	UsageCPU         float64   `db:"usage_cpu"`
+	UsageMemory      float64   `db:"usage_memory"`
+	UsageDisk        float64   `db:"usage_disk"`
```
edlerd: I don't think a single usage number is enough. I think for all three, we'd want to also have the total available.
Also might be better to switch the naming around to have the specifics first. Maybe CpuUsage and CpuTotalCount, MemoryUsage, MemoryTotalAmount, DiskUsage and DiskTotalSize?

scripts/generate_dummy_data.sh
edlerd, line 1: Can we extend this for testing with the scale case? We should have ~200 sites and they vary in their data a bit more.

internal/api/sites.go
edlerd: The type also seems weirdly named, shouldn't this be a `SiteDetail` type, as it is not a core site with the joined data? Or `CoreSiteWithDetails`?

mas-who: Yeah `CoreSiteWithDetail` sounds good. `SiteDetail` is reserved to represent a entry in the `site_details` table without any joins

internal/database/site_details.relation.go
mas-who: perhaps `CPUTotal`, `CPUUsed`, `MemoryTotal`, `MemoryUsed`, `DiskTotal`, `DiskUsed`?

mas-who: Actually, let's use your suggestions since we should also inform on the unit of measure in the variable names.

mas-who: I've updated the spec for this as well

scripts/generate_dummy_data.sh
mas-who: I've kept the script to generate 20 entries. Generating 200 entries takes way too long to be included in the package.json script. For local dev, we can use the script to generate data to test scale, but let's not include it in the build steps?

edlerd: As discussed, we should use bulk insert, that should be faster for 200 entries, than the single inserts for 20 entries currently in place.

scripts/generate_sites.sh
```diff
@@ -0,0 +1,37 @@
...
+# Insert 20 varied entries into the site_details table
+for i in $(seq 1 20)
+do
+    CORE_SITE_ID=$(( (i % 20) + 1 )) # Cycle through 1 to 20
+    STATUS="PENDING_APPROVAL"
+    if [ $((i % 2)) -eq 0 ]; then
+        STATUS="ACTIVE"
+    fi
+    INSTANCE_COUNT=$(( RANDOM % 10 + 1 ))
+    MEMBER_COUNT=$(( RANDOM % 5 + 1 ))
```
edlerd: Random will be tricky for e2e tests, can we use something more predictable? i.e. a count depending on the id. Also, let's add the error case and various site last updated timestamps.

edlerd (changes requested):
Thanks for the changes, this looks pretty good.

I think some refinements on the data generation script would be nice.

scripts/generate_sites.sh
mas-who: not sure what you mean by the error case? Each site upon insert into the db table gets a default `updated_at` set to the current timestamp.

scripts/generate_dummy_data.sh
mas-who: waaaay faster now for 200, basically instant.

scripts/generate_sites.sh
```diff
@@ -0,0 +1,47 @@
...
+done
+
+# Combine the values into a single insert statement
+SITE_DETAILS_INSERT+=$(IFS=,; echo "${SITE_DETAILS_VALUES[*]}")";"
+
+# Execute the combined SQL commands
+go run ./cmd/lxd-site-mgr --state-dir ./state/dir1 sql "
+    $CORE_SITES_INSERT
+    $SITE_DETAILS_INSERT
+"
```
edlerd:
```suggestion
"

```

edlerd (approved):
LGTM, thanks for adding :)

scripts/generate_sites.sh
edlerd: I was thinking we could have more variety in the instance status, so we have some instances in error status as well. But probably fine to keep it as is for now. If we need more specific test data, we can extend this script later on.

PR #33, opened 2024-07-03
add member specific configs endpoints [WD-12695]

mas-who:
## Done
- [x] add `Get` and `Patch` endpoints for `/1.0/member/:name/config`.
- [x] added a `GET /1.0/member/config` endpoint to get all member configs
- [ ] restart network listener if `https_address` is updated for a member (this is not possible to do at the moment in microcluster)

## Context
 - The [network listener spec](https://docs.google.com/document/d/1juo7BfQ8jmaUnGbMpm5T2_e0Dn6jhOmCi1lOEu_Cp2w/edit) for microcluster indicates network listener addresses will be stored in the state directory on each member and not in dqlite. This means first, we do not need to have `https_address` as a column in the `manager_configs` table and second, to read/modify `https_address` we will need to interact with the state directory instead of dqlite.

### DIFF
--- a/internal/api/member_configs.go
+++ b/internal/api/member_configs.go
@@ -1,0 +1,179 @@
+package api
+
+import (
+	"context"
+	"database/sql"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/url"
+
+	"github.com/canonical/lxd/lxd/response"
+	"github.com/canonical/microcluster/rest"
+	microTypes "github.com/canonical/microcluster/rest/types"
+	"github.com/canonical/microcluster/state"
+	"github.com/gorilla/mux"
+
+	"github.com/canonical/lxd-site-manager/internal/api/types"
+	"github.com/canonical/lxd-site-manager/internal/database"
+)
+
+var memberConfigCmd = rest.Endpoint{
+	Path:  "member/{name}/config",
+	Patch: rest.EndpointAction{Handler: memberConfigPatch, AllowUntrusted: true},
+	Get:   rest.EndpointAction{Handler: memberConfigGet, AllowUntrusted: true},
+}
+
+var memberConfigsCmd = rest.Endpoint{
+	Path: "member/config",
+	Get:  rest.EndpointAction{Handler: memberConfigsGet, AllowUntrusted: true},
+}
+
+// update existing member configs.
+func memberConfigPatch(s *state.State, r *http.Request) response.Response {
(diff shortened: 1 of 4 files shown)


internal/api/member_configs.go
```diff
@@ -0,0 +1,198 @@
...
+	"github.com/gorilla/mux"
+
+	"github.com/canonical/lxd-site-manager/internal/api/types"
+	"github.com/canonical/lxd-site-manager/internal/database"
+)
+
+var memberConfigCmd = rest.Endpoint{
+	Path:  "member/{name}/config",
+	Put:   rest.EndpointAction{Handler: memberConfigPut, AllowUntrusted: true},
+	Patch: rest.EndpointAction{Handler: memberConfigPatch, AllowUntrusted: true},
```
edlerd: I think we should remove put and rely on patch to have create/modify semantics. Patch will require the name to be present, and the other fields are optional. If they are set, they'll override existing values. If a DB entry exists it will be updated, otherwise created.

internal/api/member_configs.go
```diff
@@ -0,0 +1,198 @@
...
+
+	if err != nil {
+		return response.SmartError(err)
+	}
+
+	if len(dbConfigs) == 0 {
+		return response.NotFound(fmt.Errorf("Member not found"))
+	}
+
+	return response.SyncResponse(true, toMemberConfigsAPI(dbConfigs))
```
edlerd: We shouldn't return an array on a single get endpoint.

edlerd (changes requested):

internal/api/member_configs.go
mas-who: I decided to put in the `PUT` method because the `https_address` config is required for a member. Since every member must have the `https_address` set, we can't rely on `PATCH` alone. I think it would be a little weird to require `https_address` to be set in a `PATCH` endpoint no?

edlerd: I don't understand the problem with patch. When PATCH creates, you have to validate the address set. When PATCH updates, you will not change the address if it is omitted, if it is set it must not be empty. Can you clarify the problem?

mas-who: yes that logic is fine, my problem really originate from the db layer since I'm using the `INSERT OR REPLACE INTO` query. This avoids having to do a read before the write. To implement the above logic I'd have to do a read to identify if it's a creation or update, which is fine I guess since I don't expect this endpoint being hit that often :+1:

edlerd (approved):
LGTM, thanks for the adjustments.

PR #36, opened 2024-07-05
chore(demo) add configuration and build steps for demo server

edlerd:
# Done
- added Dockerfile to build demo server
- added entrypoint script to run the demo image
- added site.yaml to configur ethe demo image on running it

### DIFF
--- a/scripts/generate_sites.sh
+++ b/scripts/generate_sites.sh
@@ -40,8 +40,15 @@ done
 # Combine the values into a single insert statement
 SITE_DETAILS_INSERT+=$(IFS=,; echo "${SITE_DETAILS_VALUES[*]}")";"
 
+# if flag --local exists then use go run to run site-mgr
+if [ "$1" == "--local" ]; then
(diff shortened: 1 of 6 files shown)


scripts/generate_sites.sh
```diff
@@ -41,7 +41,7 @@ done
 SITE_DETAILS_INSERT+=$(IFS=,; echo "${SITE_DETAILS_VALUES[*]}")";"
 
 # Execute the combined SQL commands
-go run ./cmd/lxd-site-mgr --state-dir ./state/dir1 sql "
+./lxd-site-mgr --state-dir ./state/dir1 sql "
```
mas-who: this would cause running the script locally to fail unless if we compile the binaries. Maybe we can add a `--local` flag to this script so we can use `go run` or the binary itself if the flag is not specified?

edlerd: Yes, good idea. Will add the flag.

mas-who (approved):

mas-who:
LGTM :+1:

PR #37, opened 2024-07-06
fix: fix the vite dev server not loading assets

mas-who:
## Done
WIth SSL enabled for haproxy now, we need to configure the vite dev server to have https proxy target for ui assets. This PR fixes that. The fix is already in #36 but let's get this merged so long.

### DIFF
--- a/ui/vite.config.ts
+++ b/ui/vite.config.ts
@@ -9,7 +9,7 @@ export default defineConfig({
     port: 3000,
     proxy: {
       "/ui/assets": {
-        target: "http://0.0.0.0:8414/",
+        target: "https://0.0.0.0:8414/",
         rewrite: (path) => path.replace(/^\/ui/, ""),
         secure: false,
       },


edlerd (approved):
LGTM, Shall we add a test to ensure assets are loading fine as part of the E2e tests? Those rely on dotrun, I think we could ensure the favicon is loading there easily.

mas-who:
> LGTM, Shall we add a test to ensure assets are loading fine as part of the E2e tests? Those rely on dotrun, I think we could ensure the favicon is loading there easily.

Do you mean we can do some visual regression checks for the assets? It seems normal playwright assertion like the example below will pass regardless of the asset being correctly loaded or not.

```
test("logo", async ({ page }) => {
  await page.goto("/ui");
  const logo = page.locator(".l-navigation .p-panel__logo-image");
  await expect(logo).toBeVisible();
});
```

edlerd:
> Do you mean we can do some visual regression checks for the assets? It seems normal playwright assertion like the example below will pass regardless of the asset being correctly loaded or not.

We can request the favicon and compare to a screenshot.

mas-who:
Going to merge this PR so long, will create a separate PR for adding some screenshot tests. Here's the Jira [ticket](https://warthogs.atlassian.net/jira/software/c/projects/WD/boards/930?assignee=712020%3Accb7b61f-3f6a-41ec-9935-a79519dada4e&selectedIssue=WD-13063).

PR #38, opened 2024-07-08
add CLI support for managing manager and member configs [WD-12696]

mas-who:
## Done

- [x] added new CLI command for updating manager or member configs `lxd-site-mgr config set [<member>] key=val...`
- [x] updated `configs` file names to be `manager_configs` for clearer differentiation with `member_configs`
- [x] added new CLI command for showing member configs `lxd-site-mgr config show [<member>]`
- [x] added script for generating dummy configs


cmd/lxd-site-mgr/configs.go
```diff
@@ -0,0 +1,219 @@
...
+
+func (c *cmdConfigSet) command() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "set [<member>] <key>=<value>...",
+		Short: "Set member or cluster configuration keys",
+		Example: `
+			lxd-site-mgr config set [<member>] https_address=0.0.0.0:9001
+			Will set the https_address configuration key for the local member.
+
+			lxd-site-mgr config set global.address=179.0.10.1
```
edlerd: We should use some of the ips intended for docs, i.e. 203.0.113.1 as per https://datatracker.ietf.org/doc/html/rfc5737#section-3

scripts/generate_configs.sh
```diff
@@ -0,0 +1,70 @@
...
+    config set \
+        oidc.issuer=$OIDC_ISSUER \
+        oidc.client.id=$OIDC_CLIENT_ID \
+        oidc.audience=$OIDC_AUDIENCE
+
+# optionally populate manager global address
+if [ $POPULATE_GLOBAL_ADDRESS = true ]; then
+    lxd-site-mgr \
+        --state-dir ./state/dir1 \
+        config set global.address=179.10.10.1
```
edlerd: I think this address shouldn't be hard coded?

scripts/generate_configs.sh
edlerd, line 1: It seems this script has some options and some config comes via env file and some configs are hard coded. I think we should be more strict where things are configured here.

cmd/lxd-site-mgr/configs.go
mas-who: Interesting, thanks for sharing this

scripts/generate_configs.sh
mas-who: Will move everything into `.local.env` file instead since the oidc configs has to be in the env file. That way everything is aligned. If a specific env var isn't defined, then the script will skip generating related configs. This is useful for optional configs such as member `external_address`

scripts/generate_configs.sh
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env bash
+set -e
+
+alias lxd-site-mgr='go run ./cmd/lxd-site-mgr'
```
edlerd: This doesn't work, I get `scripts/generate_configs.sh: line 24: lxd-site-mgr: command not found`. Let's use the same approach with a `$COMMAND` as in the generate sites script.

edlerd (changes requested):

edlerd (approved):
LGTM

PR #43, opened 2024-07-09
chore: update cpu metrics data structure

mas-who:
## Done
 - replaced `cpu_usage` with `cpu_load_1`, `cpu_load_5` and `cpu_load_15`


internal/database/site_details.go
```diff
@@ -27,7 +27,9 @@ type SiteDetail struct {
 	Status            string `db:"primary=true"`
 	ID                int
 	CPUTotalCount     float64
-	CPUUsage          float64
+	CPULoad_1         float64
+	CPULoad_5         float64
+	CPULoad_15        float64
```
edlerd:
```suggestion
	CPULoad1         string
	CPULoad5         string
	CPULoad15        string
```
Avoid the underscore and string type?

edlerd (changes requested):

internal/database/site_details.go
mas-who: Yeah, took me a bit of time to figure out how to map the struct fields to correct column names since `CPULoad1` would map to `cpu_load1` in the database. It's all sorted now.

internal/database/site_details.go
```diff
@@ -27,7 +27,9 @@ type SiteDetail struct {
 	Status            string `db:"primary=true"`
 	ID                int
 	CPUTotalCount     float64
-	CPUUsage          float64
+	CPULoad1          float64 `db:"sql=site_details.cpu_load_1"`
+	CPULoad5          float64 `db:"sql=site_details.cpu_load_5"`
+	CPULoad15         float64 `db:"sql=site_details.cpu_load_15"`
```
edlerd: Shouldn't those be strings?

edlerd (approved):
LGTM, thanks :)

PR #44, opened 2024-07-09
sites: add POST /1.0/sites endpoint to receive join request from LXD [WD-12699]

mas-who:
## Done

- [ ] added `POST /1.0/sites` endpoint to receive LXD join request
- [ ] check token expiry
- [ ] verify HMAC
- [ ] if join request is valid i.e. token not expired and HMAC verification successful -> create site entry, create site detail and delete token

## Context
 - Currently the endpoint is registered on the site management listener. Once the microcluster [PR](https://github.com/canonical/microcluster/pull/174) for updating network listener functionalities gets merged, then we can move the endpoint to the control listener instead.

## Testing
1. `cd ui` -> `yarn backend-run` -> `yarn backend-init`
2. run `scripts/generate_site_join_token.sh`
3. copy the generated token, in a separate terminal use the token as input for `lxc site-manager join`
4. LXD should show "Joined site manager with token
5. Check site manager tables. No tokens in `core_site_tokens`. An entry is made in `core_sites` and `site_details`.


edlerd (approved):
Works like a charm. Also code LGTM

PR #45, opened 2024-07-10
Site status POST /1.0/site/status endpoint

edlerd:
# Done
- unify and fix db types
- add site status endpoint

# QA
- run `yarn backend-run`
- in 2nd terminal run `yarn backend-init`
- generate a site request `../scripts/generate_site_join_token.sh`
- build lxd and run it [from this branch](https://github.com/edlerd/lxd/tree/site-manager)
- use the token from the 3rd step in lxd `LXD_DIR=/var/lib/lxd lxc site-manager join $token`


internal/database/site_details.go
```diff
@@ -23,17 +23,17 @@ import (
 
 // SiteDetail represents all site level data.
 type SiteDetail struct {
-	CoreSiteID        int64  `db:"primary=true"`
-	Status            string `db:"primary=true"`
```
mas-who, line 27: we should remove `Status` from `SiteDetailFilter` as well

internal/api/types/site.go
```diff
@@ -46,3 +46,21 @@ const (
 	// ACTIVE is the status of a site once its join request has been approved.
 	ACTIVE SiteStatus = "ACTIVE"
 )
+
+type StatusDistribution struct {
```
mas-who: I think we can merge `StatusDistribution` and `Status` struct types in this file. They represent the same thing I believe.

internal/database/schema.go
```diff
@@ -40,14 +40,14 @@ func schemaUpdate1(ctx context.Context, tx *sql.Tx) error {
             core_site_id            INTEGER NOT NULL,
             joined_at               DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
             status                  TEXT NOT NULL CHECK(status IN ('PENDING_APPROVAL', 'ACTIVE')),
-            cpu_total_count         FLOAT NOT NULL DEFAULT 0,
+            cpu_total_count         int64 NOT NULL DEFAULT 0,
```
mas-who, line 43: why do we want to change the type for non-ID numeric fields? I think if not float, `int` should be fine? `int64` feels very specific

internal/api/sites.go
```diff
@@ -32,6 +36,136 @@ var siteCmd = rest.Endpoint{
 	Path:   "sites/{siteName}",
 	Get:    rest.EndpointAction{Handler: siteGet, AllowUntrusted: true},
 	Delete: rest.EndpointAction{Handler: siteDelete, AllowUntrusted: true},
+	// FIXME: this endpoint will be a resource for the control listener
+	Post: rest.EndpointAction{Handler: sitesStatusPost, AllowUntrusted: true},
```
mas-who, line 39: we will move this to the control listener using the approach in #47. Although I think we should use the `(rest).EndpointAction.AccessHandler` for the mtls authentication, similar to how we handle oidc authentication in #11

internal/api/sites.go
```diff
@@ -32,6 +36,136 @@ var siteCmd = rest.Endpoint{
...
+}
+
+func sitesStatusPost(s *state.State, r *http.Request) response.Response {
+	siteName, err := url.PathUnescape(mux.Vars(r)["siteName"])
+	if err != nil {
+		return response.SmartError(err)
+	}
+
+	if siteName != "status" {
```
mas-who, line 48: we won't need to do this once the endpoint is moved to the control listener I think.

internal/api/sites.go
```diff
@@ -32,6 +36,136 @@ var siteCmd = rest.Endpoint{
...
+		logger.Warn("Invalid endpoint")
+		return response.BadRequest(fmt.Errorf("invalid endpoint"))
+	}
+
+	if len(r.TLS.PeerCertificates) != 1 {
```
mas-who, line 58: I think we also should check for `r.TLS != nil` to be safe? This condition should result in an error response

internal/api/sites.go
```diff
@@ -32,6 +36,136 @@ var siteCmd = rest.Endpoint{
...
+			return err
+		}
+
+		for _, dbSite := range dbSites {
+			if dbSite.Status == string(types.PENDING_APPROVAL) {
+				continue
+			}
+
+			dbSiteCert, err := parseCertificate(dbSite.SiteCertificate)
+			if err != nil {
```
mas-who, line 77: we should add a warning log here probably since the stored cert in the db would be invalid

internal/api/sites.go
```diff
@@ -32,6 +36,136 @@ var siteCmd = rest.Endpoint{
...
+				continue
+			}
+
+			// check if public key of dbSite matches the peer certificate from the request
+			if bytes.Equal(dbSiteCert.Raw, peerCert.Raw) {
```
mas-who, line 88: nice this is how microcluster implements the trust check as well, so I think the approach is valid. Should we also check cert expiry?

internal/api/sites.go
```diff
@@ -32,6 +36,136 @@ var siteCmd = rest.Endpoint{
...
+	peerCert := r.TLS.PeerCertificates[0]
+
+	var siteID int64
+	err = s.Database.Transaction(r.Context(), func(ctx context.Context, tx *sql.Tx) error {
+		dbSites, err := database.GetCoreSitesWithDetails(ctx, tx)
+		if err != nil {
+			return err
+		}
+
+		for _, dbSite := range dbSites {
```
mas-who, line 71: do you think we might have scaling issues here with lots of sites?

internal/api/sites.go
```diff
@@ -32,6 +36,136 @@ var siteCmd = rest.Endpoint{
...
+
+	if err != nil {
+		logger.Warn("Failed to get site ID", logger.Ctx{"err": err})
+		return response.SmartError(err)
+	}
+
+	if siteID == 0 {
+		logger.Warn("Site not found", logger.Ctx{"site": siteName})
+		return response.NotFound(fmt.Errorf("site not found"))
+	}
```
mas-who, line 105: we should extract logic up until this point for tls checks into a `AccessHandler` as mentioned in an earlier comment

internal/api/sites.go
```diff
@@ -32,6 +36,136 @@ var siteCmd = rest.Endpoint{
...
+
+	payload := types.SiteStatusPost{}
+	err = json.NewDecoder(r.Body).Decode(&payload)
+	if err != nil {
+		return response.BadRequest(err)
+	}
+
+	err = s.Database.Transaction(r.Context(), func(ctx context.Context, tx *sql.Tx) error {
+		dbSite, err := database.GetSiteDetail(ctx, tx, siteID)
```
mas-who, line 114: I don't think we need this db call here. We can just construct a `database.SiteDetail` struct based on the payload and pass it to `database.UpdateSiteDetail`

internal/api/sites.go
```diff
@@ -32,6 +36,136 @@ var siteCmd = rest.Endpoint{
...
+
+		return nil
+	})
+
+	if err != nil {
+		logger.Warn("Failed to update site status", logger.Ctx{"site": siteID, "err": err})
+		return response.SmartError(err)
+	}
+
+	return response.EmptySyncResponse
```
mas-who, line 144: we need to respond with site manager addresses and next update timestamp here. Should probably setup a follow up issue in Jira for this?

internal/api/sites.go
```diff
@@ -32,6 +36,136 @@ var siteCmd = rest.Endpoint{
...
+
+	var total int64
+	for _, s := range statuses {
+		total += s.Count
+	}
+
+	return total, string(parsedStatuses)
+}
+
+func parseCertificate(pemStr string) (*x509.Certificate, error) {
```
mas-who: we can reuse the [ParseX509Certificate](https://github.com/canonical/microcluster/blob/fe0e5ecba28a2898ff81bbe71b60e87300b610da/rest/types/certificate.go#L23) function from microcluster for this, I believe it's exposed to us

internal/api/sites.go
```diff
@@ -158,8 +292,11 @@ func sitesPost(s *state.State, r *http.Request) response.Response {
 
 		// create relevant site details
 		_, err = database.CreateSiteDetail(ctx, tx, database.SiteDetail{
-			Status:     string(types.PENDING_APPROVAL),
-			CoreSiteID: siteID,
+			Status:           string(types.PENDING_APPROVAL),
+			CoreSiteID:       siteID,
+			JoinedAt:         time.Now(),
+			MemberStatuses:   "[]",
```
mas-who, line 292: we can rely on database default for `MemberStatuses` and `InstanceStatuses`

mas-who (commented):
Thanks for this @edlerd , left some comments :slightly_smiling_face:

internal/api/sites.go
edlerd: There might be fields we are not aware of (or that get added later). Loading the state from db and updating only the fields we care about ensures we do not remove such.

internal/api/sites.go
edlerd: I tested to remove those, and it resulted in null values from the JSON response on the `GET /1.0/sites`. Those were crashing the frontend.

internal/api/sites.go
edlerd: good idea!

internal/api/sites.go
edlerd: Yes, we should keep this in an in-memory struct and not load from DB each time. We still need to search for the right cert, I think comparing the bytes to find the matching cert should be fine.

internal/api/sites.go
edlerd: Yes, good idea.

internal/database/schema.go
edlerd: I prefer to be specific with `int64`, because that is what we get for most of those metrics from LXD. Relying on `int` could fail as the `int64` might or might not fit into an `int`. Depending on if `int` is interpreted as `int32` or `int64`. Float I'd avoid as it can cause issues in js or json world.

internal/api/sites.go
edlerd: Yes, agreed. Let's keep it as a follow up.

internal/database/schema.go
mas-who: right `int` adjust the size of the data based on system arch. If the numbers are large then we will have issues. Fair enough :+1:

internal/api/sites.go
mas-who: Yeah I've already created an [issue](https://warthogs.atlassian.net/jira/software/c/projects/WD/boards/930/backlog?epics=visible&selectedIssue=WD-12924&text=LXD-SM) for this next pulse. So let's resolve this now

internal/api/sites.go
mas-who: will do in follow up task

internal/api/sites.go
mas-who: should we remove this now since the network listener update is almost ready

internal/api/sites.go
mas-who: that's fair :+1:

internal/api/sites.go
```diff
@@ -32,6 +36,136 @@ var siteCmd = rest.Endpoint{
...
+
+	err = s.Database.Transaction(r.Context(), func(ctx context.Context, tx *sql.Tx) error {
+		dbSite, err := database.GetSiteDetail(ctx, tx, siteID)
+		if err != nil {
+			return err
+		}
+
+		dbSite.UpdatedAt = time.Now()
+		dbSite.DiskUsage = payload.DiskUsage
+		dbSite.MemoryUsage = payload.MemoryUsage
```
mas-who: just saw this now, `DiskUsage` and `MemoryUsage` seems to have been set twice.

internal/api/sites.go
mas-who: created follow up task for this

internal/api/sites.go
```diff
@@ -32,6 +35,133 @@ var siteCmd = rest.Endpoint{
...
+		if err != nil {
+			return err
+		}
+
+		for _, dbSite := range dbSites {
+			if dbSite.Status == string(types.PENDING_APPROVAL) {
+				continue
+			}
+
+			dbSiteCert, err := types2.ParseX509Certificate(dbSite.SiteCertificate)
```
mas-who: can we call this `microTypes` or something a bit more descriptive than `types2`?

internal/api/sites.go
edlerd: I'd keep it and only remove once the right listener is in place

internal/api/sites.go
edlerd: Whoops

mas-who:
Did some QA, initially thought it wasn't working but realised I had to remove the `PENDING_APPROVAL` check for now :sweat_smile: . It works nicely so LGTM, let's get this merged then I will make some minor related updates in #47 .

Leaving this screenshot here so we can appreciate the milestone!

Screenshot from 2024-07-12 09-05-19

mas-who (approved):

PR #46, opened 2024-07-11
feat: Implemented the LXD-SM Settings/Configuration Page [WD-12946]

Kxiru:
- Added Settings Page
- Added Setting table and form
- Added backend api query to both retrieve and patch manager level config settings.

image

### DIFF
--- a/ui/src/App.tsx
+++ b/ui/src/App.tsx
@@ -4,6 +4,7 @@ import { useQuery } from "@tanstack/react-query";
 import { queryKeys } from "util/queryKeys";
 import { fetchClusters } from "api/clusters";
 import NoMatch from "pages/NoMatch";
+import Settings from "pages/Settings";
 
 const ClusterList = lazy(() => import("pages/clusters/ClusterList"));
 const Login = lazy(() => import("pages/Login"));
@@ -40,6 +41,7 @@ const App: FC = () => {
         />
         <Route path="/ui/login" element={<Login />} />
         <Route path="/ui/sites" element={<ClusterList />} />
+        <Route path="/ui/settings" element={<Settings />} />
         <Route path="*" element={<NoMatch />} />
       </Routes>
     </Suspense>
--- a/ui/src/pages/Settings.tsx
+++ b/ui/src/pages/Settings.tsx
@@ -1,0 +1,83 @@
+import { FC } from "react";
+import { MainTable, Row } from "@canonical/react-components";
+import BaseLayout from "components/BaseLayout";
+import { queryKeys } from "util/queryKeys";
+import { fetchConfigOptions } from "api/settings";
+import { useQuery } from "@tanstack/react-query";
+import SettingForm from "./settings/SettingForm";
+
+const Settings: FC = () => {
+  const { data: configOptions } = useQuery({
+    queryKey: [queryKeys.configOptions],
+    queryFn: fetchConfigOptions,
+  });
+
+  const defaultConfig = {
+    "oidc.issuer": "",
+    "oidc.client.id": "",
+    "oidc.audience": "",
+    "global.address": "",
+  };
+
+  const headers = [
+    { content: "Scope", className: "scope" },
+    { content: "Key", className: "key" },
+    { content: "Value" },
+  ];
+
+  const configKeys = Object.keys(defaultConfig);
+  // const configKeys = Object.keys(defaultConfig);
+
+  const rows = configKeys.map((key, index) => {
+    return {
+      columns: [
+        {
+          content: (
+            <h2 className="p-heading--5">{index === 0 ? "Cluster" : ""}</h2>
+          ),
+          role: "cell",
+          className: "scope",
+          "aria-label": "Scope",
+        },
+        {
+          content: <div className="key-cell">{key}</div>,
+          role: "cell",
+          className: "key",
+          "aria-label": "Key",
+        },
+        {
+          content: (
+            <SettingForm
+              configField={key}
+              value={
+                configOptions?.config[key] ||
+                defaultConfig[key as keyof typeof defaultConfig]
+              }
+              isLast={index === length - 1}
+            />
+          ),
+          role: "cell",
+          "aria-label": "Value",
+          className: "u-vertical-align-middle",
+        },
+      ],
+    };
+  });
+
+  return (
+    <BaseLayout title="Settings">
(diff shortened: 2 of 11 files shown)


edlerd (commented):
Thank's for adding. I can't see any settings on the demo server. Does this rely on custom configuration to be present?
Can you ensure the config options are available also when the config is at default values?

Edit: i.e. this is the response of the `GET /1.0/config` api call on the demo server:

image

ui/src/pages/Settings.tsx
```diff
@@ -0,0 +1,72 @@
...
+    queryFn: fetchConfigOptions,
+  });
+
+  const headers = [
+    { content: "Scope", className: "scope" },
+    { content: "Key", className: "key" },
+    { content: "Value" },
+  ];
+
+  const configKeys = Object.keys(configOptions?.config || {});
```
edlerd: We could have a JSON object with the config options in the repository to rely on here. The API response is not reliable for this, as it will be missing any options with default values.

Kxiru:
> Thank's for adding. I can't see any settings on the demo server. Does this rely on custom configuration to be present? Can you ensure the config options are available also when the config is at default values?
> 
> Edit: i.e. this is the response of the `GET /1.0/config` api call on the demo server:
> 
> image

Yes, I have sent some information that will address this in Mattermost. That being said, I believe @mas-who is creating a PR to address this.

ui/src/pages/Settings.tsx
Kxiru: Apologies, I am not sure I understand. Please could you clarify what you mean?

edlerd: The configOptions can be empty and we will render an empty settings page as in the demo server. So instead of relying on the options from the api response, we can have the list hardcoded in the frontend, to ensure they are available in the settings page independent of the current configuration.

Kxiru: Ooh, okay sure, can do!

mas-who:
@Kxiru can you also add `"ui/settings"` to `uiServeRoutes` in `internal/api/ui.go` file?

Edit: currently if you try do a browser refresh while on the settings page, it will result in a 404 not found error from the backend because the path isn't added, this should be documented and I have planned a task for this next pulse

ui/src/App.tsx
```diff
@@ -40,6 +41,7 @@ const App: FC = () => {
         />
         <Route path="/ui/login" element={<Login />} />
         <Route path="/ui/sites" element={<ClusterList />} />
+        <Route path="/ui/settings" element={<Settings />} />
```
edlerd, line 44: This needs to be added to the `uiServeRoutes` in `ui.go` as well. So we can refresh / land directly on this URL.

ui/package.json
```diff
@@ -21,7 +21,7 @@
     "hooks-remove": "husky uninstall",
     "start": "./entrypoint 0.0.0.0:${PORT}",
     "backend-run": "cd .. && make && go run ./cmd/lxd-site-mgrd --state-dir ./state/dir1/",
-    "backend-init": "cd .. &&  go run ./cmd/lxd-site-mgr --state-dir ./state/dir1/ waitready && go run ./cmd/lxd-site-mgr --state-dir ./state/dir1/ init \"member1\" 0.0.0.0:9001 --bootstrap && scripts/generate_sites.sh --local && scripts/generate_configs.sh --local",
+    "backend-init": "cd .. &&  go run ./cmd/lxd-site-mgr --state-dir ./state/dir1/ waitready && go run ./cmd/lxd-site-mgr --state-dir ./state/dir1/ init \"member1\" 0.0.0.0:9001 --bootstrap && scripts/generate_sites.sh --local",
```
edlerd: I think this shouldn't be changed, or is there a reason for it?

ui/src/components/Animate.tsx
edlerd, line 1: This seems unused, can it be removed?

ui/src/pages/Settings.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const { data: configOptions } = useQuery({
+    queryKey: [queryKeys.configOptions],
+    queryFn: fetchConfigOptions,
+  });
+
+  const tempConfigs = {
+    "oidc.issuer": "https://dev-xjrvvfikbsv4jxn7.us.auth0.com/",
+    "oidc.client.id": "gpSHEK5SooxiKlFU1lx4xnWJMf78ydwy",
+    "oidc.audience": "https://dev-xjrvvfikbsv4jxn7.us.auth0.com/api/v2/",
+    "global.address": "0.0.0.0:8414",
```
edlerd: This should be called defaultConfig and the values should all be empty string.

ui/src/pages/Settings.tsx
```diff
@@ -0,0 +1,83 @@
...
+  };
+
+  const headers = [
+    { content: "Scope", className: "scope" },
+    { content: "Key", className: "key" },
+    { content: "Value" },
+  ];
+
+  // const configKeys = Object.keys(configOptions?.config || {});
```
edlerd:
```suggestion
```

ui/src/pages/settings/SettingForm.tsx
```diff
@@ -0,0 +1,90 @@
+import { FC, useEffect, useRef, useState } from "react";
+import { Button, Icon, useNotify } from "@canonical/react-components";
+import { updateSettings } from "api/settings";
+import { queryKeys } from "util/queryKeys";
+import { useQueryClient } from "@tanstack/react-query";
+import SettingFormInput from "./SettingFormInput";
+
+export const getConfigId = (key: string) => {
+  return key.replace(".", "___");
+};
```
edlerd: This function should not be exported and live next to the file it is used in.

ui/src/util/notifications.tsx
edlerd, line 1: This seems unused, can it be removed?

edlerd (changes requested):
Seems to work fine now, some comments below.

ui/package.json
Kxiru: Supposedly for the configs to be able to be edited in the demo server, the initial configs should not be generated. @mas-who , could you perhaps offer some clarification, here?

edlerd (approved):
LGTM, thanks for adding and improving the new config page.

PR #47, opened 2024-07-11
preparation for network listener update in microcluster [WD-13102]

mas-who:
## Done

- [x] setup control listener to listen on different address than core listener
- [x] moved `POST /1.0/sites` endpoint to control listener
- [x] initialise control listener in `PostBootstrap` and `PostJoin` hooks. In addition, the initialisation will add a entry in the `manager_member_configs` table for each member with `https_address` populated. This also means we can remove logic populating `https_address` in the `generate_configs.sh` script.
- [x] setup global site manager state to hold `microcluster.Microcluster`. We pass this state into some endpoints so we can access client functions e.g. for updating network listener addresses
- [x] updated `PATCH /1.0/member/{name}/config` endpoint so that we will forward the request to the correct member if `https_address` is present in the payload. We will still populate the data inside dqlite so that read operations will be much easier to handle (trading off with some data redundancy)
- [x] to ensure config data is synced between daemon local config and dqlite, added a revert function to rollback a network listener update if the corresponding db transaction fails

## Context
 - This PR is created in preparation for the microcluster [update](https://github.com/canonical/microcluster/pull/174) that allows modification to network listeners.
 - **Currently this PR won't work since dependencies are not yet updated.**

## Follow up task
 - cache control listener and external addresses in the daemon state. This will improve efficiency for status updates.

## Testing
 - setup cluster with multiple members
 - update `https_address` config for individual members with `lxd-site-mgr config set <member> https_address=<address>`. 
 - make sure it works exactly as before.
 - @edlerd it would be great if you can test the join process since `POST /1.0/sites` is on a different listener now. I've tested it locally already but I think it would be good if you can check it too.


go.mod
```diff
@@ -2,6 +2,9 @@ module github.com/canonical/lxd-site-manager
 
 go 1.22.4
 
+// TODO: remove this after testing
+replace github.com/canonical/microcluster => ../microcluster
```
mas-who: this will be removed after upstream merge

mas-who:
@edlerd @roosterfish this is now ready for review with upstream [PR](https://github.com/canonical/microcluster/pull/174) merged.

internal/api/server.go
```diff
@@ -34,9 +44,24 @@ var siteManagementListener = rest.Server{
...
+			},
+		},
+	},
+}
+
 // GetServers returns all the network listeners for site manager.
 func GetServers() map[string]rest.Server {
 	return map[string]rest.Server{
-		"management-listener": siteManagementListener,
+		string(ManagementListener): siteManagementListener,
```
roosterfish: I am wondering if we should make this a non `string` type already on microcluster side?
Because there are two specific server names `core` and `unix` which are special and cannot be set (you will get an error). 
Having a type with already defined constants might make it easier to reason about this restriction.

roosterfish (commented):
Looks great, nice to see that you have added the `revert` pattern.

internal/api/server.go
mas-who: Yes I think that makes sense! It will highlight there are restrictions to server names that we can set.

mas-who:
got a but in microcluster that would non-deterministically cause conflicting endpoint errors. Here is the fix for it https://github.com/canonical/microcluster/pull/180. Before that gets merged, we should not merge this PR since the build process may fail non-deterministically

internal/api/server.go
mas-who: created an [issue](https://github.com/canonical/microcluster/issues/184) for this.

internal/api/sites.go
```diff
@@ -27,29 +27,25 @@ import (
...
+
+var sitesControlCmd = rest.Endpoint{
+	Path: "sites",
 	Post: rest.EndpointAction{Handler: sitesPost, AllowUntrusted: true},
 }
 
+var sitesStatusCmd = rest.Endpoint{
+	Path: "sites/status",
+	Post: rest.EndpointAction{Handler: sitesStatusPost, AllowUntrusted: true},
+}
```
edlerd: I think we should split the control listener endpoints and the public ones into dedicated files. Keeping the code in one file is a bit of a trap for the future.

internal/api/server.go
```diff
@@ -4,37 +4,68 @@ import (
...
+type ListenerName string
+
+const (
+	// ManagementListener is the name for the listener for the management API.
+	ManagementListener ListenerName = "management-listener"
+	// ControlListener is the name for the listener for the control API.
+	ControlListener ListenerName = "control-listener"
+)
+
+func siteManagementListener(s *state.SiteManagerState) rest.Server {
```
edlerd, line 20:
```suggestion
func publicListener(s *state.SiteManagerState) rest.Server {
```
Better name?

internal/api/server.go
edlerd, line 1: Should we split this to have a dedicated file for each listener server?

internal/api/types/server.go
```diff
@@ -7,4 +7,6 @@ const (
 	APIVersionPrefix types.EndpointPrefix = "1.0"
 	// NoPrefix is the path prefix for any endpoints that should be located at server root.
 	NoPrefix types.EndpointPrefix = ""
+	// InternalEndpoint - all endpoints restricted to trusted servers.
```
edlerd:
```suggestion
	// InternalEndpoint is restricted to trusted servers.
```

edlerd (changes requested):

internal/api/sites.go
mas-who: great idea! I've split the endpoints into `sites_management.go` and `sites_control.go`

internal/api/server.go
mas-who: I would keep the naming consistent with the spec, should we update the spec to align the naming as well?

internal/api/server.go
mas-who: I think there aren't that many endpoints to make the file difficult to manage. Splitting listeners into separate files in this case seems to be a bit of an overkill and would only introduce unnecessary folder structure imo

internal/api/server.go
edlerd: I am not too opinionated on this change. Wdyt? If we change it, we should keep the spec updated, yes.

mas-who: let's keep it as is for now then. I think given we will most likely change `site` to `cluster` or something else, we might end up changing the listener names as well eventually. Updating this now may be a little premature

edlerd (changes requested):
demo build is failing due to the missing version.

image

mas-who:
> demo build is failing due to the missing version.
> 
> image

Interesting, will look into it. The version is set in the `make` target so something is wrong with the version environment variable setup.

edlerd (approved):
LGTM, thanks.

PR #49, opened 2024-07-15
cli for generating external site join token [WD-13155]

mas-who:
## Done
 - added CLI for generating external site join token

## QA
 - try `go run ./cmd/lxd-site-mgr --state-dir state/dir1 external-site-join-tokens add <site_name> [--expiry=duration]`
 - try `go run ./cmd/lxd-site-mgr --state-dir state/dir1 external-site-join-tokens show`
 - try `go run ./cmd/lxd-site-mgr --state-dir state/dir1 external-site-join-tokens revoke <site_name>`

### DIFF
--- a/cmd/lxd-site-mgr/external_site_join_tokens.go
+++ b/cmd/lxd-site-mgr/external_site_join_tokens.go
@@ -30,6 +30,9 @@ func (c *cmdExternalSiteJoinToken) command() *cobra.Command {
 	var cmdShow = cmdExternalSiteJoinTokenShow{common: c.common}
 	cmd.AddCommand(cmdShow.command())
 
+	var cmdRevoke = cmdExternalSiteJoinTokenRevoke{common: c.common}
+	cmd.AddCommand(cmdRevoke.command())
+
 	return cmd
 }
 
@@ -163,4 +166,46 @@ func (c *cmdExternalSiteJoinTokenShow) run(cmd *cobra.Command, args []string) er
 
 	sort.Sort(lxdCmd.SortColumnsNaturally(rows))
 	return lxdCmd.RenderTable(lxdCmd.TableFormatTable, headers, rows, tokens)
+}
+
+type cmdExternalSiteJoinTokenRevoke struct {
+	common *CmdControl
+}
+
+func (c *cmdExternalSiteJoinTokenRevoke) command() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "revoke <site_name>",
+		Short: "Revoke the external site join token with the given LXD site name.",
+		Example: `
+			lxd-site-mgr external-site-join-token revoke <site_name>
+			Will render the specific external site join token invalid.
+		`,
+		RunE: c.run,
+	}
+
+	return cmd
+}
+
+func (c *cmdExternalSiteJoinTokenRevoke) run(cmd *cobra.Command, args []string) error {
+	m, err := microcluster.App(microcluster.Args{
+		StateDir: c.common.FlagStateDir,
+		Verbose:  c.common.FlagLogVerbose,
+		Debug:    c.common.FlagLogDebug,
+		Version:  version.Version(),
+	})
+
+	if err != nil {
+		return err
+	}
+
+	cli, err := m.LocalClient()
+	if err != nil {
+		return err
+	}
+
+	if len(args) < 1 {
+		return cmd.Help()
+	}
+
+	return client.ExternalSiteJoinTokenDeleteCmd(cmd.Context(), cli, args[0])
 }
--- a/cmd/lxd-site-mgr/external_site_join_tokens.go
+++ b/cmd/lxd-site-mgr/external_site_join_tokens.go
@@ -1,8 +1,10 @@
 package main
 
 import (
+	"sort"
 	"time"
 
+	lxdCmd "github.com/canonical/lxd/shared/cmd"
 	"github.com/canonical/microcluster/microcluster"
 	"github.com/spf13/cobra"
 
@@ -24,6 +26,9 @@ func (c *cmdExternalSiteJoinToken) command() *cobra.Command {
 
 	var cmdAdd = cmdExternalSiteJoinTokenAdd{common: c.common}
 	cmd.AddCommand(cmdAdd.command())
+
+	var cmdShow = cmdExternalSiteJoinTokenShow{common: c.common}
+	cmd.AddCommand(cmdShow.command())
 
 	return cmd
 }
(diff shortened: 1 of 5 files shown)


cmd/lxd-site-mgr/external_site_join_tokens.go
```diff
@@ -0,0 +1,195 @@
...
+
+type cmdExternalSiteJoinTokenAdd struct {
+	common *CmdControl
+
+	flagExpiry time.Duration
+}
+
+func (c *cmdExternalSiteJoinTokenAdd) command() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "add <site_name>",
```
edlerd, line 51: Are we missing the expiry option here?

cmd/lxd-site-mgr/external_site_join_tokens.go
```diff
@@ -0,0 +1,195 @@
...
+		Short: "Revoke the external site join token with the given LXD site name.",
+		Example: `
+			lxd-site-mgr external-site-join-token revoke <site_name>
+			Will render the specific external site join token invalid.
+		`,
+		RunE: c.run,
+	}
+
+	return cmd
+}
```
edlerd, line 187: Should this be called `delete` instead of `revoke`?

scripts/generate_site_join_token.sh
```diff
@@ -1,7 +1,11 @@
...
-}' -X POST https://0.0.0.0:9001/1.0/external-site-join-token
+
+# if flag --local exists then use go run to run site-mgr
+if [ "$1" == "--local" ]; then
+    COMMAND="go run ./cmd/lxd-site-mgr"
+else
+    COMMAND="./lxd-site-mgr"
+fi
+
+$COMMAND --state-dir ./state/dir1 external-site-join-tokens add site1 -e 24h
```
edlerd:
```suggestion
$COMMAND --state-dir ./state/dir1 external-site-join-tokens add site1 -e 24h

```

cmd/lxd-site-mgr/external_site_join_tokens.go
```diff
@@ -0,0 +1,195 @@
...
+		Use:   "add <site_name>",
+		Short: "Add a new external join token for a LXD site.",
+		Example: `
+			lxd-site-mgr external-site-join-token add <site_name>
+			Will created an external site join token for a LXD site.
+		`,
+		RunE: c.run,
+	}
+
+	cmd.Flags().DurationVarP(&c.flagExpiry, "expiry", "e", 0, "Specify the duration (i.e. ns, us, ms, s, m or h) for the token to be valid")
```
edlerd: Not sure if `e` is a good shorthand, it might be ambiguous. Can we have no short version and only the long one?

cmd/lxd-site-mgr/external_site_join_tokens.go
mas-who: the flag is registered by `cmd.Flags()` down below. Do you mean we should add it to the example usage? There is help text for the flag already, but I can add it to the example if needed.

cmd/lxd-site-mgr/external_site_join_tokens.go
mas-who: I think it's probably better to keep the command consistent with lxd token related cli commands. So `revoke` instead of `delete`. I think to the user `revoke` also makes sense since `delete` is more of an operation we do at the database level

cmd/lxd-site-mgr/external_site_join_tokens.go
edlerd: I thought the Usage example also show optional flags within square brackets.

mas-who: This is the Usage for `lxc list`

Screenshot from 2024-07-17 19-08-15

The `[flags]` in the usage only references the Flags section. I don't think all the flags would show up in the usage or usage examples section?

mas-who: Screenshot from 2024-07-17 19-13-23

edlerd: This looks good, I was confused 🤔 over the flags not being mentioned before in the help at all. Makes much sense this way 👍

mas-who:
Thanks for the review @edlerd on this one. Do you have any additional comments before approving?

cmd/lxd-site-mgr/external_site_join_tokens.go
mas-who: Yeah thanks for bringing it up. It's good that we checked this to make sure behaviour aligns with LXD :slightly_smiling_face:

cmd/lxd-site-mgr/external_site_join_tokens.go
```diff
@@ -0,0 +1,195 @@
...
+			lxd-site-mgr external-site-join-token add <site_name>
+			Will created an external site join token for a LXD site.
+		`,
+		RunE: c.run,
+	}
+
+	cmd.Flags().DurationVar(&c.flagExpiry, "expiry", 0, "Specify the duration (i.e. ns, us, ms, s, m or h) for the token to be valid")
+
+	return cmd
+}
```
edlerd: This should print a response with the token to the cli.

edlerd (changes requested):
QA looks fine, but the add command should surface the token to the user. We should never show the secret or the token again after creation. So maybe show the token in the cli output together with a note to store it as it will not be possible to retrieve the token value again.

cmd/lxd-site-mgr/external_site_join_tokens.go
mas-who: yes of course, this is done. We should definitely do that in the UI as well, will leave a comment in the Jira task.

Screenshot from 2024-07-18 12-47-03

edlerd: Looks good in the screenshot. Can you push the fix for this or is this still pending further tests?

mas-who: oops forgot to push, done

cmd/lxd-site-mgr/external_site_join_tokens.go
```diff
@@ -0,0 +1,202 @@
...
+	"github.com/canonical/lxd-site-manager/version"
+)
+
+type cmdExternalSiteJoinToken struct {
+	common *CmdControl
+}
+
+func (c *cmdExternalSiteJoinToken) command() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "external-site-join-tokens",
```
edlerd:
```suggestion
		Use:   "external-site-join-token",
```
I think for consistency this should be singular

cmd/lxd-site-mgr/external_site_join_tokens.go
```diff
@@ -0,0 +1,202 @@
...
+	if err != nil {
+		return err
+	}
+
+	data, err := yaml.Marshal(tokens)
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("%s", data)
```
edlerd: I think tabular output like with `site list` is beneficial here.

edlerd (changes requested):

cmd/lxd-site-mgr/external_site_join_tokens.go
mas-who: 100%

cmd/lxd-site-mgr/external_site_join_tokens.go
```diff
@@ -0,0 +1,211 @@
...
+		Use:   "add <site_name>",
+		Short: "Add a new external join token for a LXD site.",
+		Example: `
+			lxd-site-mgr external-site-join-token add <site_name>
+			Will created an external site join token for a LXD site.
+		`,
+		RunE: c.run,
+	}
+
+	cmd.Flags().DurationVar(&c.flagExpiry, "expiry", 0, "Specify the duration (i.e. ns, us, ms, s, m or h) for the token to be valid")
```
edlerd:
```suggestion
	cmd.Flags().DurationVar(&c.flagExpiry, "expiry", 0, "Specify the duration (i.e. 5m or 48h) for the token to be valid")
```
Probably better to avoid the small units here and give more reasonable examples?

edlerd (dismissed):
LGTM, one more idea to improve the inline help a bit, also ready to merge as is.

cmd/lxd-site-mgr/external_site_join_tokens.go
mas-who: thanks, updated as per suggestion :)

mas-who:
@edlerd I think this one is ready for approval :)

edlerd (approved):
LGTM, thanks for the iterations!

PR #50, opened 2024-07-15
tests: visual regression for checking logo asset [WD-13063]

mas-who:
## Done
 - visual regression test for logo asset


ui/package.json
```diff
@@ -22,7 +22,8 @@
     "start": "./entrypoint 0.0.0.0:${PORT}",
     "backend-run": "cd .. && make && go run ./cmd/lxd-site-mgrd --state-dir ./state/dir1/",
     "backend-init": "cd .. &&  go run ./cmd/lxd-site-mgr --state-dir ./state/dir1/ waitready && go run ./cmd/lxd-site-mgr --state-dir ./state/dir1/ init \"member1\" 0.0.0.0:9001 --bootstrap && scripts/generate_sites.sh --local && scripts/generate_configs.sh --local",
-    "test-e2e": "npx playwright test"
+    "test-e2e": "npx playwright test",
+    "test-e2e-screenshots": "npx playwright test --update-snapshots"
```
edlerd:
```suggestion
    "test-e2e-update": "npx playwright test --update-snapshots"
```
Maybe clearer?

ui/tests/screenshot/logo.spec.ts-snapshots/logo-screenshot-1-chromium-linux.png
edlerd, line 1: Can we change this to have the same assets for chromium and firefox?

edlerd (changes requested):

ui/tests/screenshot/logo.spec.ts-snapshots/logo-screenshot-1-chromium-linux.png
mas-who: done :+1:

edlerd (approved):
Code LGTM, test needs changes to become green though.

mas-who:
@edlerd so looking at the CI failure, the chromium logo test is failing with a 7% pixel difference when compared to the screenshot. This is probably due to the screenshot being taken in firefox. Looking at the screenshots comparison, it almost feels like there is some kind of colour difference between firefox and chrome? Not sure if using the same screenshot for cross platform projects may result in these small differences all the time. We could try fix the pixel difference somehow, set a threshold for pixel difference in testing or go back to per project screenshots. wdyt? 

900b5bcc611bb6742af2ea338d68c11f2800f537
62e3a0a6dd73ce2473cd1f18083874bef96aced2
176ac2a1ffc9e9c733b4c54dcc629face4dd9a13

edlerd:
> @edlerd so looking at the CI failure, the chromium logo test is failing with a 7% pixel difference when compared to the screenshot. This is probably due to the screenshot being taken in firefox. Looking at the screenshots comparison, it almost feels like there is some kind of colour difference between firefox and chrome? Not sure if using the same screenshot for cross platform projects may result in these small differences all the time. We could try fix the pixel difference somehow, set a threshold for pixel difference in testing or go back to per project screenshots. wdyt?

I see, thanks for figuring this out. Maybe easier in that case to revert to the previous case of having dedicated screenshots for each browser?

edlerd (approved):
LGTM, thanks for adding and figuring out the snapshot diff.

PR #52, opened 2024-07-16
feat: [WD-13153] Added cluster form and Join Token creation functiona…

Kxiru:
…lity

**Work Done:**
- Added a new Create Cluster / Token page.
- Added url / CreateCluster component to App.tsx
- Added an API call to create a Token.
- A Formik form to accept Token attributes.
- Minimal styling.
- Input validation
- Added a Version status bar at the bottom of the page.

image

image

Minimal Input Validation:
image
image

### DIFF
--- a/ui/src/App.tsx
+++ b/ui/src/App.tsx
@@ -3,6 +3,7 @@ import { Navigate, Route, Routes, useLocation } from "react-router-dom";
 import NoMatch from "pages/NoMatch";
 import Settings from "pages/Settings";
 import { useAuth } from "context/auth";
+import CreateCluster from "pages/clusters/CreateCluster";
 
 const ClusterList = lazy(() => import("pages/clusters/ClusterList"));
 const Login = lazy(() => import("pages/Login"));
@@ -33,6 +34,7 @@ const App: FC = () => {
         <Route path="/ui/login" element={<Login />} />
         <Route path="/ui/sites" element={<ClusterList />} />
         <Route path="/ui/sites/:activeTab" element={<ClusterList />} />
+        <Route path="/ui/sites/create" element={<CreateCluster />} />
         <Route path="/ui/settings" element={<Settings />} />
         <Route path="*" element={<NoMatch />} />
       </Routes>
--- a/ui/src/components/ScrollableContainer.tsx
+++ b/ui/src/components/ScrollableContainer.tsx
@@ -1,0 +1,53 @@
+import { DependencyList, FC, ReactNode, useEffect, useRef } from "react";
+import useEventListener from "@use-it/event-listener";
+import {
+  getAbsoluteHeightBelowById,
+  getParentsBottomSpacing,
+} from "util/helpers";
+import classnames from "classnames";
+
+interface Props {
+  children: ReactNode;
+  dependencies: DependencyList;
+  belowIds?: string[];
+  className?: string;
+}
+
+const ScrollableContainer: FC<Props> = ({
+  dependencies,
+  children,
+  belowIds = ["status-bar"],
+  className,
+}) => {
+  const ref = useRef<HTMLDivElement>(null);
+
+  const updateChildContainerHeight = () => {
+    const childContainer = ref.current?.children[0];
+    if (!childContainer) {
+      return;
+    }
+
+    const above = childContainer.getBoundingClientRect().top + 1;
+    const below = belowIds.reduce(
+      (acc, belowId) => acc + getAbsoluteHeightBelowById(belowId),
+      0,
+    );
+    const parentsBottomSpacing = getParentsBottomSpacing(childContainer);
+    const offset = Math.ceil(above + below + parentsBottomSpacing);
+    const style = `height: calc(100vh - ${offset}px); min-height: calc(100vh - ${offset}px)`;
+    childContainer.setAttribute("style", style);
+  };
+
+  useEventListener("resize", updateChildContainerHeight);
+  useEffect(updateChildContainerHeight, [...dependencies, ref]);
+
+  return (
+    <div ref={ref} className={classnames("scrollable-container", className)}>
+      <div id="content-details" className="content-details">
+        {children}
+      </div>
+    </div>
+  );
+};
+
+export default ScrollableContainer;
--- a/ui/src/components/ScrollableForm.tsx
+++ b/ui/src/components/ScrollableForm.tsx
@@ -1,0 +1,21 @@
+import { FC, ReactNode } from "react";
+import ScrollableContainer from "./ScrollableContainer";
+import { useNotify } from "@canonical/react-components";
+
+interface Props {
+  children: ReactNode;
+}
+
+const ScrollableForm: FC<Props> = ({ children }) => {
+  const notify = useNotify();
+  return (
+    <ScrollableContainer
+      dependencies={[notify.notification]}
+      belowIds={["status-bar", "form-footer"]}
+    >
+      {children}
+    </ScrollableContainer>
+  );
+};
+
+export default ScrollableForm;
--- a/ui/src/components/StatusBar.tsx
+++ b/ui/src/components/StatusBar.tsx
@@ -1,0 +1,16 @@
+import { FC } from "react";
+import { AppStatus } from "@canonical/react-components";
+
+interface Props {
+  className?: string;
+}
+
+const StatusBar: FC<Props> = () => {
+  return (
+    <AppStatus className="status-bar" id="status-bar">
+      <span className="server-version p-text--small">Version 0.1</span>
+    </AppStatus>
+  );
+};
+
+export default StatusBar;
--- a/ui/src/pages/clusters/ApproveClusterButton.tsx
+++ b/ui/src/pages/clusters/ApproveClusterButton.tsx
@@ -28,6 +28,7 @@ const ApproveClusterButton: FC<Props> = ({ clusterName }) => {
 
   return (
     <ActionButton
+      className="u-no-margin--bottom"
       onClick={() => void handleApproveCluster()}
       appearance="positive"
       loading={isLoading}
--- a/ui/src/pages/clusters/ClusterCreateDetailsForm.tsx
+++ b/ui/src/pages/clusters/ClusterCreateDetailsForm.tsx
@@ -1,0 +1,58 @@
+import { FC } from "react";
+import { Col, Input, Row } from "@canonical/react-components";
+import { FormikProps } from "formik/dist/types";
+import ScrollableForm from "components/ScrollableForm";
+import { convertToISOFormat } from "util/helpers";
+
+export interface CreateClusterFormValues {
+  site_name?: string;
+  expiry?: string;
+}
+
+export const newTokenPayload = (values: CreateClusterFormValues) => {
+  const payload: Record<string, string | undefined> = {
+    site_name: values.site_name,
+    expiry: convertToISOFormat(values.expiry ?? ""),
+  };
+
+  return payload;
+};
+
+interface Props {
+  formik: FormikProps<CreateClusterFormValues>;
+}
+
+const ClusterCreateDetailsForm: FC<Props> = ({ formik }) => {
+  return (
+    <ScrollableForm>
+      <Row>
+        <Col size={12}>
+          <Input
+            id="name"
+            name="site_name"
+            type="text"
+            label="Site name"
+            placeholder="Enter Name"
+            onBlur={formik.handleBlur}
+            onChange={formik.handleChange}
+            value={formik.values.site_name}
+            error={formik.touched.site_name ? formik.errors.site_name : null}
+          />
+          <Input
+            id="expiry"
+            name="expiry"
+            type="datetime-local"
+            label="Expiry Date"
+            placeholder="Enter Date"
+            onBlur={formik.handleBlur}
+            onChange={formik.handleChange}
+            value={formik.values.expiry}
+            error={formik.touched.expiry ? formik.errors.expiry : null}
+          />
+        </Col>
+      </Row>
+    </ScrollableForm>
+  );
+};
+
+export default ClusterCreateDetailsForm;
--- a/ui/src/pages/clusters/CreateCluster.tsx
+++ b/ui/src/pages/clusters/CreateCluster.tsx
@@ -1,0 +1,91 @@
+import { FC } from "react";
+import {
+  ActionButton,
+  Button,
+  Col,
+  Form,
+  Row,
+} from "@canonical/react-components";
+import BaseLayout from "components/BaseLayout";
+import { useNavigate } from "react-router-dom";
+import { useFormik } from "formik";
+import ClusterCreateDetailsForm, {
+  CreateClusterFormValues,
+  newTokenPayload,
+} from "./ClusterCreateDetailsForm";
+import FormFooterLayout from "components/forms/FormFooterLayout";
+import { createToken } from "api/tokens";
+import * as Yup from "yup";
+import { queryKeys } from "util/queryKeys";
+import { useQueryClient } from "@tanstack/react-query";
+import { getDefaultExpiryDate } from "util/createCluster";
+
+const CreateCluster: FC = () => {
+  const navigate = useNavigate();
+  const queryClient = useQueryClient();
+
+  const ClusterSchema = Yup.object().shape({
+    site_name: Yup.string().required("Site name is required"),
+    expiry: Yup.date()
+      .optional()
+      .min(new Date(), "Expiry date cannot be in the past"),
+  });
+
+  const submit = (values: CreateClusterFormValues) => {
+    const tokenPayload = newTokenPayload(values);
+
+    createToken(JSON.stringify(tokenPayload))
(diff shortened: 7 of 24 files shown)


ui/src/pages/clusters/CreateCluster.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const getPayload = (values: CreateClusterFormValues) => {
+    return {
+      ...newTokenPayload(values),
+    };
+  };
+
+  const formik = useFormik<CreateClusterFormValues>({
+    initialValues: {
+      site_name: "testsitename",
+      expiry: "2029-07-16T18:06:48Z",
```
edlerd: The default format is not valid. I think this should 24 hours from now.

ui/src/pages/clusters/CreateCluster.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+  const getPayload = (values: CreateClusterFormValues) => {
+    return {
+      ...newTokenPayload(values),
+    };
+  };
+
+  const formik = useFormik<CreateClusterFormValues>({
+    initialValues: {
+      site_name: "testsitename",
```
edlerd:
```suggestion
      site_name: "",
```

ui/src/pages/clusters/CreateCluster.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+  const submit = (values: CreateClusterFormValues) => {
+    const newTokenPayload = getPayload(values);
+    navigate(`/ui/sites/tokens`);
+
+    createToken(JSON.stringify(newTokenPayload)).catch((e: Error) => {
+      if (e.message === "Unable to create new Token") {
+        return;
+      }
+    });
```
edlerd:
```suggestion
    createToken(JSON.stringify(newTokenPayload))
    .then(() => {
      navigate(`/ui/sites/tokens`);
    })
    .catch((e: Error) => {
      if (e.message === "Unable to create new Token") {
        return;
      }
    });
```
We should only navigate once the creation is finished.
Also should add a `.finally` that is invalidating the token cache so we can see the newly created token immediately

ui/src/pages/clusters/CreateCluster.tsx
```diff
@@ -0,0 +1,89 @@
...
+  };
+
+  const getPayload = (values: CreateClusterFormValues) => {
+    return {
+      ...newTokenPayload(values),
+    };
+  };
```
edlerd: I think this can be done inline in `submit`, no need for an extra function.

mas-who:
Some minor comments from QA:
 1. The buttons at the bottom of the form should align with the edge of the form 
Screenshot from 2024-07-17 17-57-37
**See LXD-UI for ref**
Screenshot from 2024-07-17 18-06-59

2. I think we should do validation for the expiry date that it can't be set to in the past.

ui/src/App.tsx
```diff
@@ -33,6 +34,7 @@ const App: FC = () => {
         <Route path="/ui/login" element={<Login />} />
         <Route path="/ui/sites" element={<ClusterList />} />
         <Route path="/ui/sites/:activeTab" element={<ClusterList />} />
+        <Route path="/ui/sites/create" element={<CreateCluster />} />
```
mas-who, line 37: please add this path to `internal/api/ui.go` file in the `uiServeRoutes` array

ui/src/pages/clusters/CreateCluster.tsx
```diff
@@ -0,0 +1,89 @@
...
+} from "./ClusterCreateDetailsForm";
+import FormFooterLayout from "components/forms/FormFooterLayout";
+import { createToken } from "api/tokens";
+import * as Yup from "yup";
+
+const CreateCluster: FC = () => {
+  const navigate = useNavigate();
+
+  const ClusterSchema = Yup.object().shape({
+    expiry: Yup.date().optional(),
```
mas-who: validate expiry not in the past

ui/src/pages/clusters/CreateCluster.tsx
```diff
@@ -0,0 +1,89 @@
...
+  };
+
+  const formik = useFormik<CreateClusterFormValues>({
+    initialValues: {
+      site_name: "testsitename",
+      expiry: "2029-07-16T18:06:48Z",
+    },
+    validationSchema: ClusterSchema,
+    onSubmit: (values) => {
+      console.log("test submission worked");
```
mas-who: remember to remove after testing

ui/src/pages/clusters/CreateCluster.tsx
```diff
@@ -0,0 +1,89 @@
...
+      submit(values);
+    },
+  });
+
+  return (
+    <BaseLayout
+      title="Create a new join Token"
+      contentClassName="create-cluster"
+    >
+      <Row>
```
mas-who: I don't think this `Row` is necessary here.

ui/src/pages/clusters/CreateCluster.tsx
```diff
@@ -0,0 +1,89 @@
...
+        <Form onSubmit={formik.handleSubmit} className="form">
+          <Row className="form-contents">
+            <Col size={12}>
+              {/* <NotificationRow /> */}
```
mas-who: remove comment for now. This will be related to notifications in a later PR

ui/src/sass/_forms.scss
```diff
@@ -0,0 +1,28 @@
...
+
+      .content-details {
+        max-width: 54rem;
+      }
+    }
+  }
+
+  .bottom-btns {
+    margin-top: 1rem;
+    max-width: 72rem;
```
mas-who: this may need to be tweaked to get buttons align with the form

ui/src/components/ScrollableContainer.tsx
```diff
@@ -0,0 +1,53 @@
...
+  children: ReactNode;
+  dependencies: DependencyList;
+  belowIds?: string[];
+  className?: string;
+}
+
+const ScrollableContainer: FC<Props> = ({
+  dependencies,
+  children,
+  belowIds = ["status-bar"],
```
mas-who, line 19: "status-bar" is a very specific id in lxd-ui. Let's remove this and default `belowIds` to just an empty list

ui/src/components/ScrollableForm.tsx
```diff
@@ -0,0 +1,21 @@
...
+interface Props {
+  children: ReactNode;
+}
+
+const ScrollableForm: FC<Props> = ({ children }) => {
+  const notify = useNotify();
+  return (
+    <ScrollableContainer
+      dependencies={[notify.notification]}
+      belowIds={["status-bar", "form-footer"]}
```
mas-who, line 14: remove `status-bar`

ui/src/pages/clusters/CreateCluster.tsx
mas-who: I think in the success case, we also need to invalidate the `tokens` query key

mas-who (commented):
Thanks for implementing this. Left some feedback :)

ui/src/components/ScrollableContainer.tsx
edlerd: @mas-who is raising an interesting point. We might want to add a status bar that is displaying the version. Even though we don't have a version yet. We can hardcode "Version 0.1" for now.

Kxiru: Done!

Kxiru:
Added a status bar at the bottom! So cute haha.
image

ui/src/components/StatusBar.tsx
```diff
@@ -0,0 +1,21 @@
+import { FC } from "react";
+import classnames from "classnames";
+
+interface Props {
+  className?: string;
+}
+
+const StatusBar: FC<Props> = ({ className }) => {
```
mas-who: Let's keep things consistent and use [AppStatus](https://canonical.github.io/react-components/?path=/docs/components-applicationlayout-appstatus--docs) from react-components.

ui/src/pages/clusters/CreateCluster.tsx
```diff
@@ -0,0 +1,99 @@
...
+  const ClusterSchema = Yup.object().shape({
+    site_name: Yup.string().required("Site name is required"),
+    expiry: Yup.date()
+      .optional()
+      .min(new Date(), "Expiry date cannot be in the past"),
+  });
+
+  const submit = (values: CreateClusterFormValues) => {
+    const tokenPayload = newTokenPayload(values);
+    navigate(`/ui/sites/tokens`);
```
mas-who: this navigate should be removed. We should only navigate away from this page in the success case

ui/src/pages/clusters/CreateCluster.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+    createToken(JSON.stringify(tokenPayload))
```
mas-who, line 37: for future reference, the success and failure case will need notifications in the follow up PR

ui/src/pages/clusters/CreateCluster.tsx
```diff
@@ -0,0 +1,99 @@
...
+        }
+      })
+      .finally(() => {
+        void queryClient.invalidateQueries({
+          queryKey: [queryKeys.tokens],
+        });
+      });
+  };
+
+  const getExpiryDate = () => {
```
mas-who: for functions that do not handle state, it is better to define them outside of components. Otherwise functions gets redefined unnecessarily per render cycle. We can follow the folder structure in lxd-ui here. Create a file in `/util` to hold helper functions related specifically for creating tokens e.g. `/util/createCluster.tsx`. Also I think this function can be renamed to `getDefaultExpiryDate`

ui/src/pages/clusters/CreateCluster.tsx
```diff
@@ -0,0 +1,99 @@
...
+    return now.toISOString();
+  };
+
+  const formik = useFormik<CreateClusterFormValues>({
+    initialValues: {
+      site_name: "",
+      expiry: getExpiryDate(),
+    },
+    validationSchema: ClusterSchema,
+    onSubmit: (values) => {
```
mas-who: when passing a value in the form of `(input) => func(input)`, you can just pass `func` instead

ui/src/pages/clusters/CreateCluster.tsx
```diff
@@ -0,0 +1,99 @@
...
+      </Form>
+      <FormFooterLayout>
+        <Button appearance="base" onClick={() => navigate("/")}>
+          Cancel
+        </Button>
+        <ActionButton
+          appearance="positive"
+          loading={formik.isSubmitting}
+          disabled={!formik.isValid}
+          onClick={() => submit(formik.values)}
```
mas-who, line 82: We should remove this `onClick`. and set the `type` attribute for this button to be `submit`. The form will handle the submission logic itself. Also this should prevent potentially sending multiple requests to the backend.

ui/src/sass/styles.scss
```diff
@@ -47,3 +50,9 @@ body {
 .l-main .row {
   max-width: none;
 }
+
+.p-bottom-controls {
```
mas-who: what's the context for `p-bottom-controls`, why is this needed?

mas-who:
Left some more comments, nice one implementing the status bar and getting the form buttons aligned :slightly_smiling_face:

ui/src/components/StatusBar.tsx
Kxiru: I used StatusBar from LXD-UI, is AppStatus what it should have been?

mas-who: yeah status bar has the `l-status` class which belongs to the app layout family of things

ui/src/pages/clusters/CreateCluster.tsx
Kxiru: Do you mean it should be structured like this:

``` 
onSubmit: submit(values);
```

I think onSubmit needs to have an arrow function as before. Please could you clarify?

mas-who: just `onSubmit: submit` should be fine

Kxiru:
Reviews all addressed :+1:

ui/src/components/StatusBar.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { AppStatus } from "@canonical/react-components";
+
+interface Props {
+  className?: string;
+}
+
+const StatusBar: FC<Props> = () => {
+  return (
+    <AppStatus className="status-bar" id="status-bar">
+      <aside>
```
mas-who: This `aside` isn't needed. `AppStatus` wraps children with an `aside` already

Kxiru: Thanks for the heads up :)

ui/src/pages/clusters/ClusterCreateDetailsForm.tsx
```diff
@@ -0,0 +1,58 @@
...
+          />
+          <Input
+            id="expiry"
+            name="expiry"
+            type="datetime-local"
+            label="Expiry Date"
+            placeholder="Enter Date"
+            onBlur={formik.handleBlur}
+            onChange={formik.handleChange}
+            value={formik.values.expiry}
```
edlerd, line 49: I think there is something off with this. The initial value is not displayed here. Might be a format mismatch. However, the initial value is sent in the payload on form submit and changing the date/time also is reflected in the payload. So it is solely about reflecting the initial value as an initial state of this input.

edlerd (dismissed):
Found an issue with the initial state of the expiry.

One other idea now that we have token creation is to add an action column to the token table and add a "revoke" button that hits the token `DELETE` endpoint. We can also keep that as a follow up, though.

Kxiru:
> Found an issue with the initial state of the expiry.
> 
> One other idea now that we have token creation is to add an action column to the token table and add a "revoke" button that hits the token `DELETE` endpoint. We can also keep that as a follow up, though.

Thanks for the comments. I also implemented the revoke button and API really quickly. Here it is!

image

ui/src/pages/clusters/ClusterTokenActions.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+const ClusterTokenActions: FC<Props> = ({ token }) => {
+  const navigate = useNavigate();
+  const queryClient = useQueryClient();
+
+  return (
+    <ActionButton
+      className={"u-no-margin"}
+      appearance="negative"
+      onClick={() => {
```
mas-who: it's usually a good idea to separate logic and presentation since TS mixed with component HTML is difficult to read. Would you please extract this callback into a `handleDeleteToken` function and pass it to the `onClick`?

ui/src/pages/clusters/ClusterTokenActions.tsx
```diff
@@ -0,0 +1,43 @@
...
+        deleteToken(token.site_name)
+          .then(() => {
+            navigate(`/ui/sites/tokens`);
```
mas-who: I don't think we need to do a navigate after deleting the token. Also I'd remove the `finally` block and keep the query invalidation in the success case.

ui/src/pages/clusters/ClusterTokenActions.tsx
Kxiru: Done :)

ui/src/pages/clusters/ApproveClusterButton.tsx
```diff
@@ -28,6 +28,7 @@ const ApproveClusterButton: FC<Props> = ({ clusterName }) => {
 
   return (
     <ActionButton
+      className={"u-no-margin--bottom"}
```
edlerd:
```suggestion
      className="u-no-margin--bottom"
```

ui/src/pages/clusters/ClusterTokenActions.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { ActionButton } from "@canonical/react-components";
+import { deleteToken } from "api/tokens";
+import { Token } from "types/token";
+import { queryKeys } from "util/queryKeys";
+
+interface Props {
+  token: Token;
+}
+
+const ClusterTokenActions: FC<Props> = ({ token }) => {
```
edlerd:
```suggestion
const RevokeTokenButton: FC<Props> = ({ token }) => {
```
Maybe a better name for this component and file?

ui/src/pages/clusters/DeleteClusterButton.tsx
```diff
@@ -28,6 +28,7 @@ const DeleteClusterButton: FC<Props> = ({ clusterName }) => {
 
   return (
     <ActionButton
+      className={"u-no-margin--bottom"}
```
edlerd:
```suggestion
      className="u-no-margin--bottom"
```

mas-who (dismissed):

mas-who:
LGTM :+1:

mas-who (approved):

edlerd (approved):
LGTM :+1:

PR #53, opened 2024-07-16
ui: add member configs to settings page [WD-12947]

mas-who:
## Done
- Added member configs to settings page
- Allow reset external address for members
- Use correct client for updating member configs. Previously a bug was introduced where we'd use a local client to get a remote member's server configs for updating the same entry in the database.

## Context
With #47 merged, it is now guaranteed that every member in the cluster will have configs in the `manager_member_configs` table during initialisation. So member configs will show in the settings page by default with its `https_address` setting.

### DIFF
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -4,6 +4,8 @@ import (
 	"net/http"
 
 	"github.com/canonical/lxd/lxd/response"
+	"github.com/canonical/microcluster/client"
+	"github.com/canonical/microcluster/rest/access"
 	microState "github.com/canonical/microcluster/state"
 
 	"github.com/canonical/lxd-site-manager/internal/api/types"
@@ -27,6 +29,16 @@ func authHandler(siteManagerState *state.SiteManagerState) types.AccessHandler {
 
 		_, resp, err := siteManagerState.OIDCVerifier.Auth(r.Context(), r)
 		if err != nil {
+			// check if the request is a cluster notification with valid cert
+			if client.IsNotification(r) && r.TLS != nil {
+				hostAddress := clusterState.Address().URL.Host
+				trustedCerts := clusterState.Remotes().CertificatesNative()
+				trusted, _ := access.Authenticate(clusterState, r, hostAddress, trustedCerts)
+				if trusted {
+					return true, nil
+				}
+			}
+
 			return false, resp
 		}
 
--- a/internal/api/member_configs.go
+++ b/internal/api/member_configs.go
@@ -63,12 +63,12 @@ func memberConfigPatch(siteManagerState *state.SiteManagerState) types.EndpointH
 			return response.BadRequest(err)
 		}
 
-		if payload.HTTPSAddress == "" && payload.ExternalAddress == "" {
+		if payload.HTTPSAddress == nil && payload.ExternalAddress == nil {
 			return response.BadRequest(fmt.Errorf("no fields provided to update"))
 		}
 
-		if payload.ExternalAddress != "" {
-			_, err = microTypes.ParseAddrPort(payload.ExternalAddress)
+		if payload.ExternalAddress != nil && *payload.ExternalAddress != "" {
+			_, err = microTypes.ParseAddrPort(*payload.ExternalAddress)
 			if err != nil {
 				return response.BadRequest(fmt.Errorf("invalid external_address for member %q: %w", memberName, err))
 			}
@@ -82,8 +82,8 @@ func memberConfigPatch(siteManagerState *state.SiteManagerState) types.EndpointH
 			return response.InternalError(fmt.Errorf("failed to get local client: %w", err))
 		}
 
-		if payload.HTTPSAddress != "" {
-			newAddress, err := microTypes.ParseAddrPort(payload.HTTPSAddress)
+		if payload.HTTPSAddress != nil {
+			newAddress, err := microTypes.ParseAddrPort(*payload.HTTPSAddress)
 			if err != nil {
 				return response.BadRequest(fmt.Errorf("invalid https_address for member %q: %w", memberName, err))
 			}
@@ -155,8 +155,8 @@ func memberConfigPatch(siteManagerState *state.SiteManagerState) types.EndpointH
 
 			// if no external address is provided, keep the existing one
 			externalAddress := dbConfigs[0].ExternalAddress
-			if payload.ExternalAddress != "" {
-				externalAddress = payload.ExternalAddress
+			if payload.ExternalAddress != nil {
+				externalAddress = *payload.ExternalAddress
 			}
 
 			serverConfigs, err := client.GetDaemonServerConfigs(r.Context(), localClient)
@@ -238,8 +238,8 @@ func toMemberConfigsAPI(dbConfigs []database.ManagerMemberConfig) []types.Member
 		memberConfigs = append(memberConfigs, types.MemberConfig{
 			Target: c.Target,
 			MemberConfigPatch: types.MemberConfigPatch{
-				HTTPSAddress:    c.HTTPSAddress,
-				ExternalAddress: c.ExternalAddress,
+				HTTPSAddress:    &c.HTTPSAddress,
+				ExternalAddress: &c.ExternalAddress,
 			},
 		})
 	}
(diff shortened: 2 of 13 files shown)


ui/src/pages/Settings.tsx
```diff
@@ -2,67 +2,140 @@ import { FC } from "react";
...
+          {
+            content: (
+              <h2 className="p-heading--5">{index === 0 ? "Cluster" : ""}</h2>
+            ),
+            role: "cell",
+            className: "scope",
+            "aria-label": "Scope",
+          },
+          {
+            content: <div className="key-cell">{key}</div>,
```
edlerd, line 58: There is a problem as we hide the "scope" column on mobile and it contains important info on which member the config belongs to. Maybe use the member name as a prefix to the key? So it is `member.foo.bar`.

Kxiru (commented):
LGTM! Thanks for also renaming the old configs to ManagerConfigs throughout the codebase.

ui/src/pages/Settings.tsx
```diff
@@ -2,67 +2,142 @@ import { FC } from "react";
...
+
+    for (const memberConfig of memberConfigOptions) {
+      const configKeys = Object.keys(defaultMemberConfigs);
+      const currentMemberConfigRows = configKeys.map((key, index) => {
+        return {
+          columns: [
+            {
+              content: (
+                <h2 className="p-heading--5">
+                  {index === 0 ? memberConfig.target : ""}
```
edlerd: Should we make this a bit more consistent with the "Cluster" scope above? Maybe "Member (:member-name)" or alike? We also need to be careful as the member name is user provided, so it might be very long, and it shouldn't break the ui in that case.

mas-who: Yeah that sounds good. I think text overflow elipses with a title attribute should be good. For both the Scope column and the Key column

internal/api/member_configs.go
```diff
@@ -63,12 +64,12 @@ func memberConfigPatch(siteManagerState *state.SiteManagerState) types.EndpointH
...
 		}
 
-		if payload.HTTPSAddress == "" && payload.ExternalAddress == "" {
+		if payload.HTTPSAddress == nil && payload.ExternalAddress == nil {
 			return response.BadRequest(fmt.Errorf("no fields provided to update"))
 		}
 
-		if payload.ExternalAddress != "" {
-			_, err = microTypes.ParseAddrPort(payload.ExternalAddress)
+		if payload.ExternalAddress != nil && *payload.ExternalAddress != "" {
```
edlerd: Why using a pointer here, but not on the other checks left and above?

internal/api/types/member_configs.go
```diff
@@ -2,8 +2,8 @@ package types
 
 // MemberConfigPatch represents the payload required to update configs for a single site manager member.
 type MemberConfigPatch struct {
-	HTTPSAddress    string `json:"https_address,omitempty"`
-	ExternalAddress string `json:"external_address,omitempty"`
+	HTTPSAddress    *string `json:"https_address,omitempty"`
+	ExternalAddress *string `json:"external_address,omitempty"`
```
edlerd: I think none of those should be pointers

ui/src/pages/Settings.tsx
```diff
@@ -2,67 +2,151 @@ import { FC } from "react";
...
+              "aria-label": "Key",
+            },
+            {
+              content: (
+                <SettingForm
+                  configField={key}
+                  value={
+                    memberConfig[key as keyof MemberOptions] ||
+                    defaultMemberConfigs[
+                      key as keyof typeof defaultMemberConfigs
```
edlerd: Can these casts be removed if we type `defaultMemberConfigs` in relation to MemberOptions?

ui/src/pages/settings/SettingForm.tsx
```diff
@@ -22,7 +23,11 @@ const SettingForm: FC<Props> = ({ configField, value, isLast }) => {
     const config = {
       [configField]: String(newValue),
     };
-    updateSettings(config)
+
+    (member
+      ? updateMemberConfigs(member, config)
+      : updateManagerConfigs(config)
```
edlerd, line 29: We need to display the errors and probably also should have success notifications.

internal/api/member_configs.go
mas-who: resolving this comment as the next one is related.

internal/api/types/member_configs.go
mas-who: as discussed, we will convert the payload to a config map and check if keys exist to avoid using pointers.

ui/src/pages/Settings.tsx
mas-who: yes and done :+1:

ui/src/pages/settings/SettingForm.tsx
mas-who: Yes but I think we already have a separate task planned to bring notification into site manager across the app?

internal/api/types/member_configs.go
mas-who: @edlerd I've done the changes as discussed and updated the spec for this.

internal/api/api.go
```diff
@@ -17,6 +18,11 @@ func authHandler(siteManagerState *state.SiteManagerState) types.AccessHandler {
 			return true, nil
 		}
 
+		// always allow cluster notification requests
+		if client.IsNotification(r) {
+			return true, nil
+		}
+
```
edlerd: This is dangerous. Allows to bypass any auth in case a header is set. Probably not something we should do.

internal/api/types/member_configs.go
```diff
@@ -13,9 +23,22 @@ type MemberConfig struct {
...
+	validConfigKeys := ValidMemberConfigKeys()
+	for k := range config {
+		_, ok := validConfigKeys[k]
+		if !ok {
+			return fmt.Errorf("invalid config key: %s", k)
+		}
+	}
+
+	return nil
+}
```
edlerd: I think we can go with a more lean approach: Not validate at all, to save a couple lines and any unknown config keys will be not handled / ignored.

ui/src/api/settings.ts
```diff
@@ -24,3 +24,29 @@ export const updateSettings = (config: LxdConfigPair): Promise<void> => {
...
+      method: "PATCH",
+      body: JSON.stringify({
+        config,
+      }),
+    })
+      .then(handleResponse)
+      .then(resolve)
+      .catch(reject);
+  });
+};
```
edlerd, line 52: When I click "Save" this is executed twice. It could also reproduce it in lxd-ui. I suspect there is something wrong with the form submission. If you find a fix, can you also open a PR on lxd-ui please?

ui/src/pages/settings/SettingForm.tsx
edlerd: Sounds good to me.

edlerd (changes requested):

internal/api/api.go
mas-who: This is needed since we need to forward a request to the specific member that requires config update. However, a cluster client will not have oidc cookies and will fail the auth check here. I think here if we check for the header, and that the request host is a valid member address in the cluster, then it should be sufficient wdyt?

mas-who: also this check can happen only if a request fails the oidc auth check

mas-who: actually, a better check here is to see if a forwarded request has a trusted cert

internal/api/types/member_configs.go
mas-who: yeah true, since we already validating if both `https_address` and `external_address` are missing, there is no need to do this validation. Will remove.

ui/src/api/settings.ts
mas-who: It's because in `SettingFormInput` we invoke the `onSubmit` prop for both form submission and on clicking the save button. Take out `onClick` for the save button and set its `type` attribute to `submit` fixes the issue. Will make a PR in lxd-ui as well.

edlerd (approved):

edlerd (approved):
LGTM, I could see a newly set external address in a created join token :+1:

mas-who:
> LGTM, I could see a newly set external address in a created join token 👍

Nice, thanks for checking that! :slightly_smiling_face:

PR #54, opened 2024-07-17
Add api endpoint and cli for editing site data [WD-13156]

mas-who:
## Done
 - added api endpoint `PATCH /1.0/sites/{siteName}` for updating `site_details` data
 - added cli for editing site data i.e. `go run ./cmd/lxd-site-mgr --state-dir state/dir1 site edit <site_name> [flags]`


internal/api/sites_management.go
```diff
@@ -45,6 +45,17 @@ func siteCmd(s *state.SiteManagerState) rest.Endpoint {
 	}
 }
 
+func siteApproveCmd(s *state.SiteManagerState) rest.Endpoint {
+	return rest.Endpoint{
+		Path: "sites/{siteName}/approve",
```
edlerd: We missed this in the spec, but I think the endpoint should be `POST /1.0/sites/:name` with a payload of `{status: active}`.
We don't need a dedicated endpoint for approval. A generic POST to modify a site should be fine.

edlerd (changes requested):

internal/api/sites_management.go
mas-who: yea sounds good :+1: I think we had this endpoint previously because it would modify mutliple table resources e.g. add site details then delete tokens. But we moved away from that approach so this makes sense

mas-who: made the change. I think the endpoint should be `PATCH` and not `POST` since we are updating configs. Will update the spec once this PR is finalised.

internal/api/sites_control.go
```diff
@@ -146,7 +146,7 @@ func parseStatusDistribution(statuses []types.StatusDistribution) (int64, string
 }
 
 func sitesPost(s state.State, r *http.Request) response.Response {
-	payload := types.SitePost{}
+	payload := types.SiteJoinPost{}
```
edlerd: I think this shouldn't be renamed.

cmd/lxd-site-mgr/sites.go
```diff
@@ -163,3 +170,49 @@ func (c *cmdSiteShow) run(cmd *cobra.Command, args []string) error {
...
 	return yaml.NewEncoder(os.Stdout).Encode(site)
 }
+
+type cmdsiteApprove struct {
+	common *CmdControl
+}
+
+func (c *cmdsiteApprove) command() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "approve-join <site_name>",
```
edlerd: Should this be `site edit <site_name>` with passing optional flags for setting the status?

mas-who: I think introducing a cli command like this one means we should also make the patch endpoint to allow editing all relevant site detail fields. Not sure if it's a little too early to be so generic?

edlerd: I'd be fine to limit the cli to changing status only for now. My concern was to avoid the very specific cli for approving, where we could have a more generic one that can be extended later.

mas-who:
I've updated the [spec](https://docs.google.com/document/d/1q0mlkekLO7iTVnktHFJU2VTGAE9ojtR7av1N3QHZGo8/edit) removing the `POST /1.0/sites/:name/approve` endpoint and added the `PATCH /1.0/sites/:name` endpoint with the `SitePatch` payload struct.

edlerd (approved):
LGTM :+1:

PR #56, opened 2024-07-17
feat(user-info): add /1.0 endpoint to retrieve user info and authentication state [WD-12702]

mas-who:
## Done
- added `GET /1.0` endpoint to return user info and authentication state
- adjust ui auth check to use `GET /1.0` endpoint

## QA
- make sure you are logged out. Enter `https://0.0.0.0:8414/1.0` in the browser url. You should get back the following payload
```
{
  "type": "sync",
  "status": "Failure",
  "status_code": 400,
  "operation": "",
  "error_code": 0,
  "error": "",
  "metadata": {
    "email": "",
    "name": "",
    "trusted": false
  }
}
```
- make sure you are logged in. Enter `https://0.0.0.0:8414/1.0` in the browser url. You should get back the following payload (as an example, user email and name may change)
```
{
  "type": "sync",
  "status": "Success",
  "status_code": 200,
  "operation": "",
  "error_code": 0,
  "error": "",
  "metadata": {
    "email": "mason.hu@canonical.com",
    "name": "mason.hu@canonical.com",
    "trusted": true
  }
}
```

### DIFF
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"context"
 	"net/http"
 
 	"github.com/canonical/lxd/lxd/response"
@@ -9,6 +10,7 @@ import (
 	microState "github.com/canonical/microcluster/state"
 
 	"github.com/canonical/lxd-site-manager/internal/api/types"
+	"github.com/canonical/lxd-site-manager/internal/oidc"
 	"github.com/canonical/lxd-site-manager/internal/state"
 )
 
@@ -27,7 +29,7 @@ func authHandler(siteManagerState *state.SiteManagerState) types.AccessHandler {
 			return false, response.Forbidden(nil)
 		}
 
-		_, resp, err := siteManagerState.OIDCVerifier.Auth(r.Context(), r)
+		result, resp, err := siteManagerState.OIDCVerifier.Auth(r.Context(), r)
 		if err != nil {
 			// check if the request is a cluster notification with valid cert
 			if client.IsNotification(r) && r.TLS != nil {
@@ -39,9 +41,27 @@ func authHandler(siteManagerState *state.SiteManagerState) types.AccessHandler {
 				}
 			}
 
+			if r.URL.Path == "/1.0" {
+				return false, response.SyncResponse(false, types.APIRoot{
+					Trusted: false,
+				})
+			}
+
 			return false, resp
 		}
 
+		setUserInfoInRequest(result, r)
+
 		return true, resp
 	}
+}
+
+func setUserInfoInRequest(authResult *oidc.AuthenticationResult, r *http.Request) {
+	userInfo := &types.UserInfo{
+		Email: authResult.Email,
+		Name:  authResult.Name,
+	}
+
+	userInfoCtx := context.WithValue(r.Context(), types.UserInfoKey, userInfo)
+	*r = *r.WithContext(userInfoCtx)
 }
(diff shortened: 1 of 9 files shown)


internal/api/user_info.go
```diff
@@ -0,0 +1,33 @@
...
+
+	"github.com/canonical/lxd/lxd/response"
+	"github.com/canonical/microcluster/rest"
+	microState "github.com/canonical/microcluster/state"
+
+	"github.com/canonical/lxd-site-manager/internal/api/types"
+	"github.com/canonical/lxd-site-manager/internal/state"
+)
+
+func userInfoCmd(s *state.SiteManagerState) rest.Endpoint {
```
edlerd: Should this be `apiRootCmd`?
We will likely add more info to the response later on. So this is not limited to user information.

This is also about the file name and the userInfoGet function name.

mas-who: yeah that makes sense. I've kept most of the types and added an extra `ApiRoot` type to represent the returned data from the root endpoint. The response now consists of a `user` field mapped to the `UserInfo` type. Let me know what you think, the idea is that in the future we may have other system information we'd like to return from this endpoint which might not be user related.

edlerd: Great, thanks! This is one thing I thought about previously: It was a bit odd to use the same type for the internal user context as for the 1.0 response. Because that has a danger of accidentally exposing user info later on, when it is extended with additional field meant for internal use.

I'd suggest removing the `Trusted` field from the internal user info type and have it only in the API response. The API response in my opinion should be flat, not referencing the user info as a nested field. Wdyt?

mas-who: Fair point. So the types will be something like

```
// UserInfo represents the information about a user.
type UserInfo struct {
	Email string `json:"email"`
	Name  string `json:"name"`
}

// APIRoot represents environment data for the site manager.
type APIRoot struct {
	UserInfo
	Trusted bool `json:"trusted"`
}
```

`UserInfo` is embedded in `APIRoot` so all the fields will be flat in the api response. It's just for type reuse

mas-who: @edlerd I've pushed the changes

internal/api/api.go
```diff
@@ -39,9 +41,27 @@ func authHandler(siteManagerState *state.SiteManagerState) types.AccessHandler {
 				}
 			}
 
+			if r.URL.Path == "/1.0" {
+				return false, response.SyncResponse(false, types.APIRoot{
+					Trusted: false,
+				})
+			}
+
```
edlerd, line 49: Having second thoughts on this one. I think we shouldn't have this special case here. Because this is an auth handler, it shouldn't get extra responsibilities like sending the api root endpoint response.
I think the apiRoot should be public and not even go through the auth handling process. It is special in the case that it can return more info if the user is logged in and less info if the user is not authenticated.

mas-who: That means we would have to implement the oidc check in the endpoint which is even weirder I think since we are duplicating logic in multiple places?

edlerd: Yes true, let's keep it as is then.

edlerd (approved):

edlerd (dismissed):
LGTM, shall we also add the auth check based on the new endpoint immediately in this PR?

mas-who:
> LGTM, shall we also add the auth check based on the new endpoint immediately in this PR?

Just implemented this, was going to do a follow up PR but I think it's small enough to just add to this one :slightly_smiling_face:

edlerd (approved):
LGTM, thanks for adding the frontend as well.

PR #59, opened 2024-07-19
fix: ui cluster approval should send request to `PATCH /1.0/sites`

mas-who:
## Done
 - small fix to ensure UI calls the right endpoint for approving a site.


Kxiru (approved):
LGTM, thanks for the fix!

PR #61, opened 2024-07-19
tests: add e2e tests for backend endpoints [WD-13158]

mas-who:
## Done
- Setup structure for backend e2e tests
- Add test helper functions for environment setup, cluster initialisation, test logging, creating http clients (unix and tls) and making requests
- Add e2e test for site join process
- Add e2e test for status updates


edlerd (commented):
I think we should
- either use a golang testing framework or
- have shell scripts that execute a series of command lines like `go run ./cmd/lxd-site-mgr --state-dir state/state_dir_1 external-site-join-token add foo12 --expiry=48h` followed by a `go run ./cmd/lxd-site-mgr --state-dir state/dir1 external-site-join-token list` and validate the output. Those can exit non-zero in case the output doesn't match.

mas-who:
> I think we should
> 
> * either use a golang testing framework or
> * have shell scripts that execute a series of command lines like `go run ./cmd/lxd-site-mgr --state-dir state/state_dir_1 external-site-join-token add foo12 --expiry=48h` followed by a `go run ./cmd/lxd-site-mgr --state-dir state/dir1 external-site-join-token list` and validate the output. Those can exit non-zero in case the output doesn't match.

The golang `httptest` framework won't work with the custom handlers we have to define in this project due to usage of microcluster.

The idea is not too far off from running a list of shell commands here, the reason why I chose to do the testing in go instead of just shell scripts is because I think setting up clients with proper certificates for the control listener tests etc would be much easier that way. Also we won't really need e2e tests for the management listener, since it would be covered by ui e2e tests.  Is there a particular benefit to do these tests using shell?

mas-who:
@edlerd I think it would be more organised to have another test suite for all cli commands. The e2e tests for the site join process and site status updates may cover some CLI commands but not all. It feels a bit better organised to comprehensively test the cli in a single test suite. Also happy to swap out the unix http requests in the current e2e tests to use cli though, let me know what you think.

test/e2e/main_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+var tests = []types.Test{
+	testSiteJoin,
+	testSiteStatus,
+}
+
+func TestE2E(t *testing.T) {
+	env, err := initE2E()
+	if err != nil {
+		t.Fatalf("Failed to testing initialize environment: %v", err)
```
edlerd:
```suggestion
		t.Fatalf("Failed to initialize test environment: %v", err)
```

test/e2e/main_test.go
```diff
@@ -0,0 +1,53 @@
...
+	// cleaning up environment
+	defer func() {
+		err := env.Cleanup()
+		if err != nil {
+			t.Fatalf("Failed to cleanup environment: %v", err)
+		}
+	}()
+
+	// run tests
+	t.Log("Given the need to test the LXD site manager backend")
```
edlerd:
```suggestion
```

test/e2e/main_test.go
```diff
@@ -0,0 +1,53 @@
...
+		t.Fatalf("Failed to testing initialize environment: %v", err)
+	}
+
+	// cleaning up environment
+	defer func() {
+		err := env.Cleanup()
+		if err != nil {
+			t.Fatalf("Failed to cleanup environment: %v", err)
+		}
+	}()
```
edlerd: Shouldn't this be done in the helpers `NewEnv` or `Init` function?

test/helpers/env.go
```diff
@@ -0,0 +1,173 @@
...
+		return err
+	}
+	if err := grepCmd.Wait(); err != nil {
+		return err
+	}
+	if err := psCmd.Wait(); err != nil {
+		return err
+	}
+
+	pids := bytes.Fields(awkOut.Bytes())
```
edlerd: This is not a safe way to get the pids. There might be other processes matching and being killed below. Can we record the PID after calling `daemonCmd.Start();` instead to know which pids  to kill?

test/helpers/cert.go
```diff
@@ -0,0 +1,13 @@
+package helpers
+
+import "github.com/canonical/lxd/shared"
+
+// GetClusterCert returns the cluster certificate.
+func GetClusterCert(env *Environment) (*shared.CertInfo, error) {
```
edlerd: Shouldn't this be a helper in env.go, so we can do `env.GetClustercert()`?

test/helpers/client.go
```diff
@@ -0,0 +1,257 @@
...
+		}
+
+		var d net.Dialer
+		return d.DialContext(ctx, "unix", raddr.String())
+	}
+
+	// Define the http transport
+	transport := &http.Transport{
+		DialContext:       unixDial,
+		DisableKeepAlives: true,
```
edlerd:
```suggestion
```
Why this?

test/helpers/client.go
```diff
@@ -0,0 +1,257 @@
...
+			}
+
+			return nil, fmt.Errorf("Unable to connect to %q: %w", addr, lastErr)
+		}
+	}
+
+	transport := &http.Transport{
+		TLSClientConfig:   tlsConfig,
+		DisableKeepAlives: true,
+		Proxy:             shared.ProxyFromEnvironment,
```
edlerd:
```suggestion
```
We don't need to support proxy here, should we?

test/helpers/client.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+
+	client := &http.Client{Transport: transport}
+	transport.DialTLSContext = tlsDialContext(transport)
+
+	client.CheckRedirect = func(req *http.Request, via []*http.Request) error {
+		// Replicate the headers
+		req.Header = via[len(via)-1].Header
+		return nil
+	}
```
edlerd:
```suggestion
```
We don't need to support redirects, do we?

test/helpers/client.go
```diff
@@ -0,0 +1,257 @@
...
+				if err != nil {
+					return nil, err
+				}
+
+				return conn, nil
+			}
+
+			return nil, fmt.Errorf("Unable to connect to %q: %w", addr, lastErr)
+		}
+	}
```
edlerd: Can't we rely on the default here?

test/helpers/cluster.go
```diff
@@ -0,0 +1,34 @@
...
+// InitSingleMemberCluster initializes a single member cluster for testing site manager.
+func InitSingleMemberCluster(env *Environment) error {
+	memberName := "member1"
+	coreAddress := "0.0.0.0:9100"
+	controlAddress := "0.0.0.0:9110"
+	stateDir := env.StateDir(&memberName)
+
+	// Start the daemon
+	daemonCmd := exec.Command("go", "run", env.DaemonPath(), "--state-dir", stateDir)
+	if err := daemonCmd.Start(); err != nil {
```
edlerd: After this we should be able to read the pid of the background daemon.

test/e2e/site_join.go
```diff
@@ -0,0 +1,191 @@
...
+		var tokenData types.ExternalSiteTokenBody
+
+		{
+			condition = "Should be able to create a site join token"
+			token, err = createSiteJoinToken(env, siteName)
+			if err != nil {
+				helpers.LogTestOutcome(t, condition, err)
+			}
+
+			helpers.LogTestOutcome(t, condition, nil)
```
edlerd:
```suggestion
			helpers.LogTestOutcome(t, condition, err)
```
Can keep this simple?

test/e2e/site_join.go
```diff
@@ -0,0 +1,191 @@
...
+			if site.Status != string(types.PENDING_APPROVAL) {
+				err = fmt.Errorf("invalid site")
+				helpers.LogTestOutcome(t, condition, err)
+			}
+
+			helpers.LogTestOutcome(t, condition, nil)
+		}
+
+		{
+			condition = "Should not be able to find the external site join token after receiving join request"
```
edlerd:
```suggestion
			condition = "Should have deleted the external site join token after receiving join request"
```

test/e2e/site_status.go
```diff
@@ -0,0 +1,195 @@
...
+
+		{
+			condition = "Should be able to receive a join request"
+			err = sendJoinRequest(env, tokenData)
+			if err != nil {
+				helpers.LogTestOutcome(t, condition, err)
+			}
+
+			helpers.LogTestOutcome(t, condition, nil)
+		}
```
edlerd: This is duplicating from site join. Should we make one long scenario with join and status together instead, to avoid duplicating?

edlerd (commented):
This is looking very good, thanks for adding. I ended up having a lot of questions and suggestions below.

test/e2e/main_test.go
mas-who: If we run cleanup in init, it will kill the daemon process before we hit the tests

test/helpers/env.go
mas-who: thanks for suggesting this. Yes it is possible after some reading. You can get the pid by `daemonCmd.Process.Pid`. We can introduce a `ProcessIDs` slice field for `Environment` and add all the pids to that slice during initialisation. Then in the cleanup function we can iterate those pids to terminate them.

test/helpers/cert.go
mas-who: fair point :+1:

test/helpers/client.go
mas-who: yeah no this is not required, removed.

test/helpers/client.go
mas-who: yes it's not required, we are not testing using any proxy so removed.

test/helpers/client.go
mas-who: removed.

test/helpers/client.go
mas-who: Yes indeed we can, removed.

test/helpers/cluster.go
mas-who: yes, now we have a `ProcessIDs` slice field for a instance of `Environment`. At this point we will add the pid to the slice and in `Environment.Cleanup` we will get the list of pids and terminate them.

test/e2e/site_status.go
mas-who: decided to just combine the test into a single file `site_control.go`. If we look at [kubernetes](https://github.com/kubernetes/kubernetes/blob/master/test/e2e/e2e_test.go) test folder structure, they also have a single entry point and seems they keep tests for different aspect of the system in separate folders. I don't think we are at that scale yet to make that kind of split though.

test/e2e/e2e_test.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+}
+
+func initE2E(t *testing.T) *helpers.Environment {
+	env := helpers.NewEnv()
+	err := env.Init()
+	if err != nil {
+		helpers.LogTestOutcome(t, "Initialize environment", err)
+	}
+	helpers.LogTestOutcome(t, "Initialize environment", nil)
```
edlerd:
```suggestion
	helpers.LogTestOutcome(t, "Initialize environment", err)
```

test/e2e/site_control.go
```diff
@@ -0,0 +1,321 @@
...
+
+		{
+			condition = "Should be able to get site with PENDING_APPROVAL status"
+			site, err := helpers.FindSite(env, siteName)
+			if err != nil {
+				helpers.LogTestOutcome(t, condition, err)
+			}
+
+			if site.Status != string(types.PENDING_APPROVAL) {
+				err = fmt.Errorf("invalid site")
```
edlerd:
```suggestion
				err = fmt.Errorf("invalid site status")
```

test/helpers/cluster.go
```diff
@@ -0,0 +1,57 @@
...
+	if err := daemonCmd.Run(); err != nil {
+		LogTestOutcome(t, "Build daemon", err)
+	}
+	LogTestOutcome(t, "Build daemon", nil)
+
+	// Build the CLI
+	cliCmd := exec.Command("go", "build", "-o", "lxd-site-mgr", env.CLIPath())
+	if err := cliCmd.Run(); err != nil {
+		LogTestOutcome(t, "Build CLI", err)
+	}
```
edlerd: Should we rely on make here instead?

edlerd (commented):
This looks good for the happy path. Shall we also add site join with an invalid hmac or expired cluster cert and status update with a not approved site or an invalid cluster cert?

test/helpers/cluster.go
mas-who: great idea, done! :+1:

edlerd: I see the comments resolved / answered, did you maybe forget to push the changes or are you implementing more changes?

mas-who: yes I'm just adding the failure test cases before pushing. Resolved conversations to keep track of what I've done.

mas-who:
@edlerd I've added the failure cases as well for site join and site status update. The only case I didn't cover is for an expired client cert. The reason being that can not be done easily with the lxd shared package for certificates. To create a certificate that expires immediately, the logic is a bit involved and we could reuse some of the code from lxd but we'd have to copy them over into this project. This PR is already getting quite big, so I think we can always add that case at a later stage.

test/e2e/main_test.go
edlerd: Ah right, because that function finishes early.

test/e2e/site_status.go
```diff
@@ -0,0 +1,222 @@
...
+
+		{
+			condition = "Should fail status update request with an inactive site"
+			err = sendStatusUpdate(env, tokenData)
+			if err != nil && err.Error() == "site not found" {
+				err = nil
+			}
+			helpers.LogTestOutcome(t, condition, err)
+		}
+
```
edlerd:
```suggestion
```

test/e2e/site_join.go
```diff
@@ -0,0 +1,96 @@
...
+			err = json.Unmarshal(decodedToken, &tokenData)
+			if err != nil {
+				helpers.LogTestOutcome(t, condition, err)
+			}
+
+			helpers.LogTestOutcome(t, condition, nil)
+		}
+
+		{
+			condition = "Should fail join request validation with invalid HMAC"
```
edlerd:
```suggestion
		{
			condition = "Should fail join request validation with invalid HMAC"

			token, _ = createSiteJoinToken(env, siteName, time.Time{})
			decodedToken, _ := base64.StdEncoding.DecodeString(token)
			json.Unmarshal(decodedToken, &tokenData)
```
As those prerequisites are covered by the happy path test, should we keep those failure tests focused on the failure case and avoid repeating the verbose checking of the pre-condition setup?

test/e2e/site_control.go
edlerd, line 1: Shall we call this file `site_happy` and the other ones `site_fail_join` and `site_fail_status`?

test/helpers/env.go
```diff
@@ -0,0 +1,164 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	for {
+		if _, err := os.Stat(wd + "/go.mod"); err == nil {
+			return wd
+		}
+		wd = wd[:len(wd)-1]
+	}
```
edlerd: This is a indirect way of defining the root by finding the first one that has a go.mod file in it. Maybe use something like below instead? Or maybe there is a safer way?
```
    _, b, _, _ = runtime.Caller(0)
    basepath   = filepath.Dir(b)
```

edlerd (changes requested):

test/e2e/site_status.go
mas-who: Why are we removing this?

edlerd: Sorry, forgot to comment. I think this is from the test above. This test is about another case, which is mentioned below.

test/e2e/site_join.go
mas-who: yeah good idea, will apply through out.

mas-who: I think we should still log the errors though.

edlerd: That is making the code so verbose. Is there a better way to keep the code slim?

test/e2e/site_status.go
mas-who: Aaah my bad I didn't look around the block of code.

test/e2e/site_join.go
mas-who: Since the token creation and decoding is a common sequence of events across all test cases. I've moved it into the helpers package as a token helper. So the error handling is mostly handled inside the helper function.

test/e2e/site_control.go
mas-who: maybe `site_normal` instead of `site_happy`?

test/e2e/site_join.go
edlerd: nice, thanks.

test/e2e/site_control.go
edlerd: Huh, what is normal these days :D Maybe `site_success`?

test/helpers/env.go
mas-who: I think it's safe to use the `go.mod` file since there can only be one for a go project, and we are expanding out from within the project. Problem with `runtime.Caller` is that it will return the directory from which the function is called, so it would still require some iterative logic to get back to the project root.

The current approach is a bit unsafe since we will hang the test program if `go.mod` file is missing for whatever reason in the project. I've added some logic so that if we reach the root of the file system, the program will exit with an error.

test/e2e/site_control.go
mas-who: Haha yeah that sounds better :laughing:

test/e2e/site_fail_status.go
```diff
@@ -0,0 +1,141 @@
...
+
+		{
+			condition = "Should fail status update request with an inactive site"
+
+			tokenData, err := helpers.CreateAndReturnSiteJoinToken(env, siteName, time.Time{})
+			if err != nil {
+				helpers.LogTestOutcome(t, condition, err)
+			}
+
+			err = sendStatusUpdate(env, tokenData)
```
edlerd: Don't we need to send the join request as well so the site is generated but not active?

test/e2e/site_fail_status.go
```diff
@@ -0,0 +1,141 @@
...
+
+		{
+			condition = "Should fail status update request with no client certificate"
+
+			tokenData, err := helpers.CreateAndReturnSiteJoinToken(env, siteName, time.Time{})
+			if err != nil {
+				helpers.LogTestOutcome(t, condition, err)
+			}
+
+			err = sendStatusUpdateNoCert(env, tokenData)
```
edlerd: I think we also need to send the join request and approve it so the site is active.

edlerd (commented):
This looks pretty good, I reviewed everything and have two questions left. In the site status fail case: I think we the prerequisites before sending the status request need a bit more refinement.

test/e2e/site_fail_status.go
mas-who: as discussed, I've added the join request steps to these test cases.

test/e2e/site_fail_status.go
mas-who: done.

edlerd (approved):
Good work, thanks for adding the tests and making them awesome.

PR #62, opened 2024-07-22
feat: [WD-13372] Added inline notification system.

Kxiru:
**Work done:**
- Added NotificationProvider and NotificationConsumer components.
- Implemented some initial notifications for approving and declining pending clusters, and creating new tokens.

Screenshots:
image

image

image

### DIFF
--- a/ui/src/Root.tsx
+++ b/ui/src/Root.tsx
@@ -1,22 +1,33 @@
 import { FC } from "react";
 import App from "./App";
 import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
-import { Application } from "@canonical/react-components";
-
+import {
+  Application,
+  NotificationProvider,
+  QueuedNotification,
+} from "@canonical/react-components";
 import Navigation from "./components/Navigation";
 import { AuthProvider } from "context/auth";
 import StatusBar from "components/StatusBar";
+import { useLocation } from "react-router-dom";
 
 const queryClient = new QueryClient();
 
 const Root: FC = () => {
+  const location = useLocation() as QueuedNotification;
+
   return (
     <QueryClientProvider client={queryClient}>
       <AuthProvider>
         <Application>
-          <Navigation />
-          <App />
-          <StatusBar />
+          <NotificationProvider
+            state={location.state}
+            pathname={location.pathname}
+          >
+            <Navigation />
+            <App />
+            <StatusBar />
+          </NotificationProvider>
         </Application>
       </AuthProvider>
     </QueryClientProvider>
--- a/ui/src/api/tokens.ts
+++ b/ui/src/api/tokens.ts
@@ -1,5 +1,5 @@
 import { LxdApiResponse } from "types/apiResponse";
-import { Token } from "types/token";
+import { Token, TokenPostResponse } from "types/token";
 import { handleResponse } from "util/helpers";
 
 export const fetchTokens = (): Promise<Token[]> => {
@@ -11,14 +11,14 @@ export const fetchTokens = (): Promise<Token[]> => {
   });
 };
 
-export const createToken = (body: string): Promise<void> => {
+export const createToken = (body: string): Promise<TokenPostResponse> => {
   return new Promise((resolve, reject) => {
     fetch("/1.0/external-site-join-token", {
       method: "POST",
       body: body,
     })
       .then(handleResponse)
-      .then(resolve)
+      .then((data: LxdApiResponse<TokenPostResponse>) => resolve(data.metadata))
       .catch(reject);
   });
 };
--- a/ui/src/pages/clusters/ClusterList.tsx
+++ b/ui/src/pages/clusters/ClusterList.tsx
@@ -8,6 +8,7 @@ import ClusterListActive from "./ClusterListActive";
 import ClusterListPending from "./ClusterListPending";
 import AddClusterButton from "./AddClusterButton";
 import ClusterStatusGraph from "./metrics/ClusterStatusGraph";
+import NotificationRow from "components/NotificationRow";
 
 const ClusterList: FC = () => {
   const { activeTab } = useParams<{
@@ -24,6 +25,7 @@ const ClusterList: FC = () => {
           items={[<ClusterStatusGraph key="clusterstatusgraph" />]}
         />
         <TabLinks tabs={tabs} activeTab={activeTab} tabUrl="/ui/sites" />
+        <NotificationRow />
         <div>
           {!activeTab && <ClusterListActive />}
           {activeTab === "pending" && <ClusterListPending />}
--- a/ui/src/pages/clusters/CreateCluster.tsx
+++ b/ui/src/pages/clusters/CreateCluster.tsx
@@ -5,6 +5,7 @@ import {
   Col,
   Form,
   Row,
+  useNotify,
 } from "@canonical/react-components";
 import BaseLayout from "components/BaseLayout";
 import { useNavigate } from "react-router-dom";
@@ -19,10 +20,12 @@ import * as Yup from "yup";
 import { queryKeys } from "util/queryKeys";
 import { useQueryClient } from "@tanstack/react-query";
 import { getDefaultExpiryDate } from "util/createCluster";
+import NotificationRow from "components/NotificationRow";
 
 const CreateCluster: FC = () => {
   const navigate = useNavigate();
   const queryClient = useQueryClient();
+  const notify = useNotify();
 
   const ClusterSchema = Yup.object().shape({
     site_name: Yup.string().required("Site name is required"),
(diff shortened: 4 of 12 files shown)


ui/src/pages/clusters/ClusterList.tsx
```diff
@@ -23,6 +24,7 @@ const ClusterList: FC = () => {
           inline={true}
           items={[<ClusterStatusGraph key="clusterstatusgraph" />]}
         />
+        <NotificationRow />
```
edlerd: I think this should appear below the tab links, not above them. The notifications relate to one tabs content, not to the generic charts above the tab selector.

ui/src/pages/clusters/CreateCluster.tsx
```diff
@@ -37,11 +40,13 @@ const CreateCluster: FC = () => {
     createToken(JSON.stringify(tokenPayload))
       .then(() => {
         navigate(`/ui/sites/tokens`);
+        notify.success(`Successfully created token ${tokenPayload.site_name}.`);
```
edlerd: We should also print the token itself here, with a note, that it can never be accessed again and should be saved right away.

ui/src/pages/clusters/CreateCluster.tsx
```diff
@@ -37,11 +40,13 @@ const CreateCluster: FC = () => {
...
       })
       .catch((e: Error) => {
-        if (e.message === "Unable to create new Token") {
+        if (e.message === "Unable to create new Token.") {
```
edlerd: I don't understand why we need this condition here. I think those lines should be removed.

ui/src/pages/clusters/CreateCluster.tsx
```diff
@@ -37,11 +40,13 @@ const CreateCluster: FC = () => {
...
           return;
         }
+        notify.failure("Unable to create new Token.", e);
```
edlerd:
```suggestion
        notify.failure("Unable to create token.", e);
```
Better?

ui/src/pages/clusters/RevokeTokenButton.tsx
```diff
@@ -11,16 +11,19 @@ interface Props {
...
   const queryClient = useQueryClient();
+  const notify = useNotify();
 
   const handleDeleteToken = async () => {
     await deleteToken(token.site_name)
       .then(() => {
         void queryClient.invalidateQueries({
           queryKey: [queryKeys.tokens],
         });
+        notify.success(`Successfully deleted ${token.site_name} token.`);
```
edlerd:
```suggestion
        notify.success(`Successfully deleted token for cluster ${token.site_name}.`);
```

ui/src/pages/clusters/RevokeTokenButton.tsx
```diff
@@ -11,16 +11,19 @@ interface Props {
...
       })
       .catch((e: Error) => {
-        if (e.message === "Unable to delete Token") {
+        if (e.message === "Unable to delete Token.") {
```
edlerd: There shouldn't be a conditional here. We want to surface any error to the user.

edlerd (changes requested):

ui/src/pages/clusters/RevokeTokenButton.tsx
edlerd: I think we should always use the wording "token for cluster :site-name", wdyt?

Kxiru: I suppose that does make sense, given the sitename is not the name of the token itself but instead of the cluster it is associated with.

ui/src/pages/clusters/CreateCluster.tsx
Kxiru: image

ui/src/pages/settings/SettingForm.tsx
```diff
@@ -30,6 +30,7 @@ const SettingForm: FC<Props> = ({ configField, value, isLast, member }) => {
     )
       .then(() => {
         setEditMode(false);
+        notify.success(`Successfully changed setting: ${configField}.`);
```
edlerd:
```suggestion
        notify.success(`Setting ${configField} updated.`);
```

ui/src/components/NavLink.tsx
```diff
@@ -31,12 +35,22 @@ const NavLink: FC<Props & LinkHTMLAttributes<HTMLElement>> = ({
     }
   }
 
+  const handleOnClick = () => {
+    if (onClick) {
+      onClick();
+    }
+    if (!isActive) {
+      notify.clear();
+    }
```
edlerd: We shouldn't leak information about notification state into the nav components. Otherwise, we end up adding notify.clear to all our internal links as well. Better keep the responsibilities clear. We need to pass some props to the notification provider for this to work. See the other comment.

ui/src/Root.tsx
```diff
@@ -14,9 +14,11 @@ const Root: FC = () => {
     <QueryClientProvider client={queryClient}>
       <AuthProvider>
         <Application>
-          <Navigation />
-          <App />
-          <StatusBar />
+          <NotificationProvider>
```
edlerd:
```suggestion
          <NotificationProvider state={location.state} pathname={location.pathname}>
```
This paired with adding `const location = useLocation() as QueuedNotification;` above causes the notification to be cleared on navigation or whenever the pathname changes.

ui/src/pages/clusters/CreateCluster.tsx
```diff
@@ -31,17 +35,26 @@ const CreateCluster: FC = () => {
...
     const tokenPayload = newTokenPayload(values);
 
     createToken(JSON.stringify(tokenPayload))
-      .then(() => {
+      .then((response) => {
         navigate(`/ui/sites/tokens`);
+        notify.success(
+          successMessageElement(response),
+          "Successfully created token.",
+        );
```
edlerd: When implementing the other changes above, we're probably going to need a queued notification like `navigate("/ui...", notify.queue(notify.success(...)));` here, that is passed as a 2nd argument to navigate. Otherwise, this notification is in a race and might get cleared right away before being shown.

edlerd (changes requested):

ui/src/components/NavLink.tsx
mas-who: This is my fault, I may have given some bad advice. @Kxiru please refer to [Root.tsx](https://github.com/canonical/lxd-ui/blob/main/src/Root.tsx) for reference to see how we pass location info to the `CombinedNotificationProvider` which internally has the `NotificationProvider` from react-components.

ui/src/api/tokens.ts
```diff
@@ -11,14 +11,16 @@ export const fetchTokens = (): Promise<Token[]> => {
...
+export const createToken = (body: string): Promise<TokenPostResponse> => {
   return new Promise((resolve, reject) => {
     fetch("/1.0/external-site-join-token", {
       method: "POST",
       body: body,
     })
       .then(handleResponse)
-      .then(resolve)
+      .then((data: LxdApiResponse<TokenPostResponse>) =>
+        resolve(data.metadata ?? []),
```
mas-who: we shouldn't default output of this function to `[]` since the api response structure is different. In any case, if the backend ever returned a success response with no data, it means something has gone wrong. We would want to know this so letting the UI fail will inform on the issue and we can then resolve it later.

mas-who:
> **Work done:**
> 
> * Added NotificationProvider and NotificationConsumer components.
> * Implemented some initial notifications for approving and declining pending clusters, and creating new tokens.
> 
> Screenshots: image
> 
> image
> 
> image

I think for the token success message, the `p` element that contains the token has a `max-width` of `40rem`, which is the default from vanilla. It looks a bit weird in this case, I think we should override it so that the element spans 100% of the container width. wdyt @Kxiru @edlerd ?

edlerd:
> I think for the token success message, the `p` element that contains the token has a `max-width` of `40rem`, which is the default from vanilla. It looks a bit weird in this case, I think we should override it so that the element spans 100% of the container width. wdyt @Kxiru @edlerd ?

I think p is the wrong element, we should use [code](https://vanillaframework.io/docs/base/code) instead. Then we won't need a CSS override any more, and I think it will look better automatically.

ui/src/pages/clusters/CreateCluster.tsx
Kxiru: When implementing this approach, the page does not navigate back from the cluster creation page to the token list page, although it does display the notification. Any idea why this could be?

ui/src/pages/clusters/CreateCluster.tsx
```diff
@@ -31,17 +35,30 @@ const CreateCluster: FC = () => {
...
+      .then((response) => {
+        navigate(
+          `/ui/sites/tokens`,
+          notify.queue(
+            notify.success(
+              successMessageElement(response),
+              "Successfully created token.",
+            ),
+          ),
+        );
```
edlerd:
```suggestion
        navigate(
          "/ui/sites/tokens",
          notify.queue(
            notify.success(
              response.token,
              "The token has been created and will be displayed only once. Please save it now:",
            ),
          ),
        );
```
We can't use any JSX in the queued notifications. The content must be text only. With this it works, but we need some extra scss to make the token break into multiple lines:

```
p-notification__message {
    word-break: break-word;
}
```
I think we can just add this to our scss and suggest it as a fix in vanilla as well.

Kxiru: I'll reach out to the Vanilla team now :). Thanks for your help!

edlerd (commented):
QA & code LGTM.

edlerd (approved):

PR #63, opened 2024-07-23
ui: added disk and memory distribution graphs for cluster list [WD-13150]

mas-who:
## Done
 - created `HorizontalDistGraph` component for displaying numerical data distributions along the y-axis
 - use `HorizontalDistGraph` to display disk and memory usage for clusters
 - added styling to make mobile view with the graphs list a bit easier on the eyes


ui/src/components/HorizontalDistGraph.tsx
```diff
@@ -0,0 +1,56 @@
...
+  color,
+  width,
+  height,
+}) => {
+  const maxDataValue = Math.max(...data);
+  const barWidth = height / data.length;
+
+  return (
+    <div className="horizontal-dist-graph" style={{ maxWidth: `${width}px` }}>
+      <p className="u-no-margin u-no-padding">{title}</p>
```
Kxiru: Is it worth making the title of the graph have the same styling as the title of the Cluster status graph for continuity purposes? Let me know your thoughts.
For reference, the styling of the status graph title is as follows:
```className="u-no-margin p-heading--5"```

Or perhaps that title should change to yours? In any case, I think the consistency would be useful.

**Review by Kxiru — COMMENTED** [review]
Looks great, in my opinion! Just a small idea, below.

File: ui/src/components/HorizontalDistGraph.tsx [hunk 2]
```diff
@@ -0,0 +1,56 @@
...
+  data: number[];
+  title: string;
+  color: string;
+  width: number;
+  height: number;
+};
+
+const X_GRID_LINE_RATIOS = [0.2, 0.4, 0.6, 0.8];
+
+const HorizontalDistGraph: FC<Props> = ({
```
edlerd: Maybe better `PercentileChart`?

ui/src/sass/styles.scss
edlerd, line 1: Why reorder the imports here? Makes it hard to see the actual changes

ui/src/sass/_cluster_graphs.scss
```diff
@@ -1,15 +1,26 @@
...
 
-  .cluster-status__chart {
-    width: $donut-size;
-  }
+  .cluster-status {
+    align-items: center;
+    display: flex;
+    flex-direction: row;
+    overflow: auto;
+    width: fit-content;
```
edlerd:
```suggestion
  .cluster-status {
    display: flex;
```
I think those are not needed?

ui/src/pages/clusters/metrics/ClusterMemoryGraph.tsx
```diff
@@ -0,0 +1,32 @@
+import { useQuery } from "@tanstack/react-query";
+import { fetchClusters } from "api/clusters";
+import HorizontalDistGraph from "components/HorizontalDistGraph";
+import { FC } from "react";
+import { queryKeys } from "util/queryKeys";
+
+const GRAPH_COLOR = "#B04000";
```
edlerd: I think we shouldn't have those in tsx. We define classes in `_meter.scss` with the colours for disk and memory, maybe make those reusable?

ui/src/components/HorizontalDistGraph.tsx
```diff
@@ -0,0 +1,56 @@
...
+  data,
+  title,
+  color,
+  width,
+  height,
+}) => {
+  const maxDataValue = Math.max(...data);
+  const barWidth = height / data.length;
```
edlerd: Should this be bar height as it is horizontal?

edlerd (changes requested):
This looks very good visually and so slim for the code as well.

Some small naming and minor structural comments below.

edlerd:
I think the headings and the lower end of the graphs should be aligned:

image

ui/src/components/HorizontalDistGraph.tsx
mas-who: In my head width and height is now inverted as the chart is horizontal, so `barWidth` makes sense to me since we assign it to the `height` property of the `rect` element :thinking:

ui/src/sass/_cluster_graphs.scss
mas-who: yup not needed

ui/src/sass/styles.scss
mas-who: I found there was an error that if `./breakpoints` was imported after other partials, it results in a build error. That was the primary reason for the re-order, but instead of just moving `breakpoints` higher, I decided to align the importing structure with lxd-ui

edlerd: Then can we have a new line before the other file imports and sort them alphabetically?

mas-who: I had them sorted in the wrong order previously I see :sweat:

mas-who:
> I think the headings and the lower end of the graphs should be aligned:
> 
> image

I think it's aligned now, but I may also be seeing things since it's late.

Screenshot from 2024-07-24 21-27-49

edlerd (approved):
LGTM, thanks for the adjustments.

For checking alignment I can recommend page ruler for [chrome](https://chromewebstore.google.com/detail/page-ruler/jcbmcnpepaddcedmjdcmhbekjhbfnlff?hl=en) or [firefox](https://addons.mozilla.org/en-US/firefox/addon/page-ruler/).

PR #64, opened 2024-07-23
fix: avoid flicker when loading login page

mas-who:
## Done
- Fix issue with ui showing cluster list before login page on first load.

### DIFF
--- a/ui/src/App.tsx
+++ b/ui/src/App.tsx
@@ -1,43 +1,47 @@
-import React, { FC, lazy, Suspense } from "react";
-import { Navigate, Route, Routes, useLocation } from "react-router-dom";
+import { FC, lazy, Suspense } from "react";
+import { Navigate, Route, Routes } from "react-router-dom";
 import NoMatch from "pages/NoMatch";
 import Settings from "pages/Settings";
 import { useAuth } from "context/auth";
 import CreateCluster from "pages/clusters/CreateCluster";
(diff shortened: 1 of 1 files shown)


edlerd:
This doesn't solve the problem yet. See below video. I think we should render a blank page as the loader jumps around while CSS is loading on the initial land.

[Screencast from 24.07.2024 10:42:15.webm](https://github.com/user-attachments/assets/f91c02e4-b236-496a-bf34-8c11d7f4aa09)

mas-who:
> This doesn't solve the problem yet. See below video. I think we should render a blank page as the loader jumps around while CSS is loading on the initial land.
> 
>  Screencast.from.24.07.2024.10.42.15.webm

Yeah I see, sorry I couldn't check it on the demo server yesterday (where the request delay is long enough to see this behaviour) due to a 404 error, turns out I had to clear cookies for jenkins in my browser. Will sort it out :+1:

edlerd (commented):
Thanks, this is nearly perfect now. One flicker I could find is on accessing a logged in page, when not being logged in, the no match component is briefly rendered. It also seems the no match layout is off, but that is another issue.  See video below

[Screencast from 24.07.2024 13:28:09.webm](https://github.com/user-attachments/assets/0a26f2c6-230d-4226-8c55-42a3c30d9c21)

edlerd (approved):
Now it is perfect. Thanks :)

PR #68, opened 2024-07-26
feat: [WD-13148] Added MultiMetricMeter component for nodes and insta…

Kxiru:
…nces.

Work done:
- Created MultiMeter component
- Added SCSS to Meter.scss.
- Changed the generate sites script to allow for more indicative visualisation data.

Edit (2):
image

### DIFF
--- a/scripts/generate_clusters.sh
+++ b/scripts/generate_clusters.sh
@@ -4,8 +4,8 @@ set -e
 ENTRIES=200
 
 # Instance and member statuses templates
-INSTANCE_STATUSES_TEMPLATE='[{"status": "Running", "count": %d}]'
-MEMBER_STATUSES_TEMPLATE='[{"status": "Online", "count": %d}]'
+INSTANCE_STATUSES_TEMPLATE='[{"status": "Running", "count": %d}, {"status": "Stopped", "count": %d}, {"status": "Frozen", "count": %d}, {"status": "Error", "count": %d}]'
+MEMBER_STATUSES_TEMPLATE='[{"status": "Online", "count": %d}, {"status": "Offline", "count": %d}, {"status": "Evacuated", "count": %d}, {"status": "Blocked", "count": %d}]'
 
 # Prepare bulk insert statements for core_remote_clusters
 CORE_REMOTE_CLUSTERS_INSERT="INSERT INTO core_remote_clusters (name, cluster_certificate) VALUES "
@@ -43,8 +43,8 @@ for i in $(seq 1 $ENTRIES); do
     DISK_USAGE=$(( (SEED * 12345678) % (DISK_TOTAL_SIZE + 1) ))  # 0 to DISK_TOTAL_SIZE
     # Generates a random date and time between date.now and 30 minutes ago.
     LAST_UPDATED_AT=$(date -u -d "@$(( $(date -u -d '30 minutes ago' +%s) + SEED % ($(date -u +%s) - $(date -u -d '30 minutes ago' +%s)) ))" +'%Y-%m-%d %H:%M:%S')
-    INSTANCE_STATUSES=$(printf "$INSTANCE_STATUSES_TEMPLATE" $((SEED % 200 + 1)))
-    MEMBER_STATUSES=$(printf "$MEMBER_STATUSES_TEMPLATE" $((SEED % 4)))
+    INSTANCE_STATUSES=$(printf "$INSTANCE_STATUSES_TEMPLATE" $((SEED % 200 + 1)) $((SEED % 150 + 1)) $((SEED % 100 + 1)) $((SEED % 50 + 1)))
+    MEMBER_STATUSES=$(printf "$MEMBER_STATUSES_TEMPLATE" $((SEED % 4 + 1)) $((SEED % 3 + 1)) $((SEED % 2 + 1)) $((SEED % 1 + 1)))
     REMOTE_CLUSTER_DETAILS_VALUES+=("($CORE_REMOTE_CLUSTER_ID, '$STATUS', '$CPU_TOTAL_COUNT', '$CPU_LOAD_1', '$CPU_LOAD_5', '$CPU_LOAD_15', '$MEMORY_TOTAL_AMOUNT', '$MEMORY_USAGE', '$DISK_TOTAL_SIZE', '$DISK_USAGE', '$INSTANCE_COUNT', '$INSTANCE_STATUSES', '$MEMBER_COUNT', '$MEMBER_STATUSES', '$LAST_UPDATED_AT')")
 done
 
--- a/ui/src/components/MultiMeter.tsx
+++ b/ui/src/components/MultiMeter.tsx
@@ -1,0 +1,40 @@
+import { FC } from "react";
+import { MultiMeterValue } from "types/cluster";
+
+interface Props {
+  values: MultiMeterValue[];
+  text: string;
+}
+
+const MultiMeter: FC<Props> = ({ values, text }) => {
+  function getPercentage(num: number): number {
+    let cumulativeTotal = 0;
+
+    for (const val of values) {
+      cumulativeTotal += val.amount;
+    }
+    return (num / cumulativeTotal) * 100;
+  }
+
+  return (
+    <>
+      <div className="p-meter u-no-margin--bottom">
+        {values.map((bar) => {
+          return (
+            <div
+              key={bar.status}
+              className={"meter-bar"}
+              style={{
+                width: `${getPercentage(bar.amount)}%`,
+                backgroundColor: `${bar.color}`,
+              }}
+            />
+          );
+        })}
+      </div>
+      <p className="p-text--small u-no-margin--bottom">{text}</p>
+    </>
+  );
+};
+
+export default MultiMeter;
--- a/ui/src/pages/clusters/metrics/ClusterInstances.tsx
+++ b/ui/src/pages/clusters/metrics/ClusterInstances.tsx
@@ -1,6 +1,7 @@
 import { FC } from "react";
 import { Cluster } from "types/cluster";
-import Meter from "components/Meter";
+import MultiMeter from "components/MultiMeter";
+import { Icon, Tooltip } from "@canonical/react-components";
 
 interface Props {
   cluster: Cluster;
@@ -8,15 +9,76 @@ interface Props {
 
 export const ClusterInstances: FC<Props> = ({ cluster }: Props) => {
   const runningInstances = cluster.instance_statuses.find(
-    (status) => status.status === "Running",
-  );
-  const runningCount = runningInstances ? runningInstances.count : 0;
+    (item) => item.status === "Running",
+  ) ?? { count: 0 };
+  const stoppedInstances = cluster.instance_statuses.find(
+    (item) => item.status === "Stopped",
+  ) ?? { count: 0 };
+  const frozenInstances = cluster.instance_statuses.find(
+    (item) => item.status === "Frozen",
+  ) ?? { count: 0 };
+  const errorInstances = cluster.instance_statuses.find(
+    (item) => item.status === "Error",
+  ) ?? { count: 0 };
+
+  const total =
+    runningInstances.count +
+    stoppedInstances.count +
+    frozenInstances.count +
+    errorInstances.count;
+
+  const values = [
+    {
+      amount: runningInstances.count,
+      status: "Running",
+      color: "#0E8420",
+    },
+    {
+      amount: frozenInstances.count,
+      status: "Frozen",
+      color: "#24598f",
+    },
+    {
+      amount: errorInstances.count,
+      status: "Error",
+      color: "#C7162B",
+    },
+    {
+      amount: stoppedInstances.count,
+      status: "Stopped",
+      color: "#000",
+    },
+  ];
 
   return (
-    <Meter
-      percentage={(100 / cluster.instance_count) * runningCount || 0}
-      text={`${cluster.instance_count} ( ${cluster.instance_count - runningCount} errored)`}
-      type="instances"
-    />
+    <Tooltip
+      message={
+        <div>
+          <div>
+            <Icon name="status-succeeded-small" />
+            {`Running (${runningInstances.count})`}
+          </div>
+          <div>
+            <Icon name="status-in-progress-small" />
+            {`Frozen (${frozenInstances.count})`}
+          </div>
+          <div>
+            <Icon name="status-failed-small" />
+            {`Error (${errorInstances.count})`}
+          </div>
+          <div>
+            <Icon name="status-queued-small" />
+            {`Stopped (${stoppedInstances.count})`}
+          </div>
+        </div>
+      }
+      positionElementClassName="tooltip"
+      position="btm-center"
+    >
+      <MultiMeter
+        values={values}
+        text={`${total} ( ${total - runningInstances.count} not running)`}
+      />
+    </Tooltip>
   );
 };
--- a/ui/src/pages/clusters/metrics/ClusterNodes.tsx
+++ b/ui/src/pages/clusters/metrics/ClusterNodes.tsx
@@ -1,4 +1,5 @@
-import Meter from "components/Meter";
+import { Icon, Tooltip } from "@canonical/react-components";
+import MultiMeter from "components/MultiMeter";
 import { FC } from "react";
 import { Cluster } from "types/cluster";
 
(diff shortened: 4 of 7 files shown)


ui/src/components/MultiMetricMeter.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { InstanceMeterMetrics, NodeMeterMetrics } from "types/cluster";
+
+interface Props {
+  values: NodeMeterMetrics | InstanceMeterMetrics;
+  text: string;
+  total: number;
+  type: "instances" | "cpu" | "disk" | "memory" | "nodes";
+}
+
+const MultiMetricMeter: FC<Props> = ({ values, text, type, total }) => {
```
edlerd: Maybe `MultiMeter`? Or do you have a better naming idea?

mas-who: I think this component should be generic and shouldn't contain details regarding instance metrics or node metrics. Those logic can be moved into `ClusterInstances` and `ClusterNodes`. The `values` prop can be an array in the form `{percent: number, colorStyle: string}[]`, then this component can simply iterate `values` and create as many segments as needed. That way, the `type` and `total` props will no longer be needed in this component as well.

ui/src/components/MultiMetricMeter.tsx
```diff
@@ -0,0 +1,88 @@
+import { FC } from "react";
+import { InstanceMeterMetrics, NodeMeterMetrics } from "types/cluster";
+
+interface Props {
+  values: NodeMeterMetrics | InstanceMeterMetrics;
+  text: string;
+  total: number;
+  type: "instances" | "cpu" | "disk" | "memory" | "nodes";
```
edlerd: I think the interface should be generic and not be tailored towards nodes, instances or any other use case. We can receive a generic `values: Record<string,int>` which represent the status to amount pairs. The statuses and percentiles pro status can be computed internally in the component. Colouring can be dependent on the ordering of the values. So first one is green, etc.

The total we can calculate in the component as well. The only other object in the interface can be the `caption: string` that is displayed below. Generally, we should try to keep the interfaces as small and generic as possible. Wdyt?

ui/src/sass/_meter.scss
```diff
@@ -3,16 +3,32 @@
...
+    background-color: $memory-color;
+  }
+
+  .error1 {
+    background-color: $disk-color;
+  }
+
+  .error2 {
+    background-color: $cpu-color;
+  }
```
edlerd: Reusing the colours seems a bit confusing, those are not instance/memory/disk/cpu values.

ui/src/pages/clusters/metrics/ClusterInstances.tsx
```diff
@@ -12,9 +12,30 @@ export const ClusterInstances: FC<Props> = ({ cluster }: Props) => {
   );
   const runningCount = runningInstances ? runningInstances.count : 0;
```
mas-who: as mentioned previously, in this component, we can calculate all the segments required as input for `MultiMetricMeter` i.e. `{percentage: number, colorStyle: string}[]` where `percentage` would be each possible status count divided by total instance count. `colorStyle` would be related to the css style for each status.

ui/src/pages/clusters/metrics/ClusterNodes.tsx
```diff
@@ -10,13 +10,34 @@ export const ClusterNodes: FC<Props> = ({ cluster }: Props) => {
   const runningMembers = cluster.member_statuses.find(
     (status) => status.status === "Online",
   );
-  const activeCount = runningMembers ? runningMembers.count : 0;
+  const runningCount = runningMembers ? runningMembers.count : 0;
+
+  const offlineMembers = cluster.member_statuses.find(
```
mas-who, line 14: same comment as for `ClusterInstances`

ui/src/types/cluster.d.ts
```diff
@@ -23,3 +23,17 @@ export interface Cluster {
   site_certificate: string;
   status: string;
 }
+
+export interface InstanceMeterMetrics {
```
mas-who: these interfaces won't be needed

mas-who (commented):
left some feedback. I think we can make the new `MultiMetricMeter` component more generic.

ui/src/components/MultiMetricMeter.tsx
mas-who: didn't see this comment as I reviewed the code. I proposed a big of a different approach with `values` being an array in the form `{percentage: number, colorStyle: string}[]`. I think this suggestion also works, I'm not too opinionated about which direction to take as long as `MultiMetricMeter` becomes generic. Should I remove my comment?

edlerd: I think we do need the status to amount connection for the legend.

mas-who: yeah true, I'm going to remove my suggestion then, let's not have conflicting suggestions and just go with this one :+1:

mas-who:
> …nces.
> 
> Work done:
> 
> * Create MultiMetricMeter component
> * Added SCSS to Meter.scsss
> * Added two new type interfaces for Node Metrics and Instance metrics.
> 
> image
> 
> Warning: The picture above uses hard-coded front-end generated dummy data for visualisation purposes (I was unable to create perfect-random generated statuses within the Go backend.
> 
> Todo:
> 
> * Add tooltips.

You can setup dummy data by adjusting `INSTANCE_STATUSES_TEMPLATE` and `MEMBER_STATUSES_TEMPLATE` in the `generate_sites.sh` script. For example:

```
INSTANCE_STATUSES_TEMPLATE='[{"status": "Running", "count": %d}, {"status": "Stopped", "count": %d}, {"status": "Frozen", "count": %d}, {"status": "Error", "count": %d}]'

...

INSTANCE_STATUSES=$(printf "$INSTANCE_STATUSES_TEMPLATE" $((SEED % 200 + 1)) $((SEED % 150 + 1)) $((SEED % 100 + 1)) $((SEED % 50 + 1)))
```

I think don't worry about randomness, the above code should give you some data points to work with so you don't need to hard code dummy data in the frontend for testing.

ui/src/components/MultiMetricMeter.tsx
Kxiru: @edlerd I mostly agree with your approach and would not be opposed to it, however I am not sure what randomly generated colours could look like for when there are many different attributes being shown in the meter. Are you against adding another prop to the record to pass through colour strings? It could be good for customisability too.

ui/src/components/MultiMetricMeter.tsx
Kxiru: I do not. I'll take yours :joy:

ui/src/components/MultiMetricMeter.tsx
edlerd: You mean the values become an array of tuples that represent "amount, status, colour"? Sounds good to me as well.

Kxiru:
@edlerd , I encountered an issue where it was not possible to use a Tooltip component around the status bars within the div, as it did not take a "style" prop, and was causing layout issues. As the width of the grid elements - the bar chart included - are all styled dynamically for responsivity purposes, it was not possible to adjust the tooltip through it's classnames in such a way that would allow the tooltip to span the full width of the meter component.

image

The screenshot above puts the comment into perspective. As you can see, the bar colours are taking up the width relative to their parent bar, which is 25% of the full width of the meter. @mas-who and I did not see a straightforward way to have the tooltip component take up the length of the meter, especially as the surrounding elements have widths that are dynamically set.

In lieu of this, @mas-who suggested that I use a title attribute on the bar divs. This implementation can be seen in the replaced screenshot in the PR body.

edlerd:
I think there was a misunderstanding. We can have one tooltip for the whole meter and it is a Multiline legend with amount and status per line. I think one tooltip or hover per status would be too tedious to discover

Edit: In the label we can maybe use the same colours as in the multimeter and present them as status labels (see below from vanilla docs)

image

See also this screen as an example of how the tooltip could be structured. Basically, when you haver the multi meter or it's caption, then this should open -- the tooltip can wrap the whole content.
image

Kxiru:
@edlerd 
image

Let me know what you think! :)

edlerd:
demo link: https://lxd-cluster-manager-68.demos.haus/

ui/src/components/MultiMeter.tsx
```diff
@@ -0,0 +1,51 @@
+import { Tooltip } from "@canonical/react-components";
+import { FC } from "react";
+import TooltipLegend from "./TooltipLegend";
+
+interface Props {
+  values: {
+    amount: number;
+    status: string;
+    color: string;
+  }[];
```
edlerd:
```suggestion
  values: MultiMeterValue[];
```

This is worth putting into a dedicated type, maybe `MultiMeterValue`.

ui/src/components/TooltipLegend.tsx
```diff
@@ -0,0 +1,41 @@
...
+            <Icon name="status-waiting-small" />
+            {`${values[1].status} (${values[1].amount})`}
+          </div>
+          <div>
+            <Icon name="status-failed-small" />
+            {`${values[2].status} (${values[2].amount})`}
+          </div>
+          <div>
+            <Icon name="status-queued-small" />
+            {`${values[3].status} (${values[3].amount})`}
```
edlerd: This is very opinionated of getting exactly 4 values in the right order. This is not an actual requirement from the interface of this component. The interface lets a client pass in any amount of values in any order. I think there is a better approach to how to construct the tooltip and to organize the components. Will explain below.

ui/src/components/MultiMeter.tsx
```diff
@@ -0,0 +1,51 @@
...
+    return (num / cumulativeTotal) * 100;
+  }
+
+  return (
+    <>
+      <Tooltip
+        message={<TooltipLegend values={values} />}
+        positionElementClassName="tooltip"
+        position="top-center"
+      >
```
edlerd: I think the tooltip might better be outside this component to keep this a graphing component with no assumptions on the order or amount of values.

ui/src/pages/clusters/metrics/ClusterInstances.tsx
```diff
@@ -1,22 +1,48 @@
...
+    stoppedInstances.count +
+    frozenInstances.count +
+    errorInstances.count;
 
   return (
-    <Meter
-      percentage={(100 / cluster.instance_count) * runningCount || 0}
-      text={`${cluster.instance_count} ( ${cluster.instance_count - runningCount} errored)`}
-      type="instances"
+    <MultiMeter
```
edlerd: I think we should move the tooltip into this component, together with the tooltip icons, etc. As we have the right counts already at hand with `runningInstances` and the other variables, we can use them in the right order and with the right icons attached directly in here.

ui/src/pages/clusters/metrics/ClusterInstances.tsx
```diff
@@ -1,22 +1,48 @@
...
 interface Props {
   cluster: Cluster;
 }
 
 export const ClusterInstances: FC<Props> = ({ cluster }: Props) => {
-  const runningInstances = cluster.instance_statuses.find(
-    (status) => status.status === "Running",
-  );
-  const runningCount = runningInstances ? runningInstances.count : 0;
+  const runningInstances = cluster.instance_statuses[0];
```
edlerd:
```suggestion
  const runningInstances = cluster.instance_statuses.find((item) => item.status === "running);
```
This should be a filter better than relying on the order. The order can change in the API response. Same for the other ones below and also in the cluster nodes component.

ui/src/pages/clusters/metrics/ClusterNodes.tsx
```diff
@@ -7,16 +7,42 @@ interface Props {
...
+    offlineMembers.count +
+    evacuatedMembers.count +
+    blockedMembers.count;
 
   return (
-    <Meter
-      percentage={(100 / cluster.member_count) * activeCount || 0}
-      text={`${cluster.member_count} ( ${cluster.member_count - activeCount} degraded)`}
-      type="instances" //Same Design
+    <MultiMeter
```
edlerd: Tooltip for nodes can live here.

ui/src/pages/clusters/metrics/ClusterInstances.tsx
```diff
@@ -1,22 +1,48 @@
...
+          status: frozenInstances.status,
+          color: "#C7162B",
+        },
+        {
+          amount: errorInstances.count,
+          status: errorInstances.status,
+          color: "#aaa",
+        },
+      ]}
+      text={`${total} ( ${total - runningInstances.count} errored)`}
```
edlerd:
```suggestion
      text={`${total} ( ${total - runningInstances.count} not running)`}
```
Better use exact copy.

scripts/generate_clusters.sh
```diff
@@ -43,8 +43,8 @@ for i in $(seq 1 $ENTRIES); do
     DISK_USAGE=$(( (SEED * 12345678) % (DISK_TOTAL_SIZE + 1) ))  # 0 to DISK_TOTAL_SIZE
     # Generates a random date and time between date.now and 30 minutes ago.
     LAST_UPDATED_AT=$(date -u -d "@$(( $(date -u -d '30 minutes ago' +%s) + SEED % ($(date -u +%s) - $(date -u -d '30 minutes ago' +%s)) ))" +'%Y-%m-%d %H:%M:%S')
-    INSTANCE_STATUSES=$(printf "$INSTANCE_STATUSES_TEMPLATE" $((SEED % 200 + 1)))
-    MEMBER_STATUSES=$(printf "$MEMBER_STATUSES_TEMPLATE" $((SEED % 4)))
+    INSTANCE_STATUSES=$(printf "$INSTANCE_STATUSES_TEMPLATE" $((SEED % 200 + 1)) $((SEED % 150 + 1)) $((SEED % 100 + 1)) $((SEED % 50 + 1)))
+    MEMBER_STATUSES=$(printf "$INSTANCE_STATUSES_TEMPLATE" $((SEED % 4 + 1)) $((SEED % 3 + 1)) $((SEED % 2 + 1)) $((SEED % 1 + 1)))
```
edlerd:
```suggestion
    MEMBER_STATUSES=$(printf "$MEMBER_STATUSES_TEMPLATE" $((SEED % 4 + 1)) $((SEED % 3 + 1)) $((SEED % 2 + 1)) $((SEED % 1 + 1)))
```
The members have different statuses, I think this would fix the generation to reflect that.

ui/src/components/MultiMeter.tsx
edlerd: Maybe better position the tooltip below and above?

edlerd (commented):
Looks good with the tooltip!
Some ideas to make this safer or improve structure below.

ui/src/components/MultiMeter.tsx
edlerd: Also I think the tooltip should wrap the graph and the below text "x (y degraded)".

ui/src/pages/clusters/metrics/ClusterInstances.tsx
edlerd: Edit: filter needs to be for "Running", not "running".

scripts/generate_clusters.sh
Kxiru: Oh wow, great catch, thank you

ui/src/components/MultiMeter.tsx
Kxiru: What do you mean by positioning the tooltip below AND above?

ui/src/pages/clusters/metrics/ClusterInstances.tsx
Kxiru: I am not sure I understand what you mean by having the icons in this component. Please could you review my code response to your review and let me know where to go from here?

ui/src/components/MultiMeter.tsx
edlerd: Sorry, mixup I meant it might look better below the content. In contrast to being above the content now.

ui/src/components/TooltipLegend.tsx
edlerd: I the problem above still persists, maybe inline the legend into the ClusterInstances and ClusterNodes component to have the strict connection from the values of a given status to a given icon. Then we won't need this component here.

edlerd (commented):
Thanks for the changes, this is much better. One more concern on the tooltip, I tried to explain it below.

ui/src/pages/clusters/metrics/ClusterInstances.tsx
```diff
@@ -1,22 +1,66 @@
...
+  const values = [
+    {
+      amount: runningInstances.count,
+      status: "Running",
+      color: "#0E8420",
+    },
+    {
+      amount: stoppedInstances.count,
+      status: "Stopped",
+      color: "#24598F",
```
edlerd: Those colours should match the icons we use in the legend. I think vanilla exposes them in some way, so we better reference them from it. If that's not possible also fine to encode them here and have them hardcoded as now.

ui/src/pages/clusters/metrics/ClusterInstances.tsx
edlerd: Yes, this is exactly what I meant. Just one more step. Tried to explain it below.

ui/src/pages/clusters/metrics/ClusterInstances.tsx
Kxiru: Vanilla uses all of them but the grey one. I tried to find a solution to that earlier to no avail.

Kxiru: That being said, unless you would like me to create the icons from scratch, I'll need to look into being able to re-colour the existing icon circles, as it was when I was trying to do this that I realised grey was not a vanilla colour.

edlerd: I see, fine to keep the colours here in that case. We should align them with the icons. See below.

For instances: The stopped bar and icon should probably be black, the error ones red and the frozen ones orange.

For nodes: offline: red, evacuated: black, blocked: orange.
image

edlerd: I think it is fine if we adopt the icon colours in the bars.

Kxiru: @edlerd , I have implemented your suggestions in the nodes column, if you like the look of it, I'll do the same for Instances.

image

Kxiru: SS of the Instances tooltip changes:
image

edlerd: Yes, legend/bar for nodes looks good to me.

Instance bars seem to not yet have the right colours in the 2nd screenshot, but I think you mentioned that already.

ui/src/components/TooltipLegend.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+interface Props {
+  values: {
+    amount: number;
+    status: string;
+    color: string;
+  }[];
+}
+
+const TooltipLegend: FC<Props> = ({ values }) => {
```
edlerd: I think this component is not generic, and it would be better to include the icon/status/amount pairs directly in ClusterInstances and ClusterNodes. That way, the connection between node- and instance-status and the icon becomes much more clear. Currently, that connection is hidden in the order of the values.

ui/src/pages/clusters/metrics/ClusterInstances.tsx
```diff
@@ -1,22 +1,66 @@
...
   return (
-    <Meter
-      percentage={(100 / cluster.instance_count) * runningCount || 0}
-      text={`${cluster.instance_count} ( ${cluster.instance_count - runningCount} errored)`}
-      type="instances"
-    />
+    <Tooltip
+      message={<TooltipLegend values={values} />}
+      positionElementClassName="tooltip"
+      position="top-center"
```
edlerd:
```suggestion
      position="btm-center"
```

ui/src/pages/clusters/metrics/ClusterInstances.tsx
```diff
@@ -1,22 +1,84 @@
...
+    <Tooltip
+      message={
+        <div>
+          <div>
+            <Icon name="status-succeeded-small" />
+            {`Running (${runningInstances.count})`}
+          </div>
+          <div>
+            <Icon name="status-waiting-small" />
+            {`Frozen (${frozenInstances.count})`}
```
edlerd:
```suggestion
            <Icon name="status-in-progress-small" />
            {`Frozen (${frozenInstances.count})`}
```
We should keep this in sync with `InstanceStatusIcon` in lxd-ui. Can you also adjust the colour for the frozen bar to be blue?

edlerd (dismissed):
Thanks for the adjustment, it is nearly perfect, just one colour for frozen needs adjustments to be aligned with what we use for the frozen status in lxd-ui.

ui/src/pages/clusters/metrics/ClusterInstances.tsx
Kxiru: Gotcha, all done!

edlerd (approved):
All good now, thanks for evolving it.

PR #69, opened 2024-07-29
cache: implement caching for site certificates [WD-12924]

mas-who:
## Done
- added a certificates cache to manager daemon state
- the certificates cache will be invalidated after 60 seconds, and will be rebuilt during the very next status update from lxd
- the cache is used for mtls authentication for the control listener status endpoint

### DIFF
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -14,7 +14,7 @@ import (
 	"github.com/canonical/lxd-cluster-manager/internal/state"
 )
 
-func authHandler(clusterManagerState *state.ClusterManagerState) types.AccessHandler {
+func oidcAuthHandler(clusterManagerState *state.ClusterManagerState) types.AccessHandler {
 	return func(clusterState microState.State, r *http.Request) (bool, response.Response) {
 		// always allow unix socket requests
 		if r.RemoteAddr == "@" {
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -2,9 +2,12 @@ package api
 
 import (
 	"context"
+	"fmt"
 	"net/http"
 
+	"github.com/canonical/lxd/lxd/request"
 	"github.com/canonical/lxd/lxd/response"
+	"github.com/canonical/lxd/lxd/util"
 	"github.com/canonical/microcluster/client"
 	"github.com/canonical/microcluster/rest/access"
 	microState "github.com/canonical/microcluster/state"
@@ -13,6 +16,8 @@ import (
 	"github.com/canonical/lxd-cluster-manager/internal/oidc"
 	"github.com/canonical/lxd-cluster-manager/internal/state"
 )
+
+const ctxRemoteClusterID request.CtxKey = "remote-cluster-id"
 
 func oidcAuthHandler(clusterManagerState *state.ClusterManagerState) types.AccessHandler {
 	return func(clusterState microState.State, r *http.Request) (bool, response.Response) {
@@ -53,6 +58,44 @@ func oidcAuthHandler(clusterManagerState *state.ClusterManagerState) types.Acces
 		setUserInfoInRequest(result, r)
 
 		return true, resp
+	}
+}
+
+// mtlsAuthHandler is an access handler that checks the client certificate against trusted remote cluster certificates.
+// the client certificate is checked against the trusted certificates in the CertificatesCache which expires every 60 seconds and will be rebuilt with db data.
+// if the client certificate is trusted, the certificate fingerprint and the associated remote cluster ID are set in the request context.
+// this access handler should be used for protected communication between lxd clusters and lxc cluster manager.
+func mtlsAuthHandler(clusterManagerState *state.ClusterManagerState) types.AccessHandler {
+	return func(clusterState microState.State, r *http.Request) (bool, response.Response) {
+		if r.TLS == nil || len(r.TLS.PeerCertificates) == 0 {
+			return false, response.Forbidden(fmt.Errorf("tls is required"))
+		}
+
+		if len(r.TLS.PeerCertificates) != 1 {
+			return false, response.Forbidden(fmt.Errorf("expected exactly one peer certificate"))
+		}
+
+		if clusterManagerState.CertificatesCache.Expired() {
+			err := clusterManagerState.CertificatesCache.RebuildCache(r.Context(), clusterState)
+
+			if err != nil {
+				return false, response.SmartError(err)
+			}
+		}
+
+		peerCert := r.TLS.PeerCertificates[0]
+		trustedCerts := clusterManagerState.CertificatesCache.GetTrustedCerts()
+		trusted, fingerprint := util.CheckMutualTLS(*peerCert, trustedCerts)
+
+		if !trusted {
+			return false, response.Forbidden(fmt.Errorf("invalid cluster certificate"))
+		}
+
+		remoteClusterCert, _ := clusterManagerState.CertificatesCache.GetCertificateEntry(fingerprint)
+		request.SetCtxValue(r, request.CtxUsername, fingerprint)
+		request.SetCtxValue(r, ctxRemoteClusterID, remoteClusterCert.ClusterID)
+
+		return true, nil
 	}
 }
 
--- a/internal/api/remote_cluster_control.go
+++ b/internal/api/remote_cluster_control.go
@@ -67,17 +67,7 @@ func remoteClustersStatusPost(managerState *state.ClusterManagerState) types.End
 				return fmt.Errorf("remote cluster is pending approval")
 			}
 
-			dbRemoteCluster.CPULoad1 = payload.CPULoad1
-			dbRemoteCluster.CPULoad5 = payload.CPULoad5
-			dbRemoteCluster.CPULoad15 = payload.CPULoad15
-			dbRemoteCluster.CPUTotalCount = payload.CPUTotalCount
-			dbRemoteCluster.DiskTotalSize = payload.DiskTotalSize
-			dbRemoteCluster.DiskUsage = payload.DiskUsage
-			dbRemoteCluster.InstanceCount, dbRemoteCluster.InstanceStatuses = parseStatusDistribution(payload.InstanceStatuses)
-			dbRemoteCluster.MemberCount, dbRemoteCluster.MemberStatuses = parseStatusDistribution(payload.MemberStatuses)
-			dbRemoteCluster.MemoryTotalAmount = payload.MemoryTotalAmount
-			dbRemoteCluster.MemoryUsage = payload.MemoryUsage
-			dbRemoteCluster.UpdatedAt = time.Now()
+			dbRemoteCluster.Put(payload)
 
 			err = database.UpdateRemoteClusterDetail(ctx, tx, remoteClusterID, *dbRemoteCluster)
 			if err != nil {
@@ -101,24 +91,6 @@ func remoteClustersStatusPost(managerState *state.ClusterManagerState) types.End
 			ClusterManagerAddresses: memberAddresses,
 		})
 	}
-}
-
-func parseStatusDistribution(statuses []types.StatusDistribution) (int64, string) {
-	if len(statuses) == 0 {
-		return 0, "[]"
-	}
-
-	parsedStatuses, err := json.Marshal(statuses)
-	if err != nil {
-		return 0, "[]"
-	}
-
-	var total int64
-	for _, s := range statuses {
-		total += s.Count
-	}
-
-	return total, string(parsedStatuses)
 }
 
 func remoteClustersPost(managerState *state.ClusterManagerState) types.EndpointHandler {
--- a/internal/api/remote_cluster_control.go
+++ b/internal/api/remote_cluster_control.go
@@ -1,7 +1,6 @@
 package api
 
 import (
-	"bytes"
 	"context"
 	"crypto/hmac"
 	"crypto/sha256"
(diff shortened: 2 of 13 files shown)


edlerd (commented):
Maybe an easier model is to build the cache on usage, if it is empty. Put the current timestamp into the cache next to the values. When you hit the cache and the current timestamp is more than 60 seconds after the cache time stamp, throw the cache away and rebuild it from scratch.

This way we don't need any chatter among cluster members or dependencies on add/remove certificates.

mas-who:
> Maybe an easier model is to build the cache on usage, if it is empty. Put the current timestamp into the cache next to the values. When you hit the cache and the current timestamp is more than 60 seconds after the cache time stamp, throw the cache away and rebuild it from scratch.
> 
> This way we don't need any chatter among cluster members or dependencies on add/remove certificates.

Yeah I agree with this approach, it would simplify the logic a lot. The only thing we have to double check would be the scenario where a site was delete and when the next status update takes place, the cached certificate is still valid. This will be checked at the point where we query the database with the cached site ID. In addition, I think we should set the expiry to 120 seconds instead of 60 since it is on the cusp of the status update interval which could result in the cache becoming useless.

edlerd:
> Yeah I agree with this approach, it would simplify the logic a lot. The only thing we have to double check would be the scenario where a site was delete and when the next status update takes place, the cached certificate is still valid. This will be checked at the point where we query the database with the cached site ID. In addition, I think we should set the expiry to 120 seconds instead of 60 since it is on the cusp of the status update interval which could result in the cache becoming useless.

I'd keep the cache lifetime at 60 seconds. Once you have many clusters, the cache will be effective.

For the delete case, we should be prepared for a status update to be authenticated, and the site can't be loaded. It can be solved easily, maybe the current implementation / error handling already solves it. This actually was possible before as a race condition, in case the site is deleted during the status update running slowly.

For the creation/approval case, we might reject a status update because the cache is preventing auth. But it is not really a problem, the cluster will retry in 60 seconds and then the cache is guaranteed to be rebuilt. For this reason, I'd avoid a long cache lifetime. This might we worth commenting in the code where we define the cache lifetime.

mas-who:
> > Yeah I agree with this approach, it would simplify the logic a lot. The only thing we have to double check would be the scenario where a site was delete and when the next status update takes place, the cached certificate is still valid. This will be checked at the point where we query the database with the cached site ID. In addition, I think we should set the expiry to 120 seconds instead of 60 since it is on the cusp of the status update interval which could result in the cache becoming useless.
> 
> I'd keep the cache lifetime at 60 seconds. Once you have many clusters, the cache will be effective.
> 
> For the delete case, we should be prepared for a status update to be authenticated, and the site can't be loaded. It can be solved easily, maybe the current implementation / error handling already solves it. This actually was possible before as a race condition, in case the site is deleted during the status update running slowly.
> 
> For the creation/approval case, we might reject a status update because the cache is preventing auth. But it is not really a problem, the cluster will retry in 60 seconds and then the cache is guaranteed to be rebuilt. For this reason, I'd avoid a long cache lifetime. This might we worth commenting in the code where we define the cache lifetime.

Yes the deletion case is dealt with in the status update endpoint already. In case if a site was found in the cache, but couldn't be loaded in the db, we will respond with `404 site not found`
```
// try site ID from manager cache
		siteID, certInCache := managerState.GetSiteIDByCertificate(certFingerprint)
		err := clusterState.Database().Transaction(r.Context(), func(ctx context.Context, tx *sql.Tx) error {
			var dbSites []database.CoreSiteWithDetails
			var err error

			if certInCache {
				dbSites, err = database.GetCoreSiteWithDetailByID(ctx, tx, siteID)
			} else {
				dbSites, err = database.GetCoreSitesWithDetails(ctx, tx)
			}

			if err != nil {
				return err
			}

			siteID = 0

			// certificate was not found in the db but was found in the cache
			// this means the site was deleted before the certificate cache expired
			if len(dbSites) == 0 && certInCache {
				managerState.DeleteCertificate(certFingerprint)
				return nil
			}

			for _, dbSite := range dbSites {
				if validSite(dbSite, peerCert) {
					siteID = dbSite.ID
					break
				}
			}

			return nil
		})

		if err != nil {
			logger.Warn("Failed to get site ID", logger.Ctx{"err": err})
		}

		if siteID == 0 {
			logger.Warn("Site not found")
			return response.NotFound(fmt.Errorf("site not found"))
		}
```

Regarding the cached cert lifetime,  each certificate cached will have it's own TTL so I don't think having many clusters joined would impact individual cert being removed from the cache? I think if not 120 seconds, we should at least set the TTL slightly above 60 seconds (or whatever the update interval may be).

edlerd:
> Regarding the cached cert lifetime, each certificate cached will have it's own TTL so I don't think having many clusters joined would impact individual cert being removed from the cache? I think if not 120 seconds, we should at least set the TTL slightly above 60 seconds (or whatever the update interval may be).

I think there is a misunderstanding. "The cache" means a list of all certificates and resolving them to a cluster id. This might be a map or some data structure kept in memory. This cache should be built when we authenticate an incoming status request. It stays valid for 60 seconds and any subsequent status updates will hit the cache. After 60 seconds the cache is considered stale, and the complete map is rebuilt. There is no notion of a cache per cluster or per certificate.

mas-who:
> > Regarding the cached cert lifetime, each certificate cached will have it's own TTL so I don't think having many clusters joined would impact individual cert being removed from the cache? I think if not 120 seconds, we should at least set the TTL slightly above 60 seconds (or whatever the update interval may be).
> 
> I think there is a misunderstanding. "The cache" means a list of all certificates and resolving them to a cluster id. This might be a map or some data structure kept in memory. This cache should be built when we authenticate an incoming status request. It stays valid for 60 seconds and any subsequent status updates will hit the cache. After 60 seconds the cache is considered stale, and the complete map is rebuilt. There is no notion of a cache per cluster or per certificate.

If I understand correctly, do you mean we will invalidate the entire cache at a fixed interval of 60 seconds? I'm not sure if this is efficient since new certificates may become invalid after a short period of time relative to the lifetime of the cache.  Having a TTL per certificate (or cache key) will allow each certificate to be effective for the duration of the TTL which I think works better no?

edlerd:
> If I understand correctly, do you mean we will invalidate the entire cache at a fixed interval of 60 seconds? I'm not sure if this is efficient since new certificates may become invalid after a short period of time relative to the lifetime of the cache. Having a TTL per certificate (or cache key) will allow each certificate to be effective for the duration of the TTL which I think works better no?

I think it only works in a "generate the whole cache". We anyway have to load all certs to do authentication for 1 site. Also this guards us from DDOS with invalid cert, which would go to DB on each unauthenticated request. Maybe easier to sync on this with a quick chat.

tomponline:
@edlerd @mas-who RE cache expiry, if we are expecting every site to check in regularly (once a minute?), what benefit do you see of removing cert cache entries quickly? Won't this just cause the certs to be re-parsed on every other request from each cluster?

edlerd:
> @edlerd @mas-who RE cache expiry, if we are expecting every site to check in regularly (once a minute?), what benefit do you see of removing cert cache entries quickly? Won't this just cause the certs to be re-parsed on every other request from each cluster?

Hey @tomponline :wave: 
The benefit is: we are building the cache once every 60 seconds. We build it with all certificates (because we have to load and parse all site certs anyway to build it).  Every status update within the next 60 seconds after the cache was build, will hit the cache. So the cache is effective. The cache is authoritative in that period, so even cache misses do not go to the DB.

We completely drop the cache after 60s and rebuild it from scratch. That will let us avoid the complexity of having to notify other cluster members on db writes.

markylaing:
@tomponline I've just had a call with Mason to understand this a bit better. The approach is to avoid the complexity of internal cluster notifications by accepting a trade-off that there will be occasional cache misses leading to a failure to authenticate a LXD status update (these are likely to be rare). 

Each member, separately, will:
1. Load the site certificate cache on start up, and set a TTL for the *whole cache* for 1 minute.
2. When a site sends a status update, check the cache expiry, if it has expired, reload it. Otherwise authenticate the site using the cache. If there is a cache miss, the site will not be able to authenticate.
3. If the site has been deleted from the site manager but is still in the cache, the status request is ignored.

Generally I think the approach is fine but it's acceptability is ultimately dependent on how tolerant we are about missed status updates. Also, it is likely that cluster notification system will be needed at some point anyway. 

Regarding the cluster notification system, the worry is that we wouldn't want to fail a request to e.g. add a new site if a cluster notification fails. For LXD, we refresh all caches when a cluster member comes back online (checking this in the dqlite heartbeat). We haven't encountered issues with random network failures yet, but could implement some kind of exponential backoff in this case.

tomponline:
> Regarding the cluster notification system, the worry is that we wouldn't want to fail a request to e.g. add a new site if a cluster notification fails.

LXD has the concept of notification policy, where a notification can complete successfully even if one or more of the cluster members is down (see `NewNotifier`) as well as refreshing the cache on start up and certificate add/change/delete, LXD also detects cluster member state changes during the periodic heartbeat process and triggers a refresh if a cluster member status has changed. This way after a "network interruption" the cluster members refresh their cache to be safe too.

tomponline:
> Also, it is likely that cluster notification system will be needed at some point anyway.

Does microcluster not provide something like this?

tomponline:
> We completely drop the cache after 60s and rebuild it from scratch. That will let us avoid the complexity of having to notify other cluster members on db writes.

I see, thanks for clarifying.

mas-who:
> > Also, it is likely that cluster notification system will be needed at some point anyway.
> 
> Does microcluster not provide something like this?

Yes microcluster does provide the functionality to send cluster notifications, although we would still need to setup internal endpoints and execute the notification in all the relevant endpoints. The current approach mainly benefits from a simplicity standpoint, since we rely on the cache being rebuilt at a somewhat fixed interval per member (if they receive status updates) and thus we don't need to worry about the complexity of cluster notification at the cost of some cache misses for site authentication.

markylaing:
> > Also, it is likely that cluster notification system will be needed at some point anyway.
> 
> Does microcluster not provide something like this?

microcluster has something like `cluster.Connect` which accepts a hook with a client to use to make requests to all members (including the host I believe). But it doesn't have the notification policy like LXD does.

mas-who:
@edlerd I've made the requested updates for the cache invalidation. @markylaing I've also extracted the mtls logic into its own access handler as requested. This PR is ready for a review thanks.

internal/state/cert_cache.go
```diff
@@ -0,0 +1,75 @@
...
+	return trustedCerts
+}
+
+// ReplaceAllInCache replaces all certificates in the cache with the provided map.
+func (c *CertificatesCache) ReplaceAllInCache(newCertCache map[string]CertificateCacheEntry) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	c.Certificates = newCertCache
+	c.TTL = time.Now().Add(60 * time.Second)
```
edlerd: Let's add a comment here, that the TTL is here to avoid sending cache invalidation messages to other cluster members.

internal/state/state.go
```diff
@@ -14,13 +15,19 @@ type ClusterManagerState struct {
 	OIDCVerifier *oidc.Verifier
 	// MicroCluster is the MicroCluster instance.
 	MicroCluster *microcluster.MicroCluster
-	mu           sync.RWMutex
+	// CertificateCache is a map of certificate fingerprints to cluster certificates.
```
edlerd:
```suggestion
	// CertificateCache is a map of certificate fingerprints to remote cluster certificates.
```

edlerd (dismissed):
LGTM, just two nitpicks on adding/changing comments

internal/api/api.go
```diff
@@ -2,19 +2,26 @@ package api
...
 	microState "github.com/canonical/microcluster/state"
 
 	"github.com/canonical/lxd-cluster-manager/internal/api/types"
+	"github.com/canonical/lxd-cluster-manager/internal/database"
 	"github.com/canonical/lxd-cluster-manager/internal/oidc"
 	"github.com/canonical/lxd-cluster-manager/internal/state"
 )
 
-func authHandler(clusterManagerState *state.ClusterManagerState) types.AccessHandler {
+func oidcAuthHandler(clusterManagerState *state.ClusterManagerState) types.AccessHandler {
```
edlerd, line 22: This is a nice improvement of keeping the auth logic contained and easy to understand :+1:

internal/state/cert_cache.go
mas-who: sure, although I think the comment might be better placed in the struct definition for `CertificatesCache`?

internal/state/cert_cache.go
```diff
@@ -0,0 +1,78 @@
...
+
+// CertificateCacheEntry represents a cache entry mapped to a certificate fingerprint.
+type CertificateCacheEntry struct {
+	ClusterID   int64
+	Certificate *x509.Certificate
+}
+
+// CertificateCache represent a cache of LXD cluster certificates with a TTL.
+type CertificatesCache struct {
+	Certificates map[string]CertificateCacheEntry
```
markylaing: In go it is more space efficient to use a `map[string]*CertificateCacheEntry` here.

internal/database/remote_cluster_details.relation.go
```diff
@@ -152,3 +156,53 @@ func GetCoreRemoteClusterWithDetailByName(ctx context.Context, tx *sql.Tx, remot
 
 	return objects, nil
 }
+
+// GetCoreRemoteClusterWithDetailByID fetches the remote cluster detail with core information from the database filtered by site id.
+func GetCoreRemoteClusterWithDetailByID(ctx context.Context, tx *sql.Tx, remoteClusterID int64) ([]CoreRemoteClusterWithDetail, error) {
```
markylaing: It looks like there is a lot of duplication between this function and the `GetCoreRemoteClusterWithDetailByName` function above.

internal/api/api.go
markylaing, line 22: Should this function signature have been changed in the previous commit?

internal/api/api.go
```diff
@@ -56,6 +63,65 @@ func authHandler(clusterManagerState *state.ClusterManagerState) types.AccessHan
...
+			return false, response.Forbidden(fmt.Errorf("tls is required"))
+		}
+
+		if len(r.TLS.PeerCertificates) != 1 {
+			return false, response.Forbidden(fmt.Errorf("expected exactly one peer certificate"))
+		}
+
+		// if the certificates cache is expired, rebuild it using cluster data from the database
+		if clusterManagerState.CertificatesCache.Expired() {
+			err := clusterState.Database().Transaction(r.Context(), func(ctx context.Context, tx *sql.Tx) error {
```
markylaing: It might be an idea to put this cache refresh logic into it's own function.

internal/api/remote_cluster_control.go
```diff
@@ -13,125 +12,93 @@ import (
...
-			return err
+			return response.BadRequest(err)
 		}
 
-		for _, dbRemoteCluster := range dbRemoteClusters {
-			if dbRemoteCluster.Status == string(types.PENDING_APPROVAL) {
-				continue
-			}
+		peerCert := r.TLS.PeerCertificates[0]
+		cachedCertificate, _ := managerState.CertificatesCache.GetCertificateEntry(shared.CertFingerprint(peerCert))
```
markylaing: It might be an idea to set the fingerprint of the caller certificate in the request context after calling `util.CheckMutualTLS` in the access handler. We do this in LXD as it represents the username and becomes accessible through the lifetime of the request.

You can use the `github.com/canonical/lxd/lxd/request` package to do this. E.g.
```golang
	// mtlsAuthHandler
	// ...
	trusted, fingerprint := util.CheckMutualTLS(*peerCert, trustedCerts)
	if !trusted {
		return false, response.Forbidden(fmt.Errorf("invalid cluster certificate"))
	}
	
	request.SetCtxValue(r, request.CtxUsername, fingerprint)
	return true, nil
	//...

	// remoteClusterStatusPost
	// ...
	fingerprint, err := request.GetCtxValue[string](r.Context(), request.CtxUsername)
	if err != nil {
	}
	// ...
```

internal/api/remote_cluster_control.go
```diff
@@ -13,125 +12,93 @@ import (
...
+			dbRemoteCluster.CPULoad5 = payload.CPULoad5
+			dbRemoteCluster.CPULoad15 = payload.CPULoad15
+			dbRemoteCluster.CPUTotalCount = payload.CPUTotalCount
+			dbRemoteCluster.DiskTotalSize = payload.DiskTotalSize
+			dbRemoteCluster.DiskUsage = payload.DiskUsage
+			dbRemoteCluster.InstanceCount, dbRemoteCluster.InstanceStatuses = parseStatusDistribution(payload.InstanceStatuses)
+			dbRemoteCluster.MemberCount, dbRemoteCluster.MemberStatuses = parseStatusDistribution(payload.MemberStatuses)
+			dbRemoteCluster.MemoryTotalAmount = payload.MemoryTotalAmount
+			dbRemoteCluster.MemoryUsage = payload.MemoryUsage
+			dbRemoteCluster.UpdatedAt = time.Now()
```
markylaing: I would define a `Put` method on the type of `dbRemoteCluster` here. Then if a field changes it's more clear where it needs to be updated.

markylaing (commented):
Looks good overall :)

internal/api/remote_cluster_control.go
edlerd: If we are using the request context, probably better to directly set the site id, we don't need the fingerprint later on. So we could
1. avoid calculating the fingerprint
2. simplify the cache to be a plain array instead of a map
3. store site id that we need later as a context.

internal/state/cert_cache.go
mas-who: thanks for the suggestion, I've made the change. Just one question, a `CertificateCacheEntry` here should not be mutated once it is in the map, in that case isn't copying safer? I guess the consideration here would be for memory efficiency and not for updates correct?

internal/database/remote_cluster_details.relation.go
mas-who: yes agreed, I've setup #72 to pull out this db method and refactored the file as well. Once the PR is merged I will rebase it here.

internal/api/api.go
mas-who: good catch, I've moved the change to the previous commit now

internal/api/api.go
mas-who: I've moved the logic into its own function out of `mtlsAuthHandler`

internal/api/remote_cluster_control.go
mas-who: I agree on setting the site id in the request context, since a lxd site is an "identity" in site manager. I think the `reuqest.CtxUsername` context key also works in this scenario. 

I'm not sure about converting the cache to a slice though, since we still need to get the site id from the cache based on cert fingerprint in `mtlsAuthHandler` before setting the request context. Furthermore, the `AddCertificate` method, which we need when receiving a join request, works better with a map because it guarantees no duplication in the cache.

edlerd: Ah right, also CheckMutualTLS is expecting a map with fingerprint as key.

internal/state/cert_cache.go
tomponline: The key reason for using pointers as the map values is that maps don't auto shrink when items are removed, so by using pointers you reduce the size each element takes when its not used.

internal/api/remote_cluster_control.go
mas-who: I'm not sure if I understand 100%, do you mean something like the below?

```
func (r *RemoteClusterDetail) Put(payload types.RemoteClusterStatusPost) {
	r.CPULoad1 = payload.CPULoad1
	r.CPULoad5 = payload.CPULoad5
	r.CPULoad15 = payload.CPULoad15
	r.CPUTotalCount = payload.CPUTotalCount
	r.DiskTotalSize = payload.DiskTotalSize
	r.DiskUsage = payload.DiskUsage
	r.InstanceCount, r.InstanceStatuses = parseStatusDistribution(payload.InstanceStatuses)
	r.MemberCount, r.MemberStatuses = parseStatusDistribution(payload.MemberStatuses)
	r.MemoryTotalAmount = payload.MemoryTotalAmount
	r.MemoryUsage = payload.MemoryUsage
	r.UpdatedAt = time.Now()
}
```

internal/state/cert_cache.go
mas-who: I see! thanks for this, good to know! :smiley:

mas-who:
@markylaing @edlerd  I've made suggested changes, it's ready for another review pass thanks.

internal/api/remote_cluster_control.go
markylaing: Yep that's right :)

internal/api/remote_cluster_control.go
markylaing: I still think the best approach here would be to set the fingerprint in the request context as the username. My reasoning is that a) you don't need the side ID in the access handler, so setting it here means the access handler is doing more work than it needs to, and b) this keeps site manager consistent with LXD.

If you do want to set the site ID in the request context, you could define a new context key for it:
```golang
const ctxSiteID request.CtxKey = "site-id"
```
Alternatively, you could add the whole cache entry to the request context.

internal/state/cert_cache.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// Expired returns true if the cache has expired.
+// No need to lock the cache here since TTL will not change concurrently after cache is created.
+func (c *CertificatesCache) Expired() bool {
+	return time.Now().After(c.TTL)
+}
+
+// GetCertificateEntry returns the cluster ID associated with a certificate fingerprint.
+func (c *CertificatesCache) GetCertificateEntry(fingerprint string) (*CertificateCacheEntry, bool) {
```
markylaing: I'd still return a `CertificateCacheEntry` from this function so that it's not modifiable.

```golang
entry, ok := c.Certificates[fingerprint]
if !ok || entry == nil {
    return CertificateCacheEntry{}, false
}

return *entry, true
```

internal/api/api.go
```diff
@@ -56,6 +59,39 @@ func oidcAuthHandler(clusterManagerState *state.ClusterManagerState) types.Acces
 	}
 }
 
+func mtlsAuthHandler(clusterManagerState *state.ClusterManagerState) types.AccessHandler {
```
markylaing, line 68: Although this function is not exported, it would still be useful to have a comment explaining what it does and where it should be used.

tomponline:
@mas-who @edlerd out of interest, as this is a very early stage project and the usage patterns are unknown at this point, did you consider just not implementing any caching mechanism at all at this early stage. Then when the hot paths are identified you could implement a caching mechanism that is known appropriate based on the access patterns later (which may or may not need cluster notifications - which may or may not have already been added by then based on other future unrelated requirements)?

internal/api/remote_cluster_control.go
mas-who: Sure makes sense to keep things consistent with lxd, although we don't need the site ID in the access handler, endpoints that needs to go through mtls auth will likely need the site ID. I think it makes sense to include the site ID in the request context with a custom key.

mas-who:
> @mas-who @edlerd out of interest, as this is a very early stage project and the usage patterns are unknown at this point, did you consider just not implementing any caching mechanism at all at this early stage. Then when the hot paths are identified you could implement a caching mechanism that is known appropriate based on the access patterns later (which may or may not need cluster notifications - which may or may not have already been added by then based on other future unrelated requirements)?

I agree that the caching mechanism may change in the future, happy to close this PR if that's the consensus? Although given that the PR is almost ready, and also this product isn't officially released, we can always change the approach later if necessary without major implications?

edlerd:
> @mas-who @edlerd out of interest, as this is a very early stage project and the usage patterns are unknown at this point, did you consider just not implementing any caching mechanism at all at this early stage. Then when the hot paths are identified you could implement a caching mechanism that is known appropriate based on the access patterns later (which may or may not need cluster notifications - which may or may not have already been added by then based on other future unrelated requirements)?

We have this and next week planned for focussing on cluster manager. Afterwards, we'll shift to lxd-ui. In the scope of cluster manager, we finished all the planned tasks and this was the most logical next step to include in the remaining planned timeframe.

internal/api/remote_cluster_control.go
mas-who: I've added a `ctxRemoteClusterID` key for setting lxd cluster id in the request context. We've recently changed the wording in the project that "site" is now "remoteCluster" (to create distinction from "clsuter" in microcluster)

mas-who:
@tomponline @edlerd should we go ahead with this PR? I think if we do decide to scrap caching for now, at least we should still extract the mtls auth logic out of the status endpoint as implemented in this PR (probably open a separate PR for that then).

edlerd (approved):

PR #70, opened 2024-07-30
chore: update dependencies to use lxd-cluster-manager

mas-who:
## Done

- Updated internal dependencies to use `lxd-cluster-manager` in import path


edlerd (approved):
LGTM

PR #71, opened 2024-07-30
feat(snap) add snap configuration WD-11369

edlerd:
# Done
- added snapcraft configuration
- fix the ui crashing when empty list of clusters is returned as `null` initially

# QA
1. Build the package locally: Install snapcraft with `snap install snapcraft`, checkout this branch, then run `snapcraft` from the top of this branch.
2. Install the locally built snap file, which should be in the repo directory with `sudo snap install --dangerous ./lxd-cluster-manager_0.1.0_amd64.snap`
3. Configure the service via terminal i.e. with
```
sudo lxd-cluster-manager init "member1" 0.0.0.0:9901 --bootstrap --control-address 0.0.0.0:9910
sudo lxd-cluster-manager config set oidc.issuer=https://lxd-ui-demo.us.auth0.com/
sudo lxd-cluster-manager config set oidc.client.id=OZSAeCbqAXZid3LL1gRQEkLXP9KlwZtJ
sudo lxd-cluster-manager config set oidc.audience=https://lxd-ui-demo.us.auth0.com/api/v2/
```
4. Now you can use the UI to register a cluster / connect a lxd remote cluster via https://127.0.0.1:9901/

Edit: Alternative is to `snap install lxd-cluster-manager --channel=latest/edge` to skip steps 1. and 2. above. I uploaded a local build to the snap store with the state of this branch.


snap/snapcraft.yaml
```diff
@@ -0,0 +1,123 @@
...
+      - libsqlite3-dev
+      - libuv1-dev
+      - pkg-config
+    organize:
+      usr/lib/: lib/
+    prime:
+      - lib/libdqlite*so*
+      - lib/*/libuv*
+
+  raft:
```
tomponline, line 48: @edlerd dqlite bundles raft now so this is not needed. See https://github.com/canonical/lxd-pkg-snap/blob/latest-edge/snapcraft.yaml#L337-L358

tomponline (commented):
@edlerd LGTM, when we come to publish this properly, I would like us to use the same method as we use for LXD and MicroCloud, which is to create a mirror repo in Launchpad (which triggers the snap builds, e.g. https://launchpad.net/~canonical-lxd/+snaps) and then have Github push to that when there is a merge.

This gives us an automated "edge" channel.

We should then also evaluate if we need to create a "candidate" channel and for that we would need a separate repo or branch (e.g. https://github.com/canonical/lxd-pkg-snap/tree/5.21-candidate) where we can maintain the candidate packaging files and automate building the candidate snaps.

Having a separate candidate build allows us to perform cherry-pick fixes on releases and can be invaluable.

edlerd:
Thanks for the review! Will simplify the raft bit shortly.

> create a mirror repo in Launchpad (which triggers the snap builds, e.g. https://launchpad.net/~canonical-lxd/+snaps) and then have Github push to that when there is a merge.

I think we can run the build and push to the snap store directly in a GitHub action, without the extra repo in launchpad. We had this in place for [anbox-cloud-dashboard](https://github.com/canonical/anbox-cloud-dashboard/actions/workflows/snap_release.yml) and it worked well. Though that repo is now merged into a bigger [ams](https://github.com/canonical/ams/actions/workflows/publish.yaml) repo, but that has the same strategy to build the snaps right here. The advantage is that the tagging and branching happens in one repository, and on any merge to main we push to the edge channel.

tomponline:
> I think we can run the build and push to the snap store directly in a GitHub action, without the extra repo in launchpad. We had this in place for [anbox-cloud-dashboard](https://github.com/canonical/anbox-cloud-dashboard/actions/workflows/snap_release.yml) and it worked well. Though that repo is now merged into a bigger [ams](https://github.com/canonical/ams/actions/workflows/publish.yaml) repo, but that has the same strategy to build the snaps right here. The advantage is that the tagging and branching happens in one repository, and on any merge to main we push to the edge channel.

Yeah, I realise this is possible, but it doesn't give us sufficient flexibility on the LXD and MicroCloud side.

But for a short time it may not be need for LXD cluster manager. Although I suspect in the longer term it will be required as well.

Lets discuss.

snap/snapcraft.yaml
edlerd: removing it in https://github.com/canonical/lxd-cluster-manager/pull/102

PR #75, opened 2024-07-31
feat: [WD-13152] Added Cluster detail page.

Kxiru:
Work done:
- Added a 'Fetch Cluster' api call.
- Added a Cluster Detail component.
- Added ClusterNodeGraph and ClusterInstanceGraph components.
- Added a Cluster Last Seen Component

image

Design review?

### DIFF
--- a/ui/src/App.tsx
+++ b/ui/src/App.tsx
@@ -5,6 +5,7 @@ import Settings from "pages/Settings";
 import { useAuth } from "context/auth";
 import CreateCluster from "pages/clusters/CreateCluster";
 import { logout } from "util/helpers";
+import ClusterDetail from "pages/clusters/ClusterDetail";
 
 const ClusterList = lazy(() => import("pages/clusters/ClusterList"));
 const Login = lazy(() => import("pages/Login"));
@@ -29,6 +30,7 @@ const App: FC = () => {
       <Route path="/ui/clusters" element={<ClusterList />} />
       <Route path="/ui/clusters/:activeTab" element={<ClusterList />} />
       <Route path="/ui/clusters/create" element={<CreateCluster />} />
+      <Route path="/ui/cluster/:name" element={<ClusterDetail />} />
       <Route path="/ui/settings" element={<Settings />} />
       <Route path="*" element={<NoMatch />} />
     </>
--- a/ui/src/api/clusters.ts
+++ b/ui/src/api/clusters.ts
@@ -11,6 +11,15 @@ export const fetchClusters = (): Promise<Cluster[]> => {
   });
 };
 
+export const fetchCluster = (remoteClusterName: string): Promise<Cluster> => {
+  return new Promise((resolve, reject) => {
+    fetch(`/1.0/remote-clusters/${remoteClusterName}`)
+      .then(handleResponse)
+      .then((data: LxdApiResponse<Cluster>) => resolve(data.metadata))
+      .catch(reject);
+  });
+};
+
 export const deleteCluster = (remoteClusterName: string): Promise<void> => {
   return new Promise((resolve, reject) => {
     fetch(`/1.0/remote-clusters/${remoteClusterName}`, {
--- a/ui/src/pages/clusters/ClusterDetail.tsx
+++ b/ui/src/pages/clusters/ClusterDetail.tsx
@@ -1,0 +1,83 @@
+import { Notification, List, Row, Strip } from "@canonical/react-components";
+import { useQuery } from "@tanstack/react-query";
+import { fetchCluster } from "api/clusters";
+import BaseLayout from "components/BaseLayout";
+import { FC } from "react";
+import { Link, useParams } from "react-router-dom";
+import { queryKeys } from "util/queryKeys";
+import ClusterDetailInstanceGraph from "./metrics/ClusterDetailInstanceGraph";
+import ClusterDetailNodeGraph from "./metrics/ClusterDetailNodeGraph";
+import ClusterTimer from "./metrics/ClusterTimer";
+import ClusterDetailMetrics from "./metrics/ClusterDetailMetrics";
+import Loader from "components/Loader";
+import BreadCrumbHeader from "components/BreadcrumbHeader";
+
+const ClusterDetail: FC = () => {
+  const { name } = useParams<{ name: string }>();
+  if (!name) {
+    return <>Missing name</>;
+  }
+
+  const {
+    data: cluster,
+    error,
+    isLoading,
+  } = useQuery({
+    queryKey: [queryKeys.clusters, name],
+    queryFn: () => fetchCluster(name),
+  });
+
+  if (!cluster) {
+    return <>Unable to get details</>;
+  }
+
+  return (
+    <BaseLayout
+      title={
+        <BreadCrumbHeader
+          name={`${cluster.name}`}
+          parentItems={[
+            <Link to={`/ui/clusters`} key={1}>
+              Clusters
+            </Link>,
+          ]}
+        />
+      }
+    >
+      {isLoading && <Loader text="Loading cluster details..." />}
+      {!isLoading && !cluster && !error && <>Loading cluster failed</>}
+      {error && (
+        <Strip>
+          <Notification severity="negative" title="Error">
+            {error.message}
+          </Notification>
+        </Strip>
+      )}
+      {!isLoading && cluster && (
+        <Row>
+          <List
+            className="cluster-detail-graphs"
+            inline
+            items={[
+              <ClusterTimer cluster={cluster} key="cluster-timer" />,
+              <ClusterDetailMetrics
+                cluster={cluster}
+                key="cluster-detail-metrics"
+              />,
+              <ClusterDetailNodeGraph
+                cluster={cluster}
+                key="cluster-node-graph"
+              />,
+              <ClusterDetailInstanceGraph
+                cluster={cluster}
+                key="cluster-instance-graph"
+              />,
+            ]}
+          />
+        </Row>
+      )}
+    </BaseLayout>
+  );
+};
+
+export default ClusterDetail;
--- a/ui/src/pages/clusters/metrics/ClusterDetailNodeGraph.tsx
+++ b/ui/src/pages/clusters/metrics/ClusterDetailNodeGraph.tsx
@@ -1,0 +1,82 @@
+import { Icon } from "@canonical/react-components";
+import DoughnutChart from "components/DoughnutChart";
+import { FC, ReactNode } from "react";
+import { Cluster } from "types/cluster";
+import { statusCount } from "util/helpers";
+
+interface Props {
+  cluster: Cluster;
+}
+
+const ClusterDetailNodeGraph: FC<Props> = ({ cluster }: Props) => {
+  const onlineMembers = statusCount(cluster.member_statuses, "Online");
+  const offlineMembers = statusCount(cluster.member_statuses, "Offline");
+  const evacuatedMembers = statusCount(cluster.member_statuses, "Evacuated");
+  const blockedMembers = statusCount(cluster.member_statuses, "Blocked");
+
+  const totalNodes =
+    onlineMembers + offlineMembers + evacuatedMembers + blockedMembers;
+
+  function getPercentageString(portion: number): ReactNode {
+    return (
+      <>
+        <b>{portion}</b> {`(${Math.floor((portion / totalNodes) * 100)}%) `}
+      </>
+    );
+  }
+
+  return (
+    <div className="cluster-detail-doughnut-graph">
+      <DoughnutChart
+        chartID="clusterNode"
+        segmentHoverWidth={45}
+        segmentWidth={40}
+        segments={[
+          {
+            color: "#0E8420",
+            tooltip: "Running",
+            value: onlineMembers,
+          },
+          {
+            color: "#CC7900",
+            tooltip: "Stopped",
+            value: offlineMembers,
+          },
+          {
+            color: "#C7162B",
+            tooltip: "Frozen",
+            value: evacuatedMembers,
+          },
+          { color: "#000", tooltip: "Error", value: blockedMembers },
+        ]}
+        size={150}
+      />
+      <ul className="doughnut-chart__legend u-no-margin--left">
+        <li className="u-no-margin p-heading--5 u-no-padding">
+          {totalNodes} Nodes
+        </li>
+        <li>
+          <Icon name="status-succeeded-small" />
+          {getPercentageString(onlineMembers)} Online
+        </li>
+        <li>
+          <Icon name="status-waiting-small" />
+          {getPercentageString(offlineMembers)}
+          Offline
+        </li>
+        <li>
+          <Icon name="status-failed-small" />
+          {getPercentageString(evacuatedMembers)}
+          Evacuated
+        </li>
+        <li>
+          <Icon name="status-failed-small" />
+          {getPercentageString(blockedMembers)}
+          Blocked
+        </li>
+      </ul>
+    </div>
+  );
+};
+
+export default ClusterDetailNodeGraph;
--- a/ui/src/pages/clusters/metrics/ClusterTimer.tsx
+++ b/ui/src/pages/clusters/metrics/ClusterTimer.tsx
@@ -1,0 +1,49 @@
+import { useQuery } from "@tanstack/react-query";
+import { fetchCluster } from "api/clusters";
+import { FC, useEffect, useState } from "react";
+import { Cluster } from "types/cluster";
+import { getSecondsSinceLastHeartbeat } from "util/helpers";
+import { queryKeys } from "util/queryKeys";
+
+interface Props {
+  cluster: Cluster;
+}
+
+const ClusterTimer: FC<Props> = ({ cluster }: Props) => {
+  const [seconds, setSeconds] = useState(getSecondsSinceLastHeartbeat(cluster));
+
+  const { data: data } = useQuery({
+    queryKey: [queryKeys.clusters, cluster.name],
+    queryFn: () => fetchCluster(cluster.name),
+    refetchInterval: 60000,
+    refetchIntervalInBackground: true,
+  });
+
+  useEffect(() => {
+    const timerId = setInterval(() => {
+      setSeconds((prev) => prev + 1);
+    }, 1000);
+    return () => clearInterval(timerId);
+  }, []);
+
+  useEffect(() => {
+    setSeconds(getSecondsSinceLastHeartbeat(data as Cluster));
+  }, [data]);
+
+  const getFormattedTimeLeft = () => {
+    const minutes = Math.floor(seconds / 60);
+    const displaySeconds = seconds % 60;
+
+    return `${minutes}:${displaySeconds < 10 ? `0${displaySeconds}` : displaySeconds}`;
+  };
+
+  return (
+    <div className="cluster-detail-countdown">
+      <div className="u-no-margin u-no-padding p-heading--3">
+        {getFormattedTimeLeft()}
+      </div>
+      <div>Since last heartbeat</div>
+    </div>
+  );
+};
+export default ClusterTimer;
--- a/ui/src/sass/_cluster_detail.scss
+++ b/ui/src/sass/_cluster_detail.scss
@@ -1,0 +1,54 @@
+.cluster-detail-graphs {
+  align-items: center;
+  display: flex;
+  margin-left: $sph--small;
+
+  .cluster-detail-metrics {
+    display: flex;
+    flex-direction: column;
+
+    .meter-row {
+      align-items: center;
+      display: flex;
+      gap: 1rem;
+
+      .meter-row__title {
+        min-width: 5rem;
+        text-align: right;
+      }
+
+      .meter-row__metrics {
+        width: 100%;
+      }
+    }
+  }
+
+  .cluster-detail-doughnut-graph {
+    align-items: center;
+    display: flex;
+  }
+
+  .cluster-detail-countdown {
+    justify-content: center;
+    text-align: center;
+    width: 100px;
+  }
+}
+
+@include mobile-and-tablet {
+  .cluster-detail-graphs {
+    align-items: center;
+    display: flex;
+    flex-direction: column;
+    gap: 1.5rem;
+
+    .p-inline-list__item {
+      margin-right: 1rem !important;
+    }
+
+    .doughnut-chart__legend {
+      margin-bottom: 0;
+      width: 200px;
+    }
+  }
+}
--- a/ui/src/sass/_cluster_graphs.scss
+++ b/ui/src/sass/_cluster_graphs.scss
@@ -7,16 +7,11 @@ $donut-size: 6.5rem;
   align-items: center;
   display: flex;
   flex-wrap: wrap;
-  gap: $sph--large;
 
-  .cluster-status {
+  .cluster-doughnut-graph {
     display: flex;
 
-    .cluster-status__chart {
-      width: $donut-size;
-    }
-
-    .cluster-status-legend {
+    .doughnut-chart__legend {
       margin-left: $sph--x-large;
     }
   }
--- a/ui/src/util/helpers.tsx
+++ b/ui/src/util/helpers.tsx
@@ -1,4 +1,4 @@
-import { Cluster } from "types/cluster";
+import { Cluster, StatusDistribution } from "types/cluster";
 
 interface ErrorResponse {
   error_code: number;
@@ -20,6 +20,11 @@ export const isWidthBelow = (width: number): boolean =>
 export function getMinutesSinceLastHeartbeat(cluster: Cluster): number {
   const lastSeenTime = Date.parse(cluster.last_status_update_at);
   return Math.floor((Date.now() - lastSeenTime) / 60000);
+}
+
+export function getSecondsSinceLastHeartbeat(cluster: Cluster): number {
+  const lastSeenTime = Date.parse(cluster.last_status_update_at);
+  return Math.floor((Date.now() - lastSeenTime) / 1000);
 }
 
 export const humanFileSize = (bytes: number): string => {
@@ -109,4 +114,16 @@ export const convertToISOFormat = (datetimeLocalString: string) => {
 
   // Return the ISO 8601 formatted string
   return date.toISOString();
+};
+
+// this works only for words that form the plural by adding an "s" at the end
+export const pluralizeWord = (word: string, count: number): string => {
+  return count === 1 ? word : `${word}s`;
+};
+
+export const statusCount = (
+  distribution: StatusDistribution[],
+  status: string,
+) => {
+  return distribution.find((item) => item.status === status)?.count ?? 0;
 };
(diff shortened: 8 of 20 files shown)


ui/src/App.tsx
```diff
@@ -29,6 +30,7 @@ const App: FC = () => {
       <Route path="/ui/clusters" element={<ClusterList />} />
       <Route path="/ui/clusters/:activeTab" element={<ClusterList />} />
       <Route path="/ui/clusters/create" element={<CreateCluster />} />
+      <Route path="/ui/cluster/:name" element={<ClusterDetail />} />
```
edlerd, line 33: Need to add a wildcard url to go to allow refresh on the detail page.

ui/src/api/clusters.ts
```diff
@@ -11,6 +11,15 @@ export const fetchClusters = (): Promise<Cluster[]> => {
   });
 };
 
+export const fetchCluster = (remoteClusterName: string): Promise<Cluster> => {
+  return new Promise((resolve, reject) => {
+    fetch(`/1.0/remote-clusters/${remoteClusterName}`)
```
edlerd, line 16: @mas-who there appears to be a bug on this endpoint, it only returns the first status, and avoids the other statuses for nodes and instance distribution. This seems like a database problem with the query generation to me.

ui/src/pages/clusters/ClusterDetail.tsx
```diff
@@ -0,0 +1,62 @@
...
+          className="cluster-graphs"
+          inline
+          items={[
+            <ClusterLSTimer cluster={cluster} key="cluster-timer" />,
+            <div key="cluster-disk">
+              <div className="u-margin">
+                <ClusterCpu cluster={cluster} />
+              </div>
+              <div>
+                <ClusterMemory cluster={cluster} />
```
edlerd: We should remove "memory" as a suffix from the text generated by the ClusterMemory component. Especially in the table that makes the rows look a bit odd:
image

ui/src/pages/clusters/ClusterDetail.tsx
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  return (
+    <BaseLayout title={`Clusters / ${name}`}>
+      <Row>
+        <List
+          className="cluster-graphs"
+          inline
+          items={[
+            <ClusterLSTimer cluster={cluster} key="cluster-timer" />,
```
edlerd: What is LS? Maybe better ClusterTimer ?

ui/src/pages/clusters/ClusterDetail.tsx
```diff
@@ -0,0 +1,62 @@
...
+            <div key="cluster-disk">
+              <div className="u-margin">
+                <ClusterCpu cluster={cluster} />
+              </div>
+              <div>
+                <ClusterMemory cluster={cluster} />
+              </div>
+              <div>
+                <ClusterDisk cluster={cluster} />
+              </div>
```
edlerd: This and the other graphs should have a label to the left saying "CPU", "Memory" or "Disk".
image

ui/src/pages/clusters/ClusterDetail.tsx
```diff
@@ -0,0 +1,62 @@
...
+            </div>,
+            <ClusterNodeGraph cluster={cluster} key="cluster-status-graph" />,
```
edlerd:
```suggestion
            <ClusterNodeGraph cluster={cluster} />,
```
We only need the key, when used in a map or when iterating over many elements that are the same.

ui/src/pages/clusters/ClusterDetail.tsx
```diff
@@ -0,0 +1,62 @@
...
+            <ClusterInstanceGraph
+              cluster={cluster}
+              key="cluster-status-graph"
```
edlerd:
```suggestion
```

ui/src/pages/clusters/metrics/ClusterNodeGraph.tsx
```diff
@@ -0,0 +1,93 @@
...
+import { Cluster } from "types/cluster";
+
+interface Props {
+  cluster: Cluster;
+}
+
+const ClusterNodeGraph: FC<Props> = ({ cluster }: Props) => {
+  const onlineMembers = cluster.member_statuses.find(
+    (item) => item.status === "Online",
+  ) ?? { count: 0 } ?? { count: 0 };
```
edlerd:
```suggestion
  const stausCount = (
    distribution: StatusDistribution[],
    status: string,
  ) => {
    return distribution.find((item) => item.status === status)?.count ?? 0;
  };

  const runningCount = stausCount(cluster.instance_statuses, "Running");
```
I think it is time to define a helper in a shared place. It can receive a distribution and status and return the count.

We can start using it in all the graphs that rely on find / count.

ui/src/pages/clusters/ClusterDetail.tsx
```diff
@@ -0,0 +1,62 @@
...
+    queryKey: [queryKeys.clusters, name],
+    queryFn: () => fetchCluster(name),
+  });
+
+  if (!cluster) {
+    return "Unable to get details";
+  }
+
+  return (
+    <BaseLayout title={`Clusters / ${name}`}>
```
edlerd: We should link the "Clusters" copy to the overview page.

ui/src/pages/clusters/ClusterDetail.tsx
```diff
@@ -0,0 +1,62 @@
...
+      <Row>
+        <List
+          className="cluster-graphs"
+          inline
+          items={[
+            <ClusterLSTimer cluster={cluster} key="cluster-timer" />,
+            <div key="cluster-disk">
```
edlerd:
```suggestion
            <div>
```
The key seems to be wrong.

ui/src/pages/clusters/ClusterDetail.tsx
```diff
@@ -0,0 +1,62 @@
...
+              <div className="u-margin">
```
edlerd:
```suggestion
              <div>
```
Not sure if this is needed? I think we don't need the wrapping divs at all and can keep the html structure flatter.

edlerd (dismissed):
Looking good, some suggestions below.

@mas-who can you have a look why the single cluster endpoint is only returning the first status and not all statuses? Maybe also add a test for it with the fix.

ui/src/api/clusters.ts
mas-who: interesting, will look into it :+1:

mas-who: just did a check, seems like it's working fine?

Screenshot from 2024-07-31 18-43-21

mas-who: I remember @Kxiru had an issue in her `generate_clusters.sh` script where the dummy data wasn't included due to a merge issue. This seems very similar to that

edlerd: Thanks for checking, I thought on the demo server the list page had multiple statuses, but it doesn't either. @Kxiru can you rebase this branch on main?

mas-who: Aah that makes sense then, the script changes was merged with the status distribution meter PR. If this branch was created before that got merged then a rebase and rebuild should fix the issue.

ui/src/pages/clusters/ClusterDetail.tsx
Kxiru: LS meant Last Seen.

ui/src/pages/clusters/ClusterDetail.tsx
Kxiru: Got that coming in a separate PR - a whole Breadcrumbs implementation. But I can put it in this PR instead?

ui/src/pages/clusters/ClusterDetail.tsx
Kxiru: React/lint-js complains if there is no key used in the list. :disappointed:

Kxiru:
@mas-who Do you mind giving this a once over?

ui/src/pages/clusters/ClusterDetail.tsx
```diff
@@ -0,0 +1,61 @@
...
+import ClusterNodeGraph from "./metrics/ClusterNodeGraph";
+import ClusterTimer from "./metrics/ClusterTimer";
+import ClusterDetailMetrics from "./metrics/ClusterDetailMetrics";
+import Loader from "components/Loader";
+
+const ClusterDetail: FC = () => {
+  const { name } = useParams<{ name: string }>();
+  if (!name) {
+    return <>Missing name</>;
+  }
```
mas-who, line 19:
```suggestion
  if (!name) {
    return <>Missing name</>;
  }
  
```

ui/src/pages/clusters/ClusterDetail.tsx
```diff
@@ -0,0 +1,61 @@
...
+    queryKey: [queryKeys.clusters, name],
+    queryFn: () => fetchCluster(name),
+  });
+
+  if (!cluster) {
+    return <>Unable to get details</>;
+  }
+
+  return (
+    <BaseLayout title={`Clusters / ${name}`}>
```
mas-who: this will need to be adjusted after the breadcrumbs PR gets merged.

ui/src/pages/clusters/ClusterDetail.tsx
```diff
@@ -0,0 +1,61 @@
...
+      {isLoading && <Loader text="Loading instance details..." />}
```
mas-who:
```suggestion
      {isLoading && <Loader text="Loading cluster details..." />}
```

ui/src/pages/clusters/ClusterDetail.tsx
```diff
@@ -0,0 +1,61 @@
...
+      {!isLoading && !cluster && !error && <>Loading instance failed</>}
```
mas-who:
```suggestion
      {!isLoading && !cluster && !error && <>Loading cluster failed</>}
```

ui/src/pages/clusters/ClusterDetail.tsx
```diff
@@ -0,0 +1,61 @@
...
+      {error && <Strip>{error.message}</Strip>}
```
mas-who: let's add the `Notification` component here to make the error message more visually pleasing.

ui/src/pages/clusters/metrics/ClusterInstanceGraph.tsx
```diff
@@ -0,0 +1,79 @@
...
+            color: "#CC7900",
+            tooltip: "Stopped",
+            value: stoppedInstances,
+          },
+          { color: "#C7162B", tooltip: "Frozen", value: frozenInstances },
+          { color: "#000", tooltip: "Error", value: errorInstances },
+        ]}
+        size={150}
+      />
+      <div className="cluster-status-legend">
```
mas-who: we should have semantic html here. This should be a un-ordered list `<ul>` with `<li>` elements. Can we also update the implementation in `ClusterStatusGraph` to align. Sorry I may have missed this in the Doughnut chart PR, can't recall if I reviewed it. In general we should always try to strive for semantic html, it's especially beneficial for accessibility with screen reader users, also good for SEO and general code readability.

ui/src/pages/clusters/metrics/ClusterInstanceGraph.tsx
```diff
@@ -0,0 +1,79 @@
...
+  function getPercentageString(portion: number): ReactNode {
+    return (
+      <>
+        <b>{portion}</b> {`(${Math.floor((portion / totalInstances) * 100)}%) `}
+      </>
+    );
+  }
+
+  return (
+    <div className="cluster-status">
```
mas-who: small nit, it's a little weird to have a class `cluster-status` associated to this component. We can improve the css class names to be more generic. wdyt?

```
.cluster-graphs {
  align-items: center;
  display: flex;
  flex-wrap: wrap;
  gap: $sph--large;

  .cluster-doughnut-graph {
    display: flex;

    .cluster-doughnut-graph__chart {
      width: $donut-size;
    }

    .cluster-doughnut-graph__legend {
      margin-left: $sph--x-large;
    }
  }

  .cluster-disk-bar {
    fill: $disk-color;
  }

  .cluster-memory-bar {
    fill: $memory-color;
  }
}
```

ui/src/pages/clusters/metrics/ClusterInstanceGraph.tsx
```diff
@@ -0,0 +1,79 @@
...
+      <DoughnutChart
+        className="cluster-status__chart"
+        chartID="clusterStatus"
```
mas-who:
```suggestion
        chartID="clusterInstance"
```

ui/src/pages/clusters/metrics/ClusterNodeGraph.tsx
```diff
@@ -0,0 +1,83 @@
...
+import DoughnutChart from "components/DoughnutChart";
+import { FC, ReactNode } from "react";
+import { Cluster } from "types/cluster";
+import { statusCount } from "util/helpers";
+
+interface Props {
+  cluster: Cluster;
+}
+
+const ClusterNodeGraph: FC<Props> = ({ cluster }: Props) => {
```
mas-who: same comments from `ClusterInstanceGraph`

ui/src/pages/clusters/metrics/ClusterTimer.tsx
```diff
@@ -0,0 +1,36 @@
+import { FC, useEffect, useState } from "react";
+import { Cluster } from "types/cluster";
+import { getSecondsSinceLastHeartbeat } from "util/helpers";
+
+interface Props {
+  cluster: Cluster;
+}
+
+const ClusterTimer: FC<Props> = ({ cluster }: Props) => {
+  const [seconds, setSeconds] = useState(getSecondsSinceLastHeartbeat(cluster));
```
mas-who, line 13: just trying to understand the logic here, this component is used for displaying how long it's been since the last heartbeat in `minutes:seconds` format? If so, I'm not sure this works. As long as we are on the details page, the timer will always increment and never reset which misinforms on the duration since last heartbeat.

mas-who (commented):
still have a few files to go, left some comments so long :)

ui/src/pages/clusters/ClusterDetail.tsx
```diff
@@ -0,0 +1,61 @@
...
+      {!isLoading && cluster && (
+        <Row>
+          <List
+            className="cluster-graphs"
```
mas-who: I would not reuse the `cluster-graphs` class for both `ClusterList` and `ClusterDetail`. Instead, we should follow similar approach in lxd-ui, have a `_cluster_list.scss` file and `_cluster_detail.scss` file to keep styling separate. Inside those files you will have a top level class `.cluster-list`, `.cluster-detail` with `.cluster-graphs` nested for applying page specific styling. In `_cluster_graphs.scss`, you can still have the `.cluster-graph` class with generic styling that applies to all.

ui/src/sass/_cluster_detail_metrics.scss
```diff
@@ -0,0 +1,19 @@
+// Cluster Detail Metrics SCSS
+.cluster-detail-metrics {
```
mas-who: let's put styling here inside the `_cluster_detail.scss` file as mentioned in the previous comment.

ui/src/sass/_cluster_graphs.scss
```diff
@@ -29,3 +29,21 @@ $donut-size: 6.5rem;
     fill: $memory-color;
   }
 }
+
+.cluster-countdown {
```
mas-who: these additional styling should probably all move into the `_cluster_detail.scss` file

ui/src/pages/clusters/metrics/ClusterNodeGraph.tsx
mas-who: can we rename this file to `ClusterDetailNodeGraph.tsx`? Let's follow the same convention in lxd ui where if a component exists only in a single place, we prefix the component name with the page name. In this case, `NodeGraph` only exist in the cluster detail page. Then, can we apply the same mental model for the other files as well?

ui/src/util/helpers.tsx
```diff
@@ -110,3 +115,15 @@ export const convertToISOFormat = (datetimeLocalString: string) => {
   // Return the ISO 8601 formatted string
   return date.toISOString();
 };
+
+// this works only for words that form the plural by adding an "s" at the end
+export const pluralizeWord = (word: string, count: number): string => {
+  return count === 1 ? word : `${word}s`;
+};
+
+export const statusCount = (
```
mas-who, line 124: this is an optimisation, you can decide if you want to add it or not. Instead calling this function on `instance_statuses` or `member_statuses`, we could have a function that internally holds a lookup in the form of:
```
{
  [status: string]: {
     color: string;
     count: number; // default 0
  }
}
```
Then the function would take in `instance_statuses` or `member_statuses`, iterate the statuses list, set the `count` field matched by status key. Then once all the status count values have been set, then iterate the lookup once to generated the list needed for the `DoughnutChart`. Time complexity for this approach would be O(n) + k where k is the number of possible statuses. The current approach time complexity is O(nk) since we are iterating the input for every possible status.

mas-who (commented):
Finally got through the rest of the files, I think we should neaten up the structure of how our css styles are organised. Let me know if you want to do a quick call to align.

mas-who:
Some thoughts on the visuals
Screenshot from 2024-08-02 10-42-46

1. The gap seems a bit too wide IMO between the timer and the next graph. Perhaps reduce the min-width set on the meter titles? Also we could remove the margin-right set by the `List` component and rely on the grid gap styling.
2. For the doughnut chart, can we center align the graph with the legend along the y-axis?

Screenshot from 2024-08-02 10-48-09
For the responsive view, this looks quite poor. We could use css grid to align the grid items, or an easier solution could be to set the flex direction as `column` using media query for smaller screen sizes (maybe center aligned along the x-axis).

ui/src/pages/clusters/ClusterDetail.tsx
Kxiru: Of course.

ui/src/pages/clusters/metrics/ClusterTimer.tsx
Kxiru: Alternatively we can call getSeconds...() every second?

mas-who: Calling the `getScondsSinceLastHearbeat` function only looks at a data point that's fixed in time. If the backend receives another status update from lxd, the UI wouldn't know about it unless it makes a request to the backend. We would need to set up the timer to repeatedly call the backend (at 1 minute intervals) to make this work

Kxiru:
Hiya, @mas-who !
Thanks for all of your suggestions. Please could you review my implementation of them in this latest commit.

This new commit comes with the exception of implementing the following:

1. A call to the backend every minute to re-render the latest minutes and seconds since last seen for each cluster.
2. Your optional suggestion to optimise the delivery of the doughnut segments through key graphs such as ClusterDetailNode and ClusterDetailInstance. Although I was able to create a working implementation of your suggestion, I would like to sync with you on this as the current implementation serves not only the doughnut segment delivery, but also the Doughnut legend delivery, located in the same file, and the suggested implementation complicates that secondary function. If possible, let's have a chat about this whenever you are available.

Thanks again for the review!

ui/src/sass/_cluster_detail.scss
```diff
@@ -0,0 +1,61 @@
...
+  .cluster-disk-bar {
+    fill: $disk-color;
+  }
+
+  .cluster-memory-bar {
+    fill: $memory-color;
+  }
+}
+
+.cluster-countdown {
```
mas-who:
```suggestion
.cluster-detail-countdown {
```

ui/src/sass/_cluster_detail.scss
```diff
@@ -0,0 +1,61 @@
...
+  justify-content: center;
+  text-align: center;
+  width: 100px;
+}
+
+.meter-container {
```
mas-who: `meter-container` and `meter-text` seems to be unused?

ui/src/sass/_cluster_detail.scss
```diff
@@ -0,0 +1,61 @@
+.cluster-detail-metrics {
```
mas-who: this class can be moved within `.cluster-detail-graphs`

ui/src/sass/_cluster_detail.scss
mas-who: also can move within `cluster-detail-graphs`

ui/src/sass/_cluster_detail.scss
```diff
@@ -0,0 +1,61 @@
...
+  .cluster-doughnut-graph__legend {
+    margin-left: $sph--small;
+  }
+
+  ul {
+    list-style-type: none;
+    margin-left: 0;
+  }
+
+  .cluster-disk-bar {
```
mas-who: `.cluster-disk-bar` and `.cluster-memory-bar` unused in the context of `cluster-detail-graphs`?

ui/src/pages/clusters/metrics/ClusterDetailNodeGraph.tsx
```diff
@@ -0,0 +1,83 @@
...
+  function getPercentageString(portion: number): ReactNode {
+    return (
+      <>
+        <b>{portion}</b> {`(${Math.floor((portion / totalNodes) * 100)}%) `}
+      </>
+    );
+  }
+
+  return (
+    <div className="cluster-detail-graphs">
```
mas-who: this is a duplicate use of `cluster-detail-graphs`, previously seen in `ClusterDetail`

ui/src/sass/_cluster_detail.scss
```diff
@@ -0,0 +1,61 @@
...
+
+  .cluster-doughnut-graph__chart {
+    width: $donut-size;
+  }
+
+  .cluster-doughnut-graph__legend {
+    margin-left: $sph--small;
+  }
+
+  ul {
```
mas-who: `cluster-doughnut-graph__legend` targets the `ul` arleady, try avoid tag selectors in general. Also `margin-left: 0` can be replaced by vanilla class `u-no-margin--left`

ui/src/pages/clusters/metrics/ClusterTimer.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+  const getFormattedTimeLeft = () => {
+    const minutes = Math.floor(seconds / 60);
+    const displaySeconds = seconds % 60;
+
+    return `${minutes}:${displaySeconds < 10 ? `0${displaySeconds}` : displaySeconds}`;
+  };
+
+  return (
+    <div className="cluster-countdown">
```
mas-who:
```suggestion
    <div className="cluster-detail-countdown">
```

mas-who (dismissed):
Thanks for the changes, looking much better! A few comments regarding CSS styling

mas-who:
> Hiya, @mas-who ! Thanks for all of your suggestions. Please could you review my implementation of them in this latest commit.
> 
> This new commit comes with the exception of implementing the following:
> 
> 1. A call to the backend every minute to re-render the latest minutes and seconds since last seen for each cluster.
> 2. Your optional suggestion to optimise the delivery of the doughnut segments through key graphs such as ClusterDetailNode and ClusterDetailInstance. Although I was able to create a working implementation of your suggestion, I would like to sync with you on this as the current implementation serves not only the doughnut segment delivery, but also the Doughnut legend delivery, located in the same file, and the suggested implementation complicates that secondary function. If possible, let's have a chat about this whenever you are available.
> 
> Thanks again for the review!

@Kxiru I'm happy to move on without the function optimisation to not over complicate things :slightly_smiling_face: We can sync on the minutely call for the last seen metrics.

I think we should adjust the styling for the detail page graphs so that it looks a bit better for medium to small screen sizes.

Kxiru:
@mas-who Let me know what you think of the latest changes and thank you for the thorough review.

ui/src/pages/clusters/metrics/ClusterTimer.tsx
```diff
@@ -0,0 +1,49 @@
...
+  cluster: Cluster;
+}
+
+const ClusterTimer: FC<Props> = ({ cluster }: Props) => {
+  const [seconds, setSeconds] = useState(getSecondsSinceLastHeartbeat(cluster));
+
+  const { data } = useQuery({
+    queryKey: [queryKeys.clusterTimer],
+    queryFn: () => fetchCluster(cluster.name),
+    refetchInterval: 1000,
```
mas-who: this should be a minute, assuming `refetchInterval` is in `ms`, then it would be `60000`.

ui/src/pages/clusters/metrics/ClusterTimer.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+  useEffect(() => {
+    const timerId = setInterval(() => {
+      setSeconds((prev) => prev + 1);
+    }, 1000);
+    return () => clearInterval(timerId);
+  }, []);
+
+  useEffect(() => {
+    setSeconds(getSecondsSinceLastHeartbeat(cluster));
```
mas-who: this should operated on `data` and not `cluster`

ui/src/sass/_cluster_detail.scss
```diff
@@ -0,0 +1,44 @@
+.cluster-detail-graphs {
+  align-items: center;
+  display: flex;
+  margin-left: $sph--small;
+
+  .cluster-detail-graph {
```
mas-who: maybe call this `doughnut-graph-container`?

ui/src/sass/_cluster_detail.scss
```diff
@@ -0,0 +1,44 @@
...
+        text-align: right;
+      }
+
+      .meter-row__metrics {
+        width: 100%;
+      }
+    }
+  }
+
+  .cluster-doughnut-graph__chart {
```
mas-who: `doughnuit-graph` related classes should move to within the `doughnut-graph-container` class

ui/src/pages/clusters/metrics/ClusterTimer.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+interface Props {
+  cluster: Cluster;
+}
+
+const ClusterTimer: FC<Props> = ({ cluster }: Props) => {
+  const [seconds, setSeconds] = useState(getSecondsSinceLastHeartbeat(cluster));
+
+  const { data } = useQuery({
+    queryKey: [queryKeys.clusterTimer],
```
mas-who: we don't need a new query key for this since it's using the `fetchCluster` function, which can be mapped to `[queryKeys.cluster, cluster.name]`. Also remove `clusterTimer` from `queryKeys`.

ui/src/util/queryKeys.tsx
```diff
@@ -4,4 +4,5 @@ export const queryKeys = {
   memberConfigOptions: "memberConfigOptions",
   tokens: "tokens",
   server: "server",
+  clusterTimer: "clusterTimer",
```
mas-who: this can be removed

ui/src/util/queryKeys.tsx
```diff
@@ -1,5 +1,6 @@
 export const queryKeys = {
   clusters: "clusters",
+  cluster: "cluster",
```
mas-who: no need for this key.

mas-who (approved):

PR #76, opened 2024-07-31
feat: [WD-13807] added Breadcrumb Component

Kxiru:
feat: [WD-13807] added Breadcrumb Component

image

[WD-13807]: https://warthogs.atlassian.net/browse/WD-13807?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ui/src/components/BaseLayout.tsx
+++ b/ui/src/components/BaseLayout.tsx
@@ -1,4 +1,4 @@
-import { FC, ReactNode } from "react";
+import { ElementType, FC, ReactNode } from "react";
 import { AppMain, Panel } from "@canonical/react-components";
 
 interface Props {
@@ -7,6 +7,8 @@ interface Props {
   children: ReactNode;
   mainClassName?: string;
   contentClassName?: string;
+  titleClassName?: string;
+  titleComponent?: ElementType;
 }
 
 const BaseLayout: FC<Props> = ({
@@ -15,6 +17,8 @@ const BaseLayout: FC<Props> = ({
   children,
   mainClassName,
   contentClassName,
+  titleClassName,
+  titleComponent,
 }: Props) => {
   return (
     <AppMain className={mainClassName}>
@@ -23,6 +27,8 @@ const BaseLayout: FC<Props> = ({
         controls={controls}
         wrapContent={true}
         contentClassName={contentClassName}
+        titleClassName={titleClassName}
+        titleComponent={titleComponent}
       >
         {children}
       </Panel>
--- a/ui/src/components/BreadcrumbHeader.tsx
+++ b/ui/src/components/BreadcrumbHeader.tsx
@@ -1,0 +1,40 @@
+import { Tooltip } from "@canonical/react-components";
+import { FC, ReactNode } from "react";
+
+interface Props {
+  name: string;
+  parentItems?: ReactNode[];
+}
+
+const BreadCrumbHeader: FC<Props> = ({ name, parentItems }: Props) => {
+  return (
+    <nav
+      className="p-breadcrumbs p-breadcrumbs--large"
+      aria-label="Breadcrumbs"
+    >
+      <ol className="p-breadcrumbs__items">
+        {parentItems
+          ? parentItems.map((item, key) => (
+              <li
+                className="p-heading--4 u-no-margin--bottom continuous-breadcrumb"
+                key={key}
+              >
+                {item}
+              </li>
+            ))
+          : null}
+
+        <li
+          className="p-heading--4 u-no-margin--bottom name continuous-breadcrumb"
+          title={name}
+        >
+          <Tooltip message={name} position="btm-left">
+            {name}
+          </Tooltip>
+        </li>
+      </ol>
+    </nav>
+  );
+};
+
+export default BreadCrumbHeader;
--- a/ui/src/pages/clusters/ClusterList.tsx
+++ b/ui/src/pages/clusters/ClusterList.tsx
@@ -20,7 +20,7 @@ const ClusterList: FC = () => {
   const tabs: string[] = ["Active", "Pending", "Tokens"];
 
   return (
-    <BaseLayout title={"Clusters"} controls={<AddClusterButton />}>
+    <BaseLayout title="Clusters" controls={<AddClusterButton />}>
       <Row>
         <List
           className="cluster-graphs"
--- a/ui/src/sass/styles.scss
+++ b/ui/src/sass/styles.scss
@@ -1,3 +1,4 @@
+@use "sass:map";
 @import "settings";
 @import "vanilla-framework/scss/build";
 @import "./breakpoints";
@@ -15,6 +16,7 @@
 @include vf-p-icon-export;
 @include vf-p-icon-edit;
 @include custom-p-icon;
+@import "breadcrumbs";
 @import "cluster_graphs";
 @import "cluster_list";
 @import "clustercpu";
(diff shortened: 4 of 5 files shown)


ui/src/sass/styles.scss
```diff
@@ -57,3 +57,42 @@ body {
...
+  .p-breadcrumbs__items {
+    margin-bottom: 0;
+
+    .continuous-breadcrumb {
+      display: inline;
+    }
+  }
+
+  .p-breadcrumbs__item {
+    // font-size: #{map.get($font-sizes, h4-mobile)}rem;
```
mas-who: why is this removed from the original lxd-ui styling?

ui/src/sass/styles.scss
```diff
@@ -57,3 +57,42 @@ body {
 .p-notification__message {
   word-break: break-word;
 }
+
+.p-breadcrumbs--large {
```
mas-who: let's move bread crumb styling into its own file `_breadcrumbs.scss`.

ui/src/components/BreadcrumbHeader.tsx
```diff
@@ -0,0 +1,44 @@
...
+  parentItems: ReactNode[];
+}
+
+const BreadCrumbHeader: FC<Props> = ({
+  name,
+  titleClassName,
+  parentItems,
+}: Props) => {
+  return (
+    <div className={classNames(titleClassName)}>
```
mas-who: Be careful with application layout here. This `div` should have the class `p-panel__title` and there should be a parent `div` with the class `p-panel__header`. Not sure if you have these classes already where you are using this component, but I think it's better to keep then contained here.

ui/src/components/BreadcrumbHeader.tsx
```diff
@@ -0,0 +1,44 @@
...
+      <nav
+        key="breadcrumbs"
```
mas-who: I don't think this key is needed. We only need to add keys to components in a list when doing a mapped render.

ui/src/components/BreadcrumbHeader.tsx
```diff
@@ -0,0 +1,44 @@
...
+            <li
+              className="p-heading--4 u-no-margin--bottom continuous-breadcrumb"
+              key={key}
+            >
+              {item}
+            </li>
+          ))}
+
+          <li
+            className="p-heading--4 u-no-margin--bottom name continuous-breadcrumb"
```
mas-who: the `name` class here comes from lxd-ui  in the `_rename_header.scss` file. I think it's important to apply the same in this project. I think it can all go into the breadcrumbs scss file. Also we should have a tooltip here similar to lxd-ui since the name may be very long and would be clipped if it overflows (see styling in lxd-ui).

mas-who (commented):
nice one splitting this out into a separate PR. It's much easier to focus on it. Left some feedback

ui/src/sass/styles.scss
Kxiru: it returned an error and I could not identify where it was imported from.

ui/src/components/BreadcrumbHeader.tsx
Kxiru: Should we leave the rename header styling and functionality out for now, until we are ready to implement it fully?

Happy to add the tooltip! (and remove name)

ui/src/components/BreadcrumbHeader.tsx
Kxiru: Since they're passed through as the prop to BaseLayout and therefore AppMain => Panel => It should inherit that styling.

ui/src/sass/styles.scss
mas-who: I think it's the sass map function that we are not importing causing the issue. Try put `@use "sass:map";` at the top of the file where you use `map.get`.

ui/src/components/BreadcrumbHeader.tsx
mas-who: I think we don't need to move all the styling over. The most important thing is to check what happens when the name is very long. At least we need to cater for text overflow, since for the current use case, cluster name is set by a user so we don't know how long it can get.

ui/src/components/BreadcrumbHeader.tsx
mas-who: right, in that case I don't think you need this `div`. You can use the `titleComponent`, `title`, and `titleClassName` props on the `Panel` component from react-components to achieve the same layout as lxd-ui. You'd have to adjust `BaseLayout` a bit though since currently it does not allow input of any other panel props.

ui/src/components/BreadcrumbHeader.tsx
```diff
@@ -0,0 +1,44 @@
...
+              className="p-heading--4 u-no-margin--bottom continuous-breadcrumb"
+              key={key}
+            >
+              {item}
+            </li>
+          ))}
+
+          <li
+            className="p-heading--4 u-no-margin--bottom continuous-breadcrumb"
+            title={`${name}`}
```
mas-who:
```suggestion
            title={name}
```

ui/src/components/BreadcrumbHeader.tsx
Kxiru: I see what you mean. Please let me know what you think of this next review.

ui/src/components/BreadcrumbHeader.tsx
```diff
@@ -0,0 +1,42 @@
...
+        {parentItems ? (
+          parentItems.map((item, key) => (
+            <li
+              className="p-heading--4 u-no-margin--bottom continuous-breadcrumb"
+              key={key}
+            >
+              {item}
+            </li>
+          ))
+        ) : (
```
mas-who: better to set false condition for ternary as `null`

ui/src/components/BaseLayout.tsx
```diff
@@ -23,6 +23,7 @@ const BaseLayout: FC<Props> = ({
         controls={controls}
         wrapContent={true}
         contentClassName={contentClassName}
+        titleClassName="base-layout-header"
```
mas-who: I don't see this class defined anywhere? Also I think `titleClassName` should be passed in as a prop in case if we need to customize styling for the related `title` prop.

The other prop from the `Panel` component we need here is [titleComponent](https://github.com/canonical/react-components/blob/main/src/components/Panel/Panel.tsx). It is by default set to `h4` in `Panel`, but if we want to recreate what's in lxd-ui, it needs to be a `div`.

ui/src/pages/clusters/ClusterList.tsx
```diff
@@ -20,7 +21,11 @@ const ClusterList: FC = () => {
   const tabs: string[] = ["Active", "Pending", "Tokens"];
 
   return (
-    <BaseLayout title={"Clusters"} controls={<AddClusterButton />}>
+    <BaseLayout
```
mas-who: we shouldn't do this change. For a top level page component like `ClusterList`, setting `title` to `BreadCrumbHeader` means we are introducing unnecessary html. Same thing goes for the change in `Settings`

mas-who (commented):
almost there, thanks for bring the `.name` styles over :)

Kxiru:
Thanks, @mas-who , let me know your thoughts now :)

mas-who:
LGTM thanks for the changes :+1:

mas-who (approved):

PR #77, opened 2024-08-01
Add UI e2e tests [WD-13243]

mas-who:
## Done
Added tests for:
 - create join token
 - revoke join token
 - approve pending cluster (assert cluster exist in active clusters list)
 - delete pending cluster
 - edit manager configs (global.address)
 - edit member configs

### DIFF
--- a/ui/tests/settings.spec.ts
+++ b/ui/tests/settings.spec.ts
@@ -1,0 +1,74 @@
+import { test, expect } from "@playwright/test";
+import {
+  goToSettingsPage,
+  updateManagerSetting,
+  updateMemberSetting,
+  validateSettingValue,
+} from "./helpers/settings";
+
+const MANAGER_SETTINGS = [
+  "oidc.issuer",
+  "oidc.client.id",
+  "oidc.audience",
+  "global.address",
+];
+
+const MEMBER_SETTINGS = ["https_address", "external_address"];
+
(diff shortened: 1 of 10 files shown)


ui/tests/settings.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+  );
+  await expect(memberConfigRows.nth(1)).toContainText(
+    `${MEMBER_NAME}.${MEMBER_SETTINGS[1]}`,
+  );
+});
+
+// NOTE: we DO NOT want to edit the oidc settings since it will break the login flow
+test("edit manager setting global.address", async ({ page }) => {
+  await goToSettingsPage(page);
+  await updateManagerSetting(page, "global.address", "http://localhost:8080");
```
Kxiru: I wonder if for the update Manager & Member settings functions in helpers/settings, the line(s) ```await validateSettingValue(settingRow, content);``` should be outside of that function, a) to keep the purpose of the function to one thing as opposed to changing the setting and also verifying it, and b) to have more visibility in parent tests - such as edit manager settings - so as to show that a setting has been validated after it is changed, as currently the validation part is hidden from the perspective of the parent test. What do you think?

mas-who: That's a great idea! :100:

mas-who:
@Kxiru made the suggested change

Kxiru (approved):

PR #79, opened 2024-08-01
feat: [WD-12703] Added user info in navigation footer.

Kxiru:
Work done:
- Added user info to nav footer using useServer();

image

### DIFF
--- a/ui/src/components/Navigation.tsx
+++ b/ui/src/components/Navigation.tsx
@@ -12,11 +12,13 @@ import NavLink from "components/NavLink";
 import { useMenuCollapsed } from "context/menuCollapsed";
 import { isWidthBelow, logout } from "util/helpers";
 import { useAuth } from "context/auth";
+import { useServer } from "context/useServer";
 
 const Navigation: FC = () => {
   const isSmallScreen = () => isWidthBelow(620);
   const { menuCollapsed, setMenuCollapsed } = useMenuCollapsed();
   const { isAuthenticated } = useAuth();
+  const { data: identity } = useServer();
 
   const handleLogout = () => {
     logout();
@@ -106,6 +108,18 @@ const Navigation: FC = () => {
               </li>
             </ul>
             <ul className="p-side-navigation__list sidenav-bottom-ul">
+              <li className="p-side-navigation__item">
+                <div
+                  className="p-side-navigation__link"
+                  title={`${identity?.name} (${identity?.email})`}
+                >
+                  <Icon
+                    className="p-side-navigation__icon is-dark"
+                    name="profile"
+                  />
+                  <div className="u-truncate">{identity?.name}</div>
+                </div>
+              </li>
               <hr className="is-dark navigation-hr" />
               <li className="p-side-navigation__item">
                 <a


ui/src/context/useLoggedInUser.tsx
```diff
@@ -0,0 +1,15 @@
+import { useQuery } from "@tanstack/react-query";
+import { fetchServer } from "api/server";
+import { queryKeys } from "util/queryKeys";
+
+export const useLoggedInUser = () => {
```
mas-who: I think there is no need to have this hook. We can use the `useServer` hook to get user email and name. In lxd ui we had to make two separate API calls so this hook makes sense there.

ui/src/util/queryKeys.tsx
```diff
@@ -4,4 +4,5 @@ export const queryKeys = {
   memberConfigOptions: "memberConfigOptions",
   tokens: "tokens",
   server: "server",
+  identities: "identities",
```
mas-who: this key is not needed since the `server` key represents the same api query

ui/src/components/Navigation.tsx
```diff
@@ -12,11 +12,13 @@ import NavLink from "components/NavLink";
 import { useMenuCollapsed } from "context/menuCollapsed";
 import { isWidthBelow, logout } from "util/helpers";
 import { useAuth } from "context/auth";
+import { useLoggedInUser } from "context/useLoggedInUser";
 
 const Navigation: FC = () => {
   const isSmallScreen = () => isWidthBelow(620);
   const { menuCollapsed, setMenuCollapsed } = useMenuCollapsed();
   const { isAuthenticated } = useAuth();
+  const { loggedInUserName, loggedInUserID } = useLoggedInUser();
```
mas-who:
```suggestion
  const { name: loggedInUserName, email: loggedInUserID } = useServer();
```

ui/src/components/Navigation.tsx
```diff
@@ -106,6 +108,20 @@ const Navigation: FC = () => {
               </li>
             </ul>
             <ul className="p-side-navigation__list sidenav-bottom-ul">
+              {isAuthenticated && (
```
mas-who: no need to check `isAuthenticated` here. In cluster manager, if the user isn't logged in, we do not show the navigation at all.

Kxiru:
Thanks for the review, @mas-who . Let me know what you think of these changes.

ui/src/components/Navigation.tsx
```diff
@@ -12,12 +12,20 @@ import NavLink from "components/NavLink";
...
+import { useQuery } from "@tanstack/react-query";
+import { fetchServer } from "api/server";
+import { queryKeys } from "util/queryKeys";
 
 const Navigation: FC = () => {
   const isSmallScreen = () => isWidthBelow(620);
   const { menuCollapsed, setMenuCollapsed } = useMenuCollapsed();
   const { isAuthenticated } = useAuth();
 
+  const { data: identity } = useQuery({
```
mas-who: You can use the `useServer` hook here. The API query is exactly the same.

mas-who (approved):
LGTM :+1: please can you rebase main first before merging

PR #81, opened 2024-08-06
fix: remove microcluster.Microcluster from cluster manager state

mas-who:
## Done
- previously we had `microcluster.Microcluster` inside cluster manager state in order to access it from within endpoints. This is not needed since we can simply create it with `microcluster.App`. This PR removes `microcluster.Microcluster` from the cluster manager state.


roosterfish (approved):
Looking good.

PR #83, opened 2024-08-06
feat: [WD-13078] Added search and filter for cluster list

Kxiru:
Work done:
- Implemeneted Custom Layout
- Implemented Pageheader component
- Implemented ClusterSearchFilter component alongside all relevant types
- Added ClusterSearchFilter to ClusterList
- Added Search Logic and Filter Logic
- Added scss styling for the page header.
- Minor other scss changes for consistency

Notes:
- Amended generate_cluster script so that it generates null (0) values for some of the clusters, for testing purposes. Whilst you can filter by statuses, this was not initially available in the UI. To counter this, I amended the generate_cluster script such that there became a way for "Running" instance statuses and "Online" Node statuses to be 0. Now, we can filter by those two statuses (for testing).

image

image

image
- ^ Filter by Memory where Memory is >90%.
- Note the change in URL params.

image
- ^ Filter by Memory where Memory is >90% and Node status is online.

### DIFF
--- a/ui/src/components/CustomLayout.tsx
+++ b/ui/src/components/CustomLayout.tsx
@@ -1,0 +1,30 @@
+import { FC, ReactNode } from "react";
+import classnames from "classnames";
+import { AppMain } from "@canonical/react-components";
+
+interface Props {
+  header?: ReactNode;
+  children: ReactNode;
+  mainClassName?: string;
+  contentClassName?: string;
+}
+
+const CustomLayout: FC<Props> = ({
+  header,
+  children,
+  contentClassName,
+  mainClassName,
+}: Props) => {
+  return (
+    <AppMain className={mainClassName}>
+      <div className="p-panel">
+        {header}
+        <div className={classnames("p-panel__content", contentClassName)}>
+          {children}
+        </div>
+      </div>
+    </AppMain>
+  );
+};
+
+export default CustomLayout;
--- a/ui/src/pages/clusters/ClusterList.tsx
+++ b/ui/src/pages/clusters/ClusterList.tsx
@@ -1,7 +1,6 @@
 import { FC } from "react";
 import { List, Row } from "@canonical/react-components";
-import BaseLayout from "components/BaseLayout";
-import { useParams } from "react-router-dom";
+import { useParams, useSearchParams } from "react-router-dom";
 import TabLinks from "components/TabLinks";
 import ClusterListTokens from "./ClusterListTokens";
 import ClusterListActive from "./ClusterListActive";
@@ -11,35 +10,154 @@ import ClusterStatusGraph from "./metrics/ClusterStatusGraph";
 import NotificationRow from "components/NotificationRow";
 import ClusterDiskGraph from "./metrics/ClusterDiskGraph";
 import ClusterMemoryGraph from "./metrics/ClusterMemoryGraph";
+import ClusterSearchFilter from "pages/clusters/ClusterSearchFilter";
+import CustomLayout from "components/CustomLayout";
+import PageHeader from "components/PageHeader";
+import { useQuery } from "@tanstack/react-query";
+import { queryKeys } from "util/queryKeys";
+import { fetchClusters } from "api/clusters";
+import {
+  ClusterFilters,
+  toNumericUsagePercentiles,
+  hasAllMatchingStatuses,
+  hasAllUsagePercentileBands,
+} from "util/clusterFilter";
+import {
+  ClusterInstanceStatus,
+  ClusterNodeStatus,
+  ClusterPercentiles,
+} from "types/cluster";
 
 const ClusterList: FC = () => {
   const { activeTab } = useParams<{
     activeTab?: string;
   }>();
 
+  const [searchParams] = useSearchParams();
+
   const tabs: string[] = ["Active", "Pending", "Tokens"];
 
+  const { data: clusters = [], isLoading } = useQuery({
+    queryKey: [queryKeys.clusters],
+    queryFn: fetchClusters,
+  });
+
+  const filters: ClusterFilters = {
+    queries: searchParams.getAll("query"),
+    instanceStatuses: searchParams.getAll(
+      "instance-status",
+    ) as ClusterInstanceStatus[],
+    nodeStatuses: searchParams.getAll("node-status") as ClusterNodeStatus[],
+    memoryUsage: toNumericUsagePercentiles(
+      searchParams.getAll("memory-usage"),
+    ) as ClusterPercentiles[],
+    diskUsage: toNumericUsagePercentiles(
+      searchParams.getAll("disk-usage"),
+    ) as ClusterPercentiles[],
+  };
+
+  const filteredClusters = clusters.filter((item) => {
+    if (
+      //Query search by name
+      !filters.queries.every((q) => item.name.toLowerCase().includes(q))
+    ) {
+      return false;
+    }
+
+    if (
+      //Search by Clusters with more than one instance having a particular status
+      filters.instanceStatuses.length > 0 &&
+      !hasAllMatchingStatuses(filters.instanceStatuses, item.instance_statuses)
+    ) {
+      return false;
+    }
+
+    if (
+      //Search by Clusters with more than one node having a particular status
+      filters.nodeStatuses.length > 0 &&
+      !hasAllMatchingStatuses(filters.nodeStatuses, item.member_statuses)
+    ) {
+      return false;
+    }
+
+    if (
+      //Search by Cluster memory usage
+      filters.memoryUsage.length > 0 &&
+      !hasAllUsagePercentileBands(
+        filters.memoryUsage,
+        item.memory_usage,
+        item.memory_total_amount,
+      )
+    ) {
+      return false;
+    }
+
+    if (
+      //Search by Cluster disk usage
+      filters.diskUsage.length > 0 &&
+      !hasAllUsagePercentileBands(
+        filters.diskUsage,
+        item.disk_usage,
+        item.disk_total_size,
+      )
+    ) {
+      return false;
+    }
+
+    return true;
+  });
+
   return (
-    <BaseLayout title="Clusters" controls={<AddClusterButton />}>
+    <CustomLayout
+      header={
+        <PageHeader>
+          <PageHeader.Left>
+            <PageHeader.Title>Clusters</PageHeader.Title>
+
+            <PageHeader.Search>
+              {!activeTab && <ClusterSearchFilter />}
+            </PageHeader.Search>
+          </PageHeader.Left>
+
+          <PageHeader.BaseActions>
+            <AddClusterButton />
+          </PageHeader.BaseActions>
+        </PageHeader>
+      }
+    >
       <Row>
         <List
           className="cluster-graphs"
           inline
           items={[
-            <ClusterStatusGraph key="cluster-status-graph" />,
-            <ClusterDiskGraph key="cluster-disk-graph" />,
-            <ClusterMemoryGraph key="cluster-memory-graph" />,
+            <ClusterStatusGraph
+              key="cluster-status-graph"
+              clusters={filteredClusters}
+            />,
+            <ClusterDiskGraph
+              key="cluster-disk-graph"
+              clusters={filteredClusters}
+            />,
+            <ClusterMemoryGraph
+              key="cluster-memory-graph"
+              clusters={filteredClusters}
+            />,
           ]}
         />
         <TabLinks tabs={tabs} activeTab={activeTab} tabUrl="/ui/clusters" />
         <NotificationRow />
         <div>
-          {!activeTab && <ClusterListActive />}
+          {!activeTab && (
+            <ClusterListActive
+              clusters={filteredClusters}
+              isLoading={isLoading}
+            />
+          )}
           {activeTab === "pending" && <ClusterListPending />}
           {activeTab === "tokens" && <ClusterListTokens />}
         </div>
       </Row>
-    </BaseLayout>
+    </CustomLayout>
   );
 };
 
--- a/ui/src/pages/clusters/ClusterListActive.tsx
+++ b/ui/src/pages/clusters/ClusterListActive.tsx
@@ -1,6 +1,4 @@
 import { FC } from "react";
-import { fetchClusters } from "api/clusters";
-import { useQuery } from "@tanstack/react-query";
 import { MainTable, TablePagination } from "@canonical/react-components";
 import Loader from "components/Loader";
 import { Link } from "react-router-dom";
@@ -10,15 +8,15 @@ import { ClusterMemory } from "./metrics/ClusterMemory";
 import { ClusterDisk } from "./metrics/ClusterDisk";
 import { ClusterNodes } from "./metrics/ClusterNodes";
 import ClusterHeartbeat from "./metrics/ClusterHeartbeat";
-import { queryKeys } from "util/queryKeys";
 import ClusterStatus from "./metrics/ClusterStatus";
+import { Cluster } from "types/cluster";
 
-const ClusterListActive: FC = () => {
-  const { data: clusters = [], isLoading } = useQuery({
-    queryKey: [queryKeys.clusters],
-    queryFn: fetchClusters,
-  });
+type Props = {
+  clusters: Cluster[];
+  isLoading: boolean;
+};
 
+const ClusterListActive: FC<Props> = ({ clusters, isLoading }) => {
   const filteredClusters = clusters.filter(
     (cluster) => cluster.status == "ACTIVE",
   );
@@ -88,7 +86,7 @@ const ClusterListActive: FC = () => {
       <TablePagination
         data={tableRows}
         id="pagination"
-        itemName="cluster"
+        itemName=" active cluster"
         className="u-no-margin--top"
         aria-label="Table pagination control"
       >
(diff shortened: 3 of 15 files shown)


mas-who:
This looks good! Having filtering makes it much easier to navigate the cluster list :+1: 

Have some QA comments so far:

1. The search and filter should either be hidden or disabled if we are on the "Pending" or "Tokens" tab. Although we don't need to clear the filters.
2. When I filtered by instance status, I expect the number of clusters shown to be the same as unfiltered cluster list if I selected all possible instance statuses. But it seems like that's not the case, see below screenshot (the total number of clusters is 95 instead of 100). Same goes for the other filter categories except the memory and disk ones.
Screenshot from 2024-08-07 09-41-45
3. The graphs above should also be adjusted according to the filters. This will require implementing the same filtering logic inside the active clusters component as well as the graph components. I'd suggest create a utils function for this.
4. One more case we should test with the data. What happens if no clusters should be displayed due to the filters? You can adjust the `generate_clusters.sh` script for this by removing all instance statuses except for one as an example, just don't check in these changes.

ui/src/components/ClusterSearchFilter.tsx
```diff
@@ -0,0 +1,88 @@
+import { FC, memo } from "react";
```
mas-who: this component should be in `/pages/clusters`.

ui/src/components/ClusterSearchFilter.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { paramsFromSearchData } from "util/searchAndFilter";
+import {
+  instanceStatuses,
+  nodeStatuses,
+  usagePercentiles,
+} from "util/clusterFilter";
+
+export const QUERY = "query";
+export const NAME = "name";
+export const INSTANCE_STATUS = "instanceStatus";
```
mas-who: prefer snake casing for params in url e.g `instance-status`

ui/src/util/clusterFilter.tsx
```diff
@@ -0,0 +1,70 @@
+import {
+  ClusterInstanceStatuses,
+  ClusterNodeStatuses,
+  StatusDistribution,
+} from "types/cluster";
+
+export interface ClusterFilters {
+  queries: string[];
+  instance_statuses: ClusterInstanceStatuses[];
```
mas-who: camel case for these field names e.g. `instanceStatuses`

ui/src/types/cluster.d.ts
```diff
@@ -29,3 +29,15 @@ export interface MultiMeterValue {
   status: string;
   color: string;
 }
+
+export type ClusterInstanceStatuses =
```
mas-who: The type name should be singular and not plural since it's not a list

ui/src/util/clusterFilter.tsx
```diff
@@ -0,0 +1,70 @@
...
+  node_statuses: ClusterNodeStatuses[];
+  memory_usage: number[];
+  disk_usage: number[];
+}
+
+export const instanceStatuses: string[] = [
```
mas-who: we can use the the new type `ClusterInstanceStatus`

ui/src/util/clusterFilter.tsx
```diff
@@ -0,0 +1,70 @@
...
+  "Running",
+  "Stopped",
+  "Frozen",
+  "Error",
+];
+
+export const nodeStatuses: string[] = [
```
mas-who: we can use the the new type `ClusterNodeStatus`

ui/src/util/clusterFilter.tsx
```diff
@@ -0,0 +1,70 @@
...
+  "Online",
+  "Blocked",
+  "Offline",
+  "Evacuated",
+];
+
+export const usagePercentiles: string[] = [">50%", ">75%", ">80%", ">90%"];
```
mas-who: If the list can only have distinct values, then we shouldn't assign the `string[]` type to it. We should just let typescript to implicitly assign the type in this case.

ui/src/util/clusterFilter.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+export const enrichUsages = (percentiles: string[]): number[] => {
```
mas-who: type for the `percentiles` input could be strict here instead of `string[]`? Also I find the name of the function a bit confusing, I think what this function is doing is converting the percentile filters from string to numeric format. Perhaps `toNumericUsagePercentiles` is a bit clearer?

ui/src/util/clusterFilter.tsx
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  return enrichedUsages;
+};
+
+export const hasAllMatchingStatuses = (
+  filterStatuses: string[],
+  matchingStatuses: StatusDistribution[],
+) => {
+  return filterStatuses.every((status) =>
```
mas-who: applying `every` here means we have an `AND` condition for every status filter. I think the filters should behave as `OR` conditions within a single filter category and `AND` conditions across filter categories (same in lxd UI). This explains my QA comment where not all clusters are shown when selecting every status in the instance status filter category.

ui/src/util/clusterFilter.tsx
```diff
@@ -0,0 +1,70 @@
...
+    matchingStatuses.some(
```
mas-who: here instead of iterating every `matchingStatuses` for every `filterStatuses`. You can convert `matchingStatuses` to a object `{ [status: string]: number }` with keys being the status and values being the associated count. Then you can do something like the following:
```
filterStatuses.some((status) => filterStatusCountLookup[status] >= 1)
```

ui/src/util/clusterFilter.tsx
```diff
@@ -0,0 +1,70 @@
...
+      (match) => match.status === status && match.count >= 1,
+    ),
+  );
+};
+
+export const hasAllMatchingUsageClasses = (
```
mas-who: maybe `hasAllUsagePercentileBands`?

ui/src/util/clusterFilter.tsx
```diff
@@ -0,0 +1,70 @@
...
+  "Evacuated",
+];
+
+export const usagePercentiles: string[] = [">50%", ">75%", ">80%", ">90%"];
+
+export const enrichUsages = (percentiles: string[]): number[] => {
+  const enrichedUsages = [];
+
+  if (percentiles.includes(">50%")) {
+    enrichedUsages.push(50);
```
mas-who: we could use the normalised numerical values for the percentiles e,g. 0.5, 0.75 etc. We wouldn't have to apply a multiplier when comparing to these filters later

ui/src/util/clusterFilter.tsx
```diff
@@ -0,0 +1,70 @@
...
+) => {
+  return filterStatuses.every((status) =>
+    matchingStatuses.some(
+      (match) => match.status === status && match.count >= 1,
+    ),
+  );
+};
+
+export const hasAllMatchingUsageClasses = (
+  usageClasses: number[],
```
mas-who: maybe `usagePercentileBands`?

ui/src/util/clusterFilter.tsx
```diff
@@ -0,0 +1,70 @@
...
+  use: number,
+  total: number,
+) => {
+  const usagePercentage = Math.floor((use / total) * 100);
```
mas-who: no need to do this calculation if the filters are normalised. Related to a previous comment.

ui/src/util/clusterFilter.tsx
```diff
@@ -0,0 +1,70 @@
...
+  console.log(usagePercentage);
```
mas-who: remove console.log

ui/src/components/CustomLayout.tsx
```diff
@@ -0,0 +1,29 @@
...
+}
+
+const CustomLayout: FC<Props> = ({
+  header,
+  children,
+  mainClassName,
+  contentClassName,
+}: Props) => {
+  return (
+    <main className={classnames("l-main", mainClassName)}>
```
mas-who: Looking at the component, I think we can just use the `title` prop of `BaseLayout` to setup `PageHeader` instead of having the `CustomLayout` component.

ui/src/pages/clusters/ClusterList.tsx
```diff
@@ -11,16 +10,112 @@ import ClusterStatusGraph from "./metrics/ClusterStatusGraph";
...
 
 const ClusterList: FC = () => {
   const { activeTab } = useParams<{
     activeTab?: string;
   }>();
 
+  const [searchParams] = useSearchParams();
+
   const tabs: string[] = ["Active", "Pending", "Tokens"];
+  const { data: clusters = [], isLoading } = useQuery({
```
mas-who, line 40: why did we move this query here from `ClusterListActive`?

ui/src/pages/clusters/ClusterList.tsx
```diff
@@ -11,16 +10,112 @@ import ClusterStatusGraph from "./metrics/ClusterStatusGraph";
...
+    queryKey: [queryKeys.clusters],
+    queryFn: fetchClusters,
+  });
+
+  const filters: ClusterFilters = {
+    queries: searchParams.getAll("query"),
+    instance_statuses: searchParams.getAll(
```
mas-who: field names and url params need to be adjusted here based on changes in `clusterFilter.tsx`.

ui/src/pages/clusters/ClusterList.tsx
```diff
@@ -11,16 +10,112 @@ import ClusterStatusGraph from "./metrics/ClusterStatusGraph";
...
+  const filteredClusters = clusters.filter((item) => {
+    if (
+      //Query search by name
+      !filters.queries.every((q) => item.name.toLowerCase().includes(q))
+    ) {
+      return false;
+    }
+
+    if (
+      //Search by Clusters with more than one instance in a particular status
```
mas-who:
```suggestion
      //Search by Clusters with more than one instance having a particular status
```

ui/src/pages/clusters/ClusterList.tsx
```diff
@@ -11,16 +10,112 @@ import ClusterStatusGraph from "./metrics/ClusterStatusGraph";
...
+      filters.instance_statuses.length > 0 &&
+      !hasAllMatchingStatuses(filters.instance_statuses, item.instance_statuses)
+    ) {
+      return false;
+    }
+
+    if (
+      //Search by Clusters with more than one node in a particular status
```
mas-who:
```suggestion
      //Search by Clusters with more than one node having a particular status
```

ui/src/pages/clusters/ClusterList.tsx
```diff
@@ -11,16 +10,112 @@ import ClusterStatusGraph from "./metrics/ClusterStatusGraph";
...
+    ) {
+      return false;
+    }
+
+    return true;
+  });
 
   return (
-    <BaseLayout title="Clusters" controls={<AddClusterButton />}>
+    <CustomLayout
```
mas-who, line 111: see if you can use the `BaseLayout` component `PageHeader` passed to the `title` prop?

ui/src/pages/clusters/ClusterListActive.tsx
```diff
@@ -10,15 +8,15 @@ import { ClusterMemory } from "./metrics/ClusterMemory";
 import { ClusterDisk } from "./metrics/ClusterDisk";
 import { ClusterNodes } from "./metrics/ClusterNodes";
 import ClusterHeartbeat from "./metrics/ClusterHeartbeat";
-import { queryKeys } from "util/queryKeys";
 import ClusterStatus from "./metrics/ClusterStatus";
+import { Cluster } from "types/cluster";
 
-const ClusterListActive: FC = () => {
-  const { data: clusters = [], isLoading } = useQuery({
```
mas-who, line 17: We should keep the query in this component

mas-who (changes requested):
Some feedback on code.

ui/src/util/clusterFilter.tsx
Kxiru: Originally the percentiles input was strictly typed but in the way this is done in LXD-UI, a string array becomes populated with the original types as well as the enriched types, which are also strings. In this case, as the enriched types are NOT strings, I did not feel it appropriate to create an interface/type that featured two different data types. See [here](https://github.com/Kxiru/lxd-ui/blob/dd5e43ea556cd06636fa3de07ddf33b0ec99a9e7/src/util/instanceFilter.tsx#L19) to view how this is incorporated in LXD-UI.

ui/src/components/CustomLayout.tsx
Kxiru: image

Custom Layout also comes with relevant parent styling.

ui/src/pages/clusters/ClusterList.tsx
Kxiru: It was moved because clusters needs to initialise the ClusterActiveList as well as the filteredClusters, which is then passed through. ClusterListActive does not need to recall this api in my understanding. ClusterListActive only reflects the filtered clusters, or all active clusters (no filter applied).

ui/src/pages/clusters/ClusterList.tsx
Kxiru: Please see above. Also here is what it looks like without the aforementioned styling.
image

But it's not impossible! Let me know how you would like me to proceed.

ui/src/pages/clusters/ClusterListActive.tsx
Kxiru: Could you explain why?

Kxiru:
> This looks good! Having filtering makes it much easier to navigate the cluster list 👍
> 
> Have some QA comments so far:
> 
> 1. The search and filter should either be hidden or disabled if we are on the "Pending" or "Tokens" tab. Although we don't need to clear the filters.
> 2. When I filtered by instance status, I expect the number of clusters shown to be the same as unfiltered cluster list if I selected all possible instance statuses. But it seems like that's not the case, see below screenshot (the total number of clusters is 95 instead of 100). Same goes for the other filter categories except the memory and disk ones.
>    Screenshot from 2024-08-07 09-41-45
> 3. The graphs above should also be adjusted according to the filters. This will require implementing the same filtering logic inside the active clusters component as well as the graph components. I'd suggest create a utils function for this.
> 4. One more case we should test with the data. What happens if no clusters should be displayed due to the filters? You can adjust the `generate_clusters.sh` script for this by removing all instance statuses except for one as an example, just don't check in these changes.

RE:
1) Yes, this is correct, thank you.
2) The total clusters changing to 95 is correct. As given my changes in generateClusters sh, 5 instances will be generated that have no running instances. For nodes, this goes from 100 to 50. Unless I have misunderstood, please could you clarify?
3) This can be done simply by passing the filtered clusters through to the graph components as has been done for ClusterListActive (and minor amendments to the component props). Could you advise why you think separate utils would be necessary for this?
4) I have seen and tested this. In the table, the string "No clusters found matching this search." is returned (From ClusterListActive)

ui/src/util/clusterFilter.tsx
Kxiru: I think I understand what you are saying, however the AND logic between filter categories takes place in the filteredClusters const, whilst the OR logic within a single filter category is one in the hasAllMatchingStatuses function, already.

mas-who: as discussed, please would you reach out to @piperdeck to see if AND logic makes sense within the same filter category

ui/src/components/CustomLayout.tsx
mas-who: aren't those the same props in `BaseLayout`?

ui/src/pages/clusters/ClusterList.tsx
mas-who: as discussed, let's have the data in the `ClusterList` parent component, so we can reused the filtered data in children components such as the graphs

ui/src/pages/clusters/ClusterList.tsx
mas-who: let's get the filtered clusters as prop input to the graphs as discussed

ui/src/pages/clusters/ClusterListActive.tsx
mas-who: no longer relevant

ui/src/components/CustomLayout.tsx
Kxiru: It should as they are built into Panel, right? Here is BaseLayout.
image

mas-who: I see what you are saying, we can't apply custom style to the `p-panel__header` element. Alright then let's keep the `CustomLayout` component and just use `AppMain` for the `l-main` element.

ui/src/util/clusterFilter.tsx
mas-who: this is still `string[]`?

ui/src/util/clusterFilter.tsx
mas-who: still `string[]`?

mas-who (commented):
Almost there, some minor comments. Also just need to align on the filter logic for instance and node statuses, that should be a quick one, see if you can follow up with @piperdeck sometime today.

ui/src/components/CustomLayout.tsx
```diff
@@ -0,0 +1,28 @@
...
+  contentClassName?: string;
+}
+
+const CustomLayout: FC<Props> = ({
+  header,
+  children,
+  contentClassName,
+}: Props) => {
+  return (
+    <AppMain>
```
mas-who: we should still get `mainClassName` here similar to `BaseLayout`

mas-who (approved):

mas-who:
LGTM :+1:

PR #86, opened 2024-08-15
fix(cluster) simplify timer logic and ensure whole cluster detail pag…

edlerd:
…e refreshes, fix graph captions

### DIFF
--- a/ui/src/pages/clusters/metrics/ClusterTimer.tsx
+++ b/ui/src/pages/clusters/metrics/ClusterTimer.tsx
@@ -1,9 +1,6 @@
-import { useQuery } from "@tanstack/react-query";
-import { fetchCluster } from "api/clusters";
 import { FC, useEffect, useState } from "react";
 import { Cluster } from "types/cluster";
 import { getSecondsSinceLastHeartbeat } from "util/helpers";
-import { queryKeys } from "util/queryKeys";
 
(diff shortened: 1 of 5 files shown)


ui/src/pages/clusters/metrics/ClusterTimer.tsx
```diff
@@ -12,23 +9,12 @@ interface Props {
 const ClusterTimer: FC<Props> = ({ cluster }: Props) => {
   const [seconds, setSeconds] = useState(getSecondsSinceLastHeartbeat(cluster));
 
-  const { data: data } = useQuery({
```
mas-who, line 15: If we do the refetch at the `ClusterDetail` parent level, then all children components will re-render, is that okay?

edlerd: I think they should, because the graphs and stats will have changed.

mas-who: yeah that makes sense, LGTM then :+1:

mas-who (approved):

mas-who:
@edlerd code and QA LGTM, thanks for the fixes :+1: Please would you have a look at #87 which fixes the CI errors. Once those are merged then we can rebase here.

PR #94, opened 2024-09-12
chore: fix microcluster version to v2 LTS

mas-who:
## Done
- update microcluster dependency to v2 :)


mas-who:
## NB

microcluster v2 tags track lxd at specific commits that presumably will not change. This means we can't use `go update-gomod` to update dependencies since our lxd version should be fixed to the commit that the microcluster v2 tag requires.

edlerd (approved):
LGTM

PR #102, opened 2024-10-24
Remove dedicated raft part from snap build

edlerd:
# Done
- Remove dedicated raft part from snap build

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -24,46 +24,29 @@ apps:
 
 parts:
   dqlite:
-    after:
-      - raft
     source: https://github.com/canonical/dqlite
     source-type: git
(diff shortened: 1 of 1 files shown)


mas-who (approved):
LGTM :+1: I assume this change means the dqlite part can be configured to install raft with dependencies?

edlerd:
> LGTM 👍 I assume this change means the dqlite part can be configured to install raft with dependencies?

Yes, that is what @tomponline suggested in another review a while back and here I wanted to apply the simplification.

edlerd:
> LGTM 👍 I assume this change means the dqlite part can be configured to install raft with dependencies?

Yes, that is what @tomponline suggested in another review a while back and here I wanted to apply the simplification.

PR #103, opened 2024-10-24
experiment: Add PostgreSQL support

mas-who:
## Done
- This is an very rough initial POC implementation for adding Postgres support to cluster manager.
- The experiment is mostly based on the abstraction layer approach detailed in this [spec](https://docs.google.com/document/d/1dgrZwlMIJ2uIyrKwvrt2hAAS7SqJHOCsUlCa0ZFW3pE/edit?tab=t.0).
- Please excuse the rough code :sweat_smile: 

## Screenshot
Screenshot from 2024-10-24 16-14-38
Screenshot from 2024-10-24 16-15-01

### DIFF
--- a/Dockerfile.pg
+++ b/Dockerfile.pg
@@ -1,0 +1,9 @@
+FROM postgres:latest
+
+ENV POSTGRES_USER="admin"
+ENV POSTGRES_PASSWORD="admin"
+ENV POSTGRES_DB="lxd_cluster_manager"
+
+EXPOSE 5432
+
+CMD ["postgres"]
--- a/cmd/lxd-cluster-mgrd/main.go
+++ b/cmd/lxd-cluster-mgrd/main.go
@@ -70,6 +70,8 @@ func (c *cmdDaemon) run(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	usePG := true
+
 	clusterManagerState := state.New(m)
 
 	dargs := microcluster.DaemonArgs{
@@ -86,6 +88,13 @@ func (c *cmdDaemon) run(cmd *cobra.Command, args []string) error {
 
 	dargs.Hooks = &microState.Hooks{
 		PostBootstrap: func(ctx context.Context, clusterState microState.State, initConfig map[string]string) error {
+			dbManager, err := database.NewDBConnManager(ctx, clusterState, usePG, true)
+			if err != nil {
+				return err
+			}
+
+			clusterManagerState.SetDatabase(dbManager.DB)
+
 			return InitialiseControlListener(ctx, clusterState, m, initConfig)
 		},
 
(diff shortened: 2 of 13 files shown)


Dockerfile.pg
```diff
@@ -0,0 +1,9 @@
+FROM postgres:latest
```
mas-who, line 1: this is for testing with local development postgres database

cmd/lxd-cluster-mgrd/main.go
```diff
@@ -70,6 +70,8 @@ func (c *cmdDaemon) run(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	usePG := true
```
mas-who, line 73: we could take this in as a cli flag from the user

internal/api/remote_cluster_control.go
```diff
@@ -57,7 +57,7 @@ func remoteClustersStatusPost(managerState *state.ClusterManagerState) types.End
 			return response.SmartError(err)
 		}
 
-		err = clusterState.Database().Transaction(r.Context(), func(ctx context.Context, tx *sql.Tx) error {
+		err = managerState.Database.Transaction(r.Context(), func(ctx context.Context, tx *sql.Tx) error {
```
mas-who, line 60: For this experiment, I only adjusted the join tokens endpoint to use potgres database

internal/database/core_remote_cluster_tokens.mapper.go
```diff
@@ -12,43 +12,43 @@ import (
 
 	"github.com/canonical/lxd/lxd/db/query"
 	"github.com/canonical/lxd/shared/api"
-	"github.com/canonical/microcluster/v2/cluster"
 )
 
 var _ = api.ServerEnvironment{}
 
-var coreRemoteClusterTokenObjects = cluster.RegisterStmt(`
+var coreRemoteClusterTokenObjects = `
```
mas-who, line 19: I manually adjusted the CRUD functions to align with Postgres sql syntax here. For proper implementation, we probably will have separate dqlite and postgres functions or maybe have some logic to do small adjustments on a single CRUD function based on database connection.

mas-who:
If you want to try this out, make sure you have `docker` installed. Then from the root of the project run `scripts/create_pg_db.sh`. Once the postgres container is running, start up the project as usual.

internal/database/db_conn_manager.go
```diff
@@ -0,0 +1,56 @@
...
+	if usePostgres {
+		host := "localhost"
+		port := "5432"
+		user := "admin"
+		password := "admin"
+		dbName := "lxd_cluster_manager"
+		// TODO: fine tune connection pooling settings
+		maxIdleConns := 10
+		maxOpenConns := 10
+		connMaxLifetime := time.Duration(5) * time.Minute
```
mas-who, line 25: these would be provided as configs by the user

PR #140, opened 2025-01-22
feat: add cos-lite for local dev environment

mas-who:
## Done

- enable cos-lite for local kind development k8s cluster
- enable cos-lite for microk8s cluster in CI

## QA
 - Run `make install-juju`. This should install juju on your system.
 - Run `make dev`. This should setup the k8s cluster, with cm deployed and also cos-lite. Note if this is a fresh kind cluster, it will take a well before the cluster gets going. After the cluster is running, run `juju status` and check that all applications are active.
 - While the cluster is running, press `ctrl + C` and then run `kubectl get all` check that all cm related resources (under default namesapce) have been removed by skaffold. Also run `juju status` to check that cos-lite is still running fine.
 - Run `make dev` again, check that the deployment should be much quicker now, look through the logs to check that cos-lite deployment was skipped. Run `juju status` again, make sure cos-lite is still running.
 - Lastly, stop the cluster, run `make nuke`. Check that there are no clusters at all (i.e. no docker containers), the `cluster-manager` cloud is not there in juju (run `juju clouds`) and the `cm-controller` is no longer registered with juju (run `juju controllers).


.github/workflows/pr.yaml
```diff
@@ -120,23 +120,42 @@ jobs:
...
+          sudo chown -f -R $USER:$USER /var/snap/juju/current/microk8s/credentials/client.config
+          sudo juju bootstrap microk8s cm-controller
       
       - name: Deploy k8s resources
         run: make deploy-ci-k8s-cluster IMAGE_NAME=${{ needs.build-rock.outputs.image-name }}
+
+      - name: Deploy COS-lite
+        run: |
+          sudo juju add-model cos && sudo juju switch cos
+          sudo juju deploy cos-lite --trust
```
edlerd: I think all of this should better be in a single shell script that gets the image name as an input. Then we can reuse it below and also run it locally.

deployment/k8s/dev/dns/dns.yaml
```diff
@@ -0,0 +1,167 @@
...
+metadata:
+  name: coredns
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: coredns
+subjects:
+  - kind: ServiceAccount
+    name: coredns
+    namespace: kube-system
```
edlerd:
```suggestion
    namespace: kube-system

```

makefile
```diff
@@ -377,4 +388,45 @@ add-hosts:
...
+		juju add-model cos && juju switch cos; \
+		juju deploy cos-lite --trust; \
+	else \
+		echo "COS-Lite already deployed."; \
+	fi
+
+.PHONY: dev-clean-juju
+dev-clean-juju:
+	juju unregister cm-controller --no-prompt
+	juju remove-cloud cluster-manager
```
edlerd:
```suggestion
	juju remove-cloud cluster-manager

```

deployment/k8s/dev/metallb/metallb.yaml
edlerd, line 1: What is this massive file? Can we reference it instead of having our own copy?

deployment/k8s/dev/metallb/addresspool.yaml
```diff
@@ -0,0 +1,14 @@
+---
+apiVersion: metallb.io/v1beta1
```
edlerd, line 2: This seems like an unstable version.

deployment/k8s/dev/dns/dns.yaml
```diff
@@ -0,0 +1,167 @@
...
+        ready
+        log . {
+          class error
+        }
+        kubernetes cluster.local in-addr.arpa ip6.arpa {
+          pods insecure
+          fallthrough in-addr.arpa ip6.arpa
+        }
+        prometheus :9153
+        forward . 8.8.8.8 8.8.4.4
```
edlerd, line 35: We depend on a 3rd party here, shouldn't we depend on the DNS server from the upstream network instead?

.github/workflows/pr.yaml
```diff
@@ -120,23 +120,42 @@ jobs:
...
+          channel: '1.31-strict/stable'
           # load local rock image into k8s cluster
           # https://microk8s.io/docs/sideload
           sideload-images-path: '${{ github.workspace }}/docker-image'
+
+      - name: Configure microk8s
+        run: |
+          sudo microk8s config > ~/.kube/config
+          sudo microk8s enable hostpath-storage dns
+          IPADDR=$(ip -4 -j route get 2.2.2.2 | jq -r '.[] | .prefsrc')
```
edlerd: This seems like an odd way to get the ip address, no?

.github/workflows/pr.yaml
mas-who: great idea :+1:

deployment/k8s/dev/metallb/metallb.yaml
mas-who: So this is the metallb k8s yaml available [here](https://raw.githubusercontent.com/metallb/metallb/v0.14.9/config/manifests/metallb-native.yaml). We could deploy the file directly into the cluster by referencing the hosted url or using helm. However, there are specific resource configs that require changes so it makes sense to store it locally so we can commit those changes. An example of that would be the microk8s metallb [addon](https://github.com/canonical/microk8s-core-addons/blob/main/addons/metallb/metallb.yaml).

metallb is a load balancer that you could deploy to bare metal k8s clusters which can also be used for development purposes for external ingress mapping. The reason why we use it in this project is to provide an external load balancer for the cos-lite traefik ingress.

deployment/k8s/dev/metallb/addresspool.yaml
mas-who: this is setup according to the [metallb docs](https://metallb.io/configuration/_advanced_ipaddresspool_configuration/) for assigning address pools for the load balancer. There seems to be only beta api versions from their [docs](https://metallb.io/apis/). I can look into using something else either than metallb, but it is a widely used tool for local development in the absence of cloud provider load balancers. Also it is used by cos-lite with microk8s according to this [example](https://charmhub.io/topics/canonical-observability-stack/tutorials/install-microk8s ).

deployment/k8s/dev/dns/dns.yaml
mas-who: This is set according to the microk8s `dns` addon [resource](https://github.com/canonical/microk8s-core-addons/blob/main/addons/dns/coredns.yaml) which has parameters dynamically set by the enable [script](https://github.com/canonical/microk8s-core-addons/blob/main/addons/dns/enable). According to the logic there, it first looks for a `resolve.conf` file in the snap environment, if nothing is found then it looks in its host environment for the same file, if nothing is found then these generic name servers are used. You can also provide name servers as parameters to the enable script. However, I think for local development this is sufficient.

.github/workflows/pr.yaml
mas-who: To be honest I didn't look deeply into it but it is extracted from this [example](https://charmhub.io/topics/canonical-observability-stack/tutorials/install-microk8s) of setting up microk8s for cos-lite. I think the purpose of this is to identify the source IP that the host system would use for traffic to any address. So `2.2.2.2` is just a placeholder and can be any publicly accessible IP, not sure if there is a more elegant way to do this?

deployment/k8s/dev/metallb/metallb.yaml
edlerd: Can we reference it instead of including it here?

deployment/k8s/dev/dns/dns.yaml
edlerd: Can we use the microk8s add-on to avoid duplicating the coredns.yaml file from there in our repo?

deployment/k8s/dev/metallb/metallb.yaml
mas-who: Actually looking at it again, yes I think so. I haven't changed anything in the `metallb.yaml` so there is no config changes that's specific to our dev environment.

edlerd: I don't see any modifications made to the remote file.

deployment/k8s/dev/dns/dns.yaml
mas-who: the microk8s dns addon works in conjunction with the enable script which is run when you run `microk8s enable dns`, and before the dns resources are deployed to the k8s cluster. We could reference it but will need to write a script to adjust the content of the downloaded yaml, the script would need to be specific to our dev environment. I'm not sure if we should go that far since it's much simpler to just include the dns resources this way with values we need.

deployment/k8s/dev/metallb/metallb.yaml
mas-who: yeah exactly, will remove metallb.yaml and just deploy it with the remote reference

deployment/k8s/dev/dns/dns.yaml
edlerd: I meant: why can't we directly run `microk8s enable dns`? We shouldn't worry how it internally does things or pull its config into our repo.

mas-who: this is only for local dev environment, we are not running microk8s. In CI we are using microk8s, and it is currently running `enable dns`

deployment/k8s/cicd/deploy.sh
```diff
@@ -0,0 +1,39 @@
...
+}
+
+main() {
+  configure_microk8s
+  setup_juju_controller
+  deploy_k8s_resources
+  deploy_cos_lite
+}
+
+main
```
edlerd:
```suggestion
main

```

deployment/k8s/cicd/deploy.sh
edlerd, line 1: Shall this be named deploy-cos-on-microk8s.sh ?

edlerd (commented):
Thanks for the iteration, this looks much better now.  Two tiny suggestions below. Apart from that LGTM.

deployment/k8s/cicd/deploy.sh
mas-who: good idea! :+1:

edlerd (approved):
LGTM, thanks for the iterations :+1:

PR #147, opened 2025-02-11
Fix token format

edlerd:
# Done
- allow the dev domain in vite dev server
- ensure token format is compatible with microcloud expectation

### DIFF
--- a/internal/pkg/api/models/v1/remote_cluster_join_token.go
+++ b/internal/pkg/api/models/v1/remote_cluster_join_token.go
@@ -28,7 +28,7 @@ type RemoteClusterToken struct {
 type RemoteClusterTokenBody struct {
 	Secret      string    `json:"secret"`
 	ExpiresAt   time.Time `json:"expires_at,omitempty"`
-	Address     string    `json:"address"`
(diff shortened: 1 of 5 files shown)


internal/pkg/api/models/v1/remote_cluster_join_token.go
```diff
@@ -28,7 +28,7 @@ type RemoteClusterToken struct {
 type RemoteClusterTokenBody struct {
 	Secret      string    `json:"secret"`
 	ExpiresAt   time.Time `json:"expires_at,omitempty"`
-	Address     string    `json:"address"`
+	Addresses   []string  `json:"addresses"`
```
mas-who, line 31: I believe `addresses` was a list previously to cater for how microcluster works. But now cluster-connector will only be exposed via one entry point, does it still make sense to have a list here?

edlerd: This is mandatory to stay in line with the share join token data structure. It is okay, that we have a list with a single entry.

mas-who (approved):
LGTM :+1:

PR #170, opened 2025-03-27
feat(join) simplify to rely on raw secret and avoid hmac validation

edlerd:


internal/app/cluster-connector/api/v1/remote_cluster.go
```diff
@@ -92,17 +92,17 @@ func remoteClustersPost(rc types.RouteConfig) types.EndpointHandler {
...
-		if time.Now().After(token.Expiry) {
-			return response.Forbidden(fmt.Errorf("token has expired")).Render(w, r)
+		// check if tokenFromDb has expired
+		if time.Now().After(tokenFromDb.Expiry) {
+			return response.Forbidden(fmt.Errorf("tokenFromDb has expired")).Render(w, r)
 		}
 
-		hmacOK, err := auth.VerifyHMAC(payload, r, token.Secret)
-		if err != nil || !hmacOK {
+		isTokenValid := strings.EqualFold(payload.Token, tokenFromDb.Secret)
```
roosterfish: Based on the recent discussion to keep it consistent across LXD/Microcluster, the `payload.Token` field should contain the entire base64 encoded token, not only the secret.

roosterfish (changes requested):

internal/app/cluster-connector/api/v1/remote_cluster.go
edlerd: Oh I see. Adjusted to storing the full encoded token in here then so we can have it for comparison. Also, adjusted microcluster.

internal/pkg/database/schema/migrations/00001_db.sql
```diff
@@ -29,7 +29,7 @@ CREATE TRIGGER remote_clusters_updated_at_trigger
 
 CREATE TABLE IF NOT EXISTS remote_cluster_tokens (
     id SERIAL PRIMARY KEY,
-    secret TEXT NOT NULL,
+    encoded_token TEXT NOT NULL,
```
roosterfish, line 32: In LXD the corresponding column name is `metadata`. Just mentioning it here too.

roosterfish (approved):

PR #342, opened 2025-12-11
Add Prometheus metrics parsing tests

omarelkashef:
Added metrics tests for successful and unsuccessful parsing of Prometheus metrics.

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -9,10 +9,13 @@ import (
 
 var tests = []types.Test{
 	testRemoteClusterSuccess,
+	testRemoteClusterSuccessWithMetrics,
 	testRemoteClusterJoinInvalid,
 	testRemoteClusterJoinExpiredToken,
 	testRemoteClusterStatusNoCert,
 	testRemoteClusterStatusInvalidCert,
+	testRemoteClusterStatusUnsupportedMetricsType,
+	testRemoteClusterStatusMalformedMetrics,
 }
 
 func TestE2E(t *testing.T) {
--- a/test/e2e/remote_cluster_fail_status.go
+++ b/test/e2e/remote_cluster_fail_status.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"net/http"
+	"strings"
 	"testing"
 	"time"
 
@@ -14,23 +15,18 @@ import (
 
 func testRemoteClusterStatusNoCert(env *helpers.Environment) (testName string, testFunc func(t *testing.T)) {
 	return "lxd remote cluster status update with no certificate", func(t *testing.T) {
-		remoteClusterName := "remote_cluster_status_no_cert"
+		remoteClusterName := helpers.GetRandomName("remote_cluster_status_no_cert")
 		var condition string
 
 		{
 			condition = "Should fail status update request with no client certificate"
 
-			tokenData, err := helpers.CreateAndReturnRemoteClusterJoinToken(env, remoteClusterName, time.Time{})
+			tokenData, err := helpers.RegisterRemoteCluster(env, remoteClusterName)
 			if err != nil {
 				helpers.LogTestOutcome(t, condition, err)
 			}
 
-			err = sendJoinRequest(env, tokenData)
-			if err != nil {
-				helpers.LogTestOutcome(t, condition, err)
-			}
-
-			err = sendStatusUpdateNoCert(env, tokenData)
+			err = sendStatusUpdateNoCert(env, *tokenData)
 			if err != nil && err.Error() == "Forbidden" {
 				err = nil
 			} else {
@@ -47,27 +43,78 @@ func testRemoteClusterStatusNoCert(env *helpers.Environment) (testName string, t
 
 func testRemoteClusterStatusInvalidCert(env *helpers.Environment) (testName string, testFunc func(t *testing.T)) {
 	return "lxd remote cluster status update with an invalid certificate", func(t *testing.T) {
-		remoteClusterName := "remote_cluster_status_invalid_cert"
+		remoteClusterName := helpers.GetRandomName("remote_cluster_status_invalid_cert")
 		var condition string
 
 		{
 			condition = "Should fail status update request with an invalid certificate"
 
-			tokenData, err := helpers.CreateAndReturnRemoteClusterJoinToken(env, remoteClusterName, time.Time{})
+			tokenData, err := helpers.RegisterRemoteCluster(env, remoteClusterName)
 			if err != nil {
 				helpers.LogTestOutcome(t, condition, err)
 			}
 
-			err = sendJoinRequest(env, tokenData)
-			if err != nil {
-				helpers.LogTestOutcome(t, condition, err)
-			}
-
-			err = sendStatusUpdateInvalidCert(env, tokenData)
+			err = sendStatusUpdateInvalidCert(env, *tokenData)
 			if err != nil && err.Error() == "Not Found" {
 				err = nil
 			} else {
 				err = errors.New("expected not found error not received")
+			}
+
+			helpers.LogTestOutcome(t, condition, err)
+		}
+
+		env.RemoveRemoteClusterToken(remoteClusterName)
+		env.RemoveRemoteCluster(remoteClusterName)
+	}
+}
+
+func testRemoteClusterStatusUnsupportedMetricsType(env *helpers.Environment) (testName string, testFunc func(t *testing.T)) {
+	return "lxd remote cluster status update with unsupported metric type", func(t *testing.T) {
+		remoteClusterName := helpers.GetRandomName("remote_cluster_status_unsupported_metrics_type")
+		var condition string
+
+		{
+			condition = "Should fail status update request with an unsupported metric type"
+
+			tokenData, err := helpers.RegisterRemoteCluster(env, remoteClusterName)
+			if err != nil {
+				helpers.LogTestOutcome(t, condition, err)
+			}
+
+			err = sendStatusUpdateUnSupportedMetricType(env, *tokenData)
+			if err != nil && strings.Contains(err.Error(), "failed to parse Prometheus metrics") {
+				err = nil
+			} else {
+				err = errors.New("expected parsing error not received")
+			}
+
+			helpers.LogTestOutcome(t, condition, err)
+		}
+
+		env.RemoveRemoteClusterToken(remoteClusterName)
+		env.RemoveRemoteCluster(remoteClusterName)
+	}
+}
+
+func testRemoteClusterStatusMalformedMetrics(env *helpers.Environment) (testName string, testFunc func(t *testing.T)) {
+	return "lxd remote cluster status update with malformed metrics", func(t *testing.T) {
+		remoteClusterName := helpers.GetRandomName("remote_cluster_status_malformed_metrics")
+		var condition string
+
+		{
+			condition = "Should fail status update request with malformed metrics"
+
+			tokenData, err := helpers.RegisterRemoteCluster(env, remoteClusterName)
+			if err != nil {
+				helpers.LogTestOutcome(t, condition, err)
+			}
+
+			err = sendStatusUpdateMalformedMetrics(env, *tokenData)
+			if err != nil && strings.Contains(err.Error(), "failed to parse Prometheus metrics") {
+				err = nil
+			} else {
+				err = errors.New("expected parsing error not received")
 			}
 
 			helpers.LogTestOutcome(t, condition, err)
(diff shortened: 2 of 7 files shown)


test/e2e/remote_cluster_fail_status.go
```diff
@@ -78,6 +80,72 @@ func testRemoteClusterStatusInvalidCert(env *helpers.Environment) (testName stri
...
+			err = sendJoinRequest(env, tokenData)
+			if err != nil {
+				helpers.LogTestOutcome(t, condition, err)
+			}
+
+			err = sendStatusUpdateUnSupportedMetricType(env, tokenData)
+			if err != nil && strings.Contains(err.Error(), "failed to parse Prometheus metrics") {
+				err = nil
+			} else {
+				err = errors.New("expected not found error not received")
```
edlerd: The message here seems not indicating the situation correctly.

test/e2e/remote_cluster_fail_status.go
```diff
@@ -78,6 +80,72 @@ func testRemoteClusterStatusInvalidCert(env *helpers.Environment) (testName stri
...
+			err = sendJoinRequest(env, tokenData)
+			if err != nil {
+				helpers.LogTestOutcome(t, condition, err)
+			}
+
+			err = sendStatusUpdateMalformedMetrics(env, tokenData)
+			if err != nil && strings.Contains(err.Error(), "failed to parse Prometheus metrics") {
+				err = nil
+			} else {
+				err = errors.New("expected not found error not received")
```
edlerd: The message here seems not indicating the situation correctly.

test/e2e/remote_cluster_fail_status.go
```diff
@@ -118,3 +186,112 @@ func sendStatusUpdateInvalidCert(env *helpers.Environment, tokenData models.Remo
...
+		return err
+	}
+
+	metricsText := `# HELP http_requests_total Total HTTP requests
+	# TYPE http_requests_total counter
+	http_requests_total{method="GET",status="200"} 1024
+	# HELP http_request_duration_seconds A summary of the HTTP request durations in seconds.
+	# TYPE http_request_duration_seconds summary
+	http_request_duration_seconds{quantile="0.5"} 0.12
+	`
```
edlerd, line 178: Can you add a comment what is unsupported here?

test/e2e/remote_cluster_fail_status.go
```diff
@@ -118,3 +186,112 @@ func sendStatusUpdateInvalidCert(env *helpers.Environment, tokenData models.Remo
...
+			{Status: "active", Count: 1},
+			{Status: "inactive", Count: 2},
+		},
+		Metrics: metricsText,
+	}
+
+	var output models.RemoteClusterStatusPostResponse
+	path := api.NewURL().Scheme("https").Host(tokenData.Addresses[0]).Path("1.0", "remote-cluster", "status")
+	err = tlsClient.Query(ctx, http.MethodPost, path, input, &output, nil)
+	return err
```
edlerd: Can you extract this into a helper that receives the metrics to avoid duplicating it three times?
I think `sendStatusUpdate` is already generic, just needs a parameter for metrics.

test/e2e/remote_cluster_fail_status.go
```diff
@@ -78,6 +80,72 @@ func testRemoteClusterStatusInvalidCert(env *helpers.Environment) (testName stri
...
+
+			tokenData, err := helpers.CreateAndReturnRemoteClusterJoinToken(env, remoteClusterName, time.Time{})
+			if err != nil {
+				helpers.LogTestOutcome(t, condition, err)
+			}
+
+			err = sendJoinRequest(env, tokenData)
+			if err != nil {
+				helpers.LogTestOutcome(t, condition, err)
+			}
```
edlerd: This could be one helper to do both. Something like `helpers.RegisterCluster(env, remoteClusterName)`

test/e2e/remote_cluster_fail_status.go
omarelkashef: Oh, I realize now that this reads expected "not found" error  and not "did not find the expected"

test/e2e/remote_cluster_fail_status.go
```diff
@@ -118,3 +165,71 @@ func sendStatusUpdateInvalidCert(env *helpers.Environment, tokenData models.Remo
...
+		InstanceStatuses: []models.StatusDistribution{
+			{Status: "running", Count: 1},
+			{Status: "stopped", Count: 2},
+		},
+		MemberStatuses: []models.StatusDistribution{
+			{Status: "active", Count: 1},
+			{Status: "inactive", Count: 2},
+		},
+		Metrics: metricsText,
+	}
```
edlerd: Can we put this without the Metrics field into a helper to fetch? we can set the metrics here and in the 3 other places afterwards on fetching the object.

test/e2e/remote_cluster_success.go
```diff
@@ -94,7 +90,7 @@ func testRemoteClusterSuccess(env *helpers.Environment) (testName string, testFu
 
 		{
 			condition = "Should be able to receive a status update"
-			response, err := sendStatusUpdate(env, tokenData)
+			response, err := sendSuccessfulStatusUpdate(env, tokenData)
```
edlerd: I think we should keep one case without metrics, to cover the early return on `payload.Metrics` with this test.

We can have a similar test with valid metrics as an additional case. Wdyt?

edlerd (commented):
Good stuff, thanks for improving the test!

Two small things left, I'll look into how to fix the success case with metrics. We might have to disable that specific test before merging and re-enable it later on. But I'd isolate it into a new test case first.

test/e2e/remote_cluster_success.go
omarelkashef: Yes, absolutely, we need to test both cases.

test/e2e/remote_cluster_success.go
```diff
@@ -111,6 +108,26 @@ func testRemoteClusterSuccess(env *helpers.Environment) (testName string, testFu
...
+		// 	if !reflect.DeepEqual(response.ClusterManagerAddress, expected) {
+		// 		fmt.Println(response.ClusterManagerAddress)
+		// 		fmt.Println(expected)
+		// 		err = fmt.Errorf("invalid Cluster Manager address")
+		// 		helpers.LogTestOutcome(t, condition, err)
+		// 	}
+
+		// 	helpers.LogTestOutcome(t, condition, nil)
+		// }
+
```
edlerd: This should move into a new test, as we don't want to change existing test cases. The idea is to keep covering the case where no metrics are being sent. We can already include the new test that is sending metrics. Only comment out the "adding metrics" part.

test/e2e/e2e_test.go
```diff
@@ -13,6 +13,8 @@ var tests = []types.Test{
 	testRemoteClusterJoinExpiredToken,
 	testRemoteClusterStatusNoCert,
 	testRemoteClusterStatusInvalidCert,
+	testRemoteClusterStatusUnsupportedMetricsType,
+	testRemoteClusterStatusMalformedMetrics,
```
edlerd, line 18: We should also add a new test case to  `testRemoteClusterSuccessWithMetrics` here, even though it might be a no-op test for now.

test/helpers/remote_clusters.go
edlerd, line 1: Good work extracting those generic functions.

edlerd (commented):
Thanks for the iteration. I think there was a misunderstanding in how to add the new test case with metrics as deactivated new test. Some clarification below.

test/e2e/remote_cluster_success.go
edlerd: Now that https://github.com/canonical/microcloud-cluster-manager/pull/346 is merged, also the success test with sending metrics should be passing.

test/e2e/remote_cluster_success.go
```diff
@@ -157,87 +154,54 @@ func testRemoteClusterSuccess(env *helpers.Environment) (testName string, testFu
...
-	encodedToken, err := tokenData.Encode()
-	if err != nil {
-		return err
-	}
+			if !reflect.DeepEqual(response.ClusterManagerAddress, expected) {
+				fmt.Println(response.ClusterManagerAddress)
+				fmt.Println(expected)
+				err = fmt.Errorf("invalid Cluster Manager address")
+				helpers.LogTestOutcome(t, condition, err)
+			}
```
edlerd:
```suggestion
```
We don't need to check the response here like in the other test. Doing it once should be enough.

test/e2e/remote_cluster_success.go
```diff
@@ -157,87 +154,54 @@ func testRemoteClusterSuccess(env *helpers.Environment) (testName string, testFu
...
+			response, err := sendSuccessfulStatusUpdateWithMetrics(env, *tokenData)
+			if err != nil {
+				helpers.LogTestOutcome(t, condition, err)
+			}
 
-	tlsClient, err := helpers.NewTLSHTTPClient(api.URL{}, clientCert, clusterConnectorCertPublicKey, env.ClusterConnectorHost())
-	if err != nil {
-		return err
-	}
+			expected := env.ClusterConnectorHostPort()
```
edlerd:
```suggestion
```
Better validate here that Prometheus ingested the metrics we sent. As suggested, put some random string as instance name as part of the `sendSuccessfulStatusUpdateWithMetrics` payload.

In here we should query Prometheus similar to the examples below. The prometheus address we can get from the config enpoint of the app and the path would be `api/v1/query?query=lxd_cpu_seconds_total` -- without the `cos-prometheus-0`, as we don't use COS in ci any more.

image


image

edlerd (commented):
Thanks for the iteration, this looks pretty good already. Let's add the payload validation from prometheus to have the full end to end flow of metrics from sending and proxying all the way back to reading our own write covered.

test/e2e/remote_cluster_success.go
```diff
@@ -157,87 +155,53 @@ func testRemoteClusterSuccess(env *helpers.Environment) (testName string, testFu
...
+				helpers.LogTestOutcome(t, condition, err)
+			}
 
-	return tlsClient.Query(ctx, http.MethodPost, path, input, nil, nil)
-}
+			metricsText := `# HELP lxd_cpu_seconds_total Total CPU time used
+			# TYPE lxd_cpu_seconds_total gauge
+			lxd_cpu_seconds_total{instance="inst1",project="default"} 100.0
+			lxd_cpu_seconds_total{instance="inst2",project="default"} 200.0
+			lxd_cpu_seconds_total{instance="e2e-random-instance-1234",project="other"} 300.0
```
edlerd: Instead of `e2e-random-instance-1234`, we should use a randomized string. We should also remember the randomized string to check for it further down when checking the prometheus response.

test/e2e/remote_cluster_success.go
edlerd, line 1: Some issues from the linter on this and other files.

edlerd (commented):
Good work, the test is passing, two tiny nitpicks below, then this is good to merge from my side.

test/e2e/remote_cluster_success.go
```diff
@@ -157,87 +155,53 @@ func testRemoteClusterSuccess(env *helpers.Environment) (testName string, testFu
...
-	input := models.RemoteClusterPost{
-		ClusterName:        tokenData.ServerName,
-		ClusterCertificate: string(clientCert.PublicKey()),
-		Token:              encodedToken,
-	}
+			tokenData, err := helpers.RegisterRemoteCluster(env, remoteClusterName)
+			if err != nil {
+				helpers.LogTestOutcome(t, condition, err)
+			}
+			instanceName := helpers.GetRandomName("e2e-random-instance-1234")
```
edlerd:
```suggestion
			instanceName := helpers.GetRandomName("e2e-random-instance")
```
We only want one number suffix.

makefile
```diff
@@ -21,7 +21,7 @@ lint-backend:
 ifeq ($(shell command -v golangci-lint 2> /dev/null),)
 	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $$(go env GOPATH)/bin
 endif
-	golangci-lint run --timeout 10m
+	$$(go env GOPATH)/bin/golangci-lint run --timeout 10m
```
edlerd: The `$$(go env GOPATH)/bin` should be part of a local `$PATH` environment. We should call `golangci-lint` without path here.

```suggestion
	golangci-lint run --timeout 10m
```

edlerd (commented):
nearly perfect, two nitpicks below.

makefile
omarelkashef: It could not find the linter on my host machine so I figured this was needed

edlerd: I think you should update your path variable then so the os looks into the go bin directory.

edlerd (dismissed):
LGTM, thanks for adding coverage on the metrics!

edlerd (approved):

PR #360, opened 2025-12-19
forward metrics from all servers in a cluster to prometheus

edlerd:
relates to https://github.com/canonical/microcloud/pull/1166

### DIFF
--- a/internal/app/cluster-connector/api/v1/remote_cluster.go
+++ b/internal/app/cluster-connector/api/v1/remote_cluster.go
@@ -224,7 +224,7 @@ func remoteClusterStatusPost(rc types.RouteConfig) types.EndpointHandler {
 				return err
 			}
 
-			if payload.Metrics == "" {
+			if len(payload.ServerMetrics) == 0 {
 				return nil
 			}
 
@@ -233,16 +233,24 @@ func remoteClusterStatusPost(rc types.RouteConfig) types.EndpointHandler {
 				return nil
 			}
 
-			timeSeries, err := parsePrometheusMetrics(payload.Metrics, dbRemoteCluster.Name)
-			if err != nil {
-				logger.Log.Warnw("Failed to parse metrics", "remote cluster", remoteClusterID, "err", err)
-				return fmt.Errorf("failed to parse Prometheus metrics: %w", err)
-			}
+			for i := range payload.ServerMetrics {
+				serverMetrics := payload.ServerMetrics[i]
+				if serverMetrics.Service != "lxd" {
+					logger.Log.Warnw("Unsupported service metrics received, skipping.", "service", serverMetrics.Service, "remote cluster", remoteClusterID)
+					continue
+				}
 
-			err = forwardMetricsToPrometheus(timeSeries, rc)
-			if err != nil {
-				logger.Log.Warnw("Failed to forward metrics to Prometheus", "remote cluster", remoteClusterID, "err", err)
-				return fmt.Errorf("failed to forward metrics to Prometheus: %w", err)
+				timeSeries, err := parsePrometheusMetrics(serverMetrics.Metrics, dbRemoteCluster.Name)
+				if err != nil {
+					logger.Log.Warnw("Failed to parse metrics, skipping.", "remote cluster", remoteClusterID, "err", err)
+					continue
+				}
+
+				err = forwardMetricsToPrometheus(timeSeries, rc)
+				if err != nil {
+					logger.Log.Warnw("Failed to forward metrics to Prometheus", "remote cluster", remoteClusterID, "err", err)
+					return fmt.Errorf("failed to forward metrics to Prometheus: %w", err)
+				}
 			}
 
 			return nil
--- a/test/e2e/remote_cluster_fail_status.go
+++ b/test/e2e/remote_cluster_fail_status.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"net/http"
-	"strings"
 	"testing"
 	"time"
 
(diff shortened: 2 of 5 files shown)


internal/app/cluster-connector/api/v1/remote_cluster.go
```diff
@@ -233,16 +233,24 @@ func remoteClusterStatusPost(rc types.RouteConfig) types.EndpointHandler {
...
-			if err != nil {
-				logger.Log.Warnw("Failed to parse metrics", "remote cluster", remoteClusterID, "err", err)
-				return fmt.Errorf("failed to parse Prometheus metrics: %w", err)
-			}
-
-			err = forwardMetricsToPrometheus(timeSeries, rc)
-			if err != nil {
-				logger.Log.Warnw("Failed to forward metrics to Prometheus", "remote cluster", remoteClusterID, "err", err)
-				return fmt.Errorf("failed to forward metrics to Prometheus: %w", err)
+			for i := range payload.ServerMetrics {
```
omarelkashef, line 236: Can this loop be parallelized instead of waiting for every metrics entry to parse and forward to Prometheus?

edlerd: We can yes. This is not time-critical, as we hit this only in server to server traffic where one cluster is sending a heartbeat, and we handle the payload here. So I think keeping it simple and iterative is okay.

internal/app/cluster-connector/api/v1/remote_cluster.go
```diff
@@ -233,16 +233,24 @@ func remoteClusterStatusPost(rc types.RouteConfig) types.EndpointHandler {
...
+				serverMetrics := payload.ServerMetrics[i]
+				if serverMetrics.Service != "lxd" {
+					logger.Log.Warnw("Unsupported service metrics received, skipping.", "service", serverMetrics.Service, "remote cluster", remoteClusterID)
+					continue
+				}
+
+				timeSeries, err := parsePrometheusMetrics(serverMetrics.Metrics, dbRemoteCluster.Name)
+				if err != nil {
+					logger.Log.Warnw("Failed to parse metrics", "remote cluster", remoteClusterID, "err", err)
+					return fmt.Errorf("failed to parse Prometheus metrics: %w", err)
```
omarelkashef: Do we not want to continue processing/forwarding other metrics even if one fails?

edlerd: Good idea, Yes!

internal/app/cluster-connector/api/v1/remote_cluster.go
```diff
@@ -233,16 +233,24 @@ func remoteClusterStatusPost(rc types.RouteConfig) types.EndpointHandler {
...
+				timeSeries, err := parsePrometheusMetrics(serverMetrics.Metrics, dbRemoteCluster.Name)
+				if err != nil {
+					logger.Log.Warnw("Failed to parse metrics, skipping.", "remote cluster", remoteClusterID, "err", err)
+					continue
+				}
+
+				err = forwardMetricsToPrometheus(timeSeries, rc)
+				if err != nil {
+					logger.Log.Warnw("Failed to forward metrics to Prometheus", "remote cluster", remoteClusterID, "err", err)
+					return fmt.Errorf("failed to forward metrics to Prometheus: %w", err)
```
omarelkashef, line 252: Should we also continue here instead of returning?

test/e2e/remote_cluster_success.go
```diff
@@ -205,3 +199,79 @@ func testRemoteClusterSuccessWithMetrics(env *helpers.Environment) (testName str
...
+			input := helpers.CreateStatusPostData()
+			input.ServerMetrics = []models.ServerMetrics{
+				{
+					Member:  "member1",
+					Service: "lxd",
+					Metrics: metricsMember1,
+				},
+			}
+
+			instanceMember2 := helpers.GetRandomName("e2e-random-instanceMember1-2")
```
omarelkashef: This overwrites the metrics from member1 instead of adding to the array.

test/e2e/remote_cluster_fail_status.go
```diff
@@ -69,62 +68,6 @@ func testRemoteClusterStatusInvalidCert(env *helpers.Environment) (testName stri
 	}
 }
 
-func testRemoteClusterStatusUnsupportedMetricsType(env *helpers.Environment) (testName string, testFunc func(t *testing.T)) {
```
omarelkashef, line 72: Why are we removing the failure tests.

omarelkashef (commented):
Just a few more comments and the rest looks good.

internal/app/cluster-connector/api/v1/remote_cluster.go
edlerd: This is Prometheus being down, so I think we should return the error not to overload it with more requests.

test/e2e/remote_cluster_success.go
edlerd: oh yes, good point!

test/e2e/remote_cluster_fail_status.go
edlerd: We are not returning failures any more, but skip the metrics for servers that we can't parse from. Hence, I was removing those tests.

test/e2e/remote_cluster_success.go
```diff
@@ -205,3 +199,77 @@ func testRemoteClusterSuccessWithMetrics(env *helpers.Environment) (testName str
 		}
 	}
 }
+
+func testRemoteClusterSuccessWithMetricsMultipleServers(env *helpers.Environment) (testName string, testFunc func(t *testing.T)) {
```
omarelkashef: testRemoteClusterSuccessWithMetrics seems redundant now because the processing logic is the same for one or multiple servers.

omarelkashef (commented):
One last thing and I think it is ready to merge.

omarelkashef (approved):
Thank you for all the changes. Looks great

PR #361, opened 2025-12-19
fix: [WD-31358] handle suppressed errors

omarelkashef:
Handles and logs suppressed errors.

### DIFF
--- a/internal/app/cluster-connector/api/v1/remote_cluster.go
+++ b/internal/app/cluster-connector/api/v1/remote_cluster.go
@@ -344,7 +344,10 @@ func forwardMetricsToPrometheus(timeSeries []prompb.TimeSeries, rc types.RouteCo
 	defer resp.Body.Close()
 
 	if resp.StatusCode > 299 {
-		body, _ := io.ReadAll(resp.Body)
+		body, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return fmt.Errorf("failed to read Prometheus response body: %w", err)
+		}
 		return fmt.Errorf("failed to send metrics to Prometheus, status: %s, response: %s", resp.Status, string(body))
 	}
 
--- a/internal/app/cluster-connector/core/auth/middleware.go
+++ b/internal/app/cluster-connector/core/auth/middleware.go
@@ -4,6 +4,7 @@ import (
 	"net/http"
 
 	"github.com/canonical/lxd/lxd/response"
+	"github.com/canonical/microcloud-cluster-manager/internal/pkg/logger"
 	"github.com/canonical/microcloud-cluster-manager/internal/pkg/types"
 	"github.com/gorilla/mux"
 )
@@ -15,13 +16,19 @@ func AuthMiddleware(rc types.RouteConfig) mux.MiddlewareFunc {
 	middlewareFunc := func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			if verifier == nil || !ok {
-				_ = response.Forbidden(nil).Render(w, r)
+				err := response.Forbidden(nil).Render(w, r)
+				if err != nil {
+					logger.Log.Errorw("Failed rendering forbidden response due to invalid verifier: %w", err)
+				}
 				return
 			}
 
 			_, err := verifier.Auth(r.Context(), w, r)
 			if err != nil {
-				_ = response.Forbidden(nil).Render(w, r)
+				err := response.Forbidden(nil).Render(w, r)
+				if err != nil {
+					logger.Log.Errorw("Failed rendering forbidden response due to authentication error: %w", err)
+				}
 				return
 			}
 
--- a/internal/app/management-api/core/auth/oidc.go
+++ b/internal/app/management-api/core/auth/oidc.go
@@ -129,7 +129,10 @@ func (o *Verifier) Auth(ctx context.Context, w http.ResponseWriter, r *http.Requ
 	if err != nil {
 		// Cookies are present but we failed to decrypt them. They may have been tampered with, so delete them to force
 		// the user to log in again.
-		_ = o.setCookies(w, nil, uuid.UUID{}, "", "", true)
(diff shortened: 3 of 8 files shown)


internal/app/cluster-connector/api/v1/remote_cluster.go
```diff
@@ -344,7 +344,10 @@ func forwardMetricsToPrometheus(timeSeries []prompb.TimeSeries, rc types.RouteCo
 	defer resp.Body.Close()
 
 	if resp.StatusCode > 299 {
-		body, _ := io.ReadAll(resp.Body)
+		body, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return fmt.Errorf("failed to read response body: %w", err)
```
edlerd:
```suggestion
			return fmt.Errorf("failed to read Prometheus response body: %w", err)
```
I think we shall give a hint which response it relates to.

internal/app/cluster-connector/core/auth/middleware.go
```diff
@@ -15,13 +15,19 @@ func AuthMiddleware(rc types.RouteConfig) mux.MiddlewareFunc {
 	middlewareFunc := func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			if verifier == nil || !ok {
-				_ = response.Forbidden(nil).Render(w, r)
+				err := response.Forbidden(nil).Render(w, r)
+				if err != nil {
+					http.Error(w, "Forbidden", http.StatusForbidden)
```
edlerd: This is a middleware function. I am not sure if we can just use the `http.Error` helper here and in the other places. Its documentation states `the caller should ensure no further writes are done to w`, which I think we can't guarantee here. Because upstream caller might send further things into the writer.

Then again, I have no immediate better idea how to handle the error yet.

edlerd (commented):
One nitpick on the message and a more general question on handling errors in the middleware.

edlerd:
Please rebase on main branch as the CI pipeline was simplified and unblocked through changes in it.

internal/app/cluster-connector/core/auth/middleware.go
```diff
@@ -15,13 +16,19 @@ func AuthMiddleware(rc types.RouteConfig) mux.MiddlewareFunc {
 	middlewareFunc := func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			if verifier == nil || !ok {
-				_ = response.Forbidden(nil).Render(w, r)
+				err := response.Forbidden(nil).Render(w, r)
+				if err != nil {
+					logger.Log.Errorw("%w", err)
```
edlerd:
```suggestion
					logger.Log.Errorw("Failed rendering forbidden response due to invalid verifier: %w", err)
```
We should prefix all log lines with a unique clear string to make debugging easier.

This applies to various places that currently only log the plain error.

internal/app/management-api/core/auth/oidc.go
```diff
@@ -129,7 +129,10 @@ func (o *Verifier) Auth(ctx context.Context, w http.ResponseWriter, r *http.Requ
 	if err != nil {
 		// Cookies are present but we failed to decrypt them. They may have been tampered with, so delete them to force
 		// the user to log in again.
-		_ = o.setCookies(w, nil, uuid.UUID{}, "", "", true)
+		err := o.setCookies(w, nil, uuid.UUID{}, "", "", true)
+		if err != nil {
+			return false, fmt.Errorf("Failed to delete invalid login cookies: %w", err)
```
edlerd, line 134:
```suggestion
			return false, fmt.Errorf("Failed to set OIDC cookie: %w", err)
```
We need to make sure the logged error prefix is correct.

internal/app/management-api/core/auth/oidc.go
```diff
@@ -205,7 +208,10 @@ func (o *Verifier) Login(w http.ResponseWriter, r *http.Request, stateTokenStr s
 	err := o.ensureConfig(r.Context(), r)
 	if err != nil {
 		logger.Log.Info("AUTHN invalid OIDC configuration")
-		_ = response.ErrorResponse(http.StatusInternalServerError, fmt.Errorf("Login failed: %w", err).Error()).Render(w, r)
+		err := response.ErrorResponse(http.StatusInternalServerError, fmt.Errorf("Login failed: %w", err).Error()).Render(w, r)
+		if err != nil {
+			logger.Log.Errorw("Login failed: %w", err)
```
```suggestion
			logger.Log.Errorw("Failed to set error response due to invalid OIDC configuration: %w", err)
```
Please revise all other error prefixes that they state correctly the situation.

internal/app/management-api/core/auth/oidc.go
```diff
@@ -218,7 +224,10 @@ func (o *Verifier) Login(w http.ResponseWriter, r *http.Request, stateTokenStr s
 func (o *Verifier) Logout(w http.ResponseWriter, r *http.Request) {
 	err := o.setCookies(w, nil, uuid.UUID{}, "", "", true)
 	if err != nil {
-		_ = response.ErrorResponse(http.StatusInternalServerError, fmt.Errorf("Failed to delete login information: %w", err).Error()).Render(w, r)
+		err = response.ErrorResponse(http.StatusInternalServerError, fmt.Errorf("Failed to delete login information: %w", err).Error()).Render(w, r)
+		if err != nil {
+			http.Error(w, fmt.Errorf("Failed to delete login information: %w", err).Error(), http.StatusInternalServerError)
```
edlerd: This will very likely fail, as we already failed to set the error response on the writer. We should log here (and below) instead.

edlerd (commented):
Thanks for the iteration, this looks much better. Please revise the error prefixes to ensure they are correct.

internal/app/management-api/core/auth/oidc.go
edlerd: Ah wait, I think this one was correct.

edlerd (approved):
LGTM, thanks for handling all the errors :+1:

PR #385, opened 2026-01-22
feat: add rate limiting WD-20341

omarelkashef:
Add rate limiting middleware to POST `1.0/remote-clusters' endpoint.


internal/app/cluster-connector/core/rate_limit/middleware.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"net/http"
+	"time"
+
+	"github.com/canonical/microcloud-cluster-manager/internal/pkg/logger"
+	"github.com/canonical/microcloud-cluster-manager/internal/pkg/types"
+	"github.com/gorilla/mux"
+)
+
+func RateLimitMiddlewareFactory(rls *RateLimiterStore) func(rc types.RouteConfig) mux.MiddlewareFunc {
```
edlerd: Let's remove the factory pattern here and make this a plain `RateLimitMiddleware`. Similar to the `AuthMiddleware`. That should remove a bit of complexity from here and the usages.

internal/app/cluster-connector/api/v1/remote_cluster.go
```diff
@@ -25,6 +26,8 @@ import (
 	"github.com/prometheus/prometheus/prompb"
 )
 
+var RateLimiterStore = rate_limit.NewRateLimiterStore(1, 3, 30*time.Second, 10*time.Second, 2000, 30*time.Second)
```
edlerd: We don't need to make this configurable immediately, but let's introduce constants for the various configurations we hardcode here.

Actually, it would be good to expose this in the `Config` with those defaults. This is very likely something we want to customize. Though we need to find a way to access it -- possibly through the `rc.env.config`, like in the `AuthMiddleware`.

makefile
```diff
@@ -150,6 +150,10 @@ test-e2e:
 test-ui-e2e:
 	cd ui && CI=$(CI) npx playwright test $(if $(PROJECT),--project $(PROJECT))
 
+.PHONY: test-unit
+test-unit:
+	go test -count=1 -v ./test/unit | grep -P '(PASS|FAIL|RUN|===|---|\x{2713}|\x{2717})'
```
edlerd:
```suggestion
	go test -count=1 -v ./test/unit
```
Do we need the grep here?

edlerd (commented):
Good start, some thoughts and suggestions below.

makefile
omarelkashef: The test report gets flooded with the logs of the application itself which is not helpful to whoever is running the test.

internal/app/cluster-connector/core/rate_limit/middleware.go
omarelkashef: I am using a factory pattern because `RouteMiddleware` type only takes `RouteConfig` as parameter and we need to pass the `RateLimiterStore to `RateLimitMiddlewareFactory` somehow.

edlerd: Let's add the store to the `RouteConfig` object. Similar to `verifier, ok := rc.Auth.(*Verifier)` in the `AuthMiddleware` we can then fetch it from within. This might also help with initializing the store with non-hardcoded configuration.

makefile
edlerd: Let's keep the logs. The extra output helps future debugging in case something fails.

cmd/cluster-connector/cluster-connector.go
```diff
@@ -128,6 +130,18 @@ func Run() (err error) {
 
 	mtlsAuthenticator := auth.NewMtlsAuthenticator(db)
 
+	// =========================================================================
+	// Initialize rate limiting support
+	rateLimitRefillRate := rate.Limit(cfg.RateLimitRefillRate)
+	rateLimitBucketSize := cfg.RateLimitBucketSize
+	if cfg.Version == "development" {
+		rateLimitBucketSize = 1000
+	}
```
edlerd:
```suggestion
```
I don't think this should be here in the production code. We better set an environment variable to influence the config value in development environment. Or rely on the default value right away. Why change it at all?

cmd/cluster-connector/cluster-connector.go
```diff
@@ -128,6 +130,18 @@ func Run() (err error) {
...
+	rateLimitTTL := time.Duration(cfg.RateLimitTTLSeconds) * time.Second
+	rateLimitCleanupInterval := time.Duration(cfg.RateLimitCleanupIntervalSeconds) * time.Second
+	rateLimitLogInterval := time.Duration(cfg.RateLimitLogIntervalSeconds) * time.Second
```
edlerd: I think all of those conversions (and the ones above) should be done on initializing the config. So we can have the right unit of seconds in `cfg.RateLimitTTLSeconds`. Then this initialization step can directly forward it to `NewTokenBucketRateLimiter`, and the change here becomes a single line. Wdyt?

Or maybe an even better option, also reducing the number of arguments, is to keep the config as is and do `tokenBucketRateLimiter := rate_limit.NewTokenBucketRateLimiter(cfg)`. Then we do all of those conversions inside the instantiation function from the config fields.

internal/app/cluster-connector/core/rate_limit/middleware.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"net/http"
+
+	"github.com/canonical/lxd/lxd/response"
+	"github.com/canonical/microcloud-cluster-manager/internal/pkg/logger"
+	"github.com/canonical/microcloud-cluster-manager/internal/pkg/types"
+	"github.com/gorilla/mux"
+)
+
+func RateLimitMiddleware(rc types.RouteConfig) mux.MiddlewareFunc {
```
edlerd, line 12: this looks quite nice now, good work :+1:

internal/app/cluster-connector/core/rate_limit/rate_limiter.go
```diff
@@ -0,0 +1,147 @@
...
+// rateLimiterStore manages per-client limiters with cleanup.
+type TokenBucketRateLimiter struct {
+	mu              sync.Mutex
+	refillRate      rate.Limit
+	bucketSize      int
+	ttl             time.Duration // duration after which inactive clients are cleaned up
+	maxClients      int
+	clients         map[string]*TokenBucketClientLimiter
+	cleanupInterval time.Duration
+	logInterval     time.Duration // duration between logs to avoid I/O strain
```
edlerd: Can we add comments for `cleanupInterval`, `refillRate` and `bucketSize`? I think most fields here need a bit more description.

Also, I don't like the naming mismatch between `logInterval`, `cleanupInterval` and `ttl`. Can we find a more descriptive name for the client clean up -- maybe `clientActiveInterval`?

makefile
```diff
@@ -472,4 +476,4 @@ update-traefik-port:
 	    fi; \
 	    sleep 2; \
 	done
-	kubectl patch svc/traefik-lb -n cos -p '{"spec":{"ports":[{"port":80, "nodePort":30001}]}}'
+	kubectl patch svc/traefik-lb -n cos -p '{"spec":{"ports":[{"port":80, "nodePort":30001}]}}'
```
edlerd:
```suggestion
	kubectl patch svc/traefik-lb -n cos -p '{"spec":{"ports":[{"port":80, "nodePort":30001}]}}'

```

internal/app/cluster-connector/core/rate_limit/rate_limiter.go
```diff
@@ -0,0 +1,147 @@
...
+	"net"
+	"net/http"
+	"sync"
+	"time"
+
+	"github.com/canonical/microcloud-cluster-manager/internal/pkg/logger"
+	"golang.org/x/time/rate"
+)
+
+type TokenBucketClientLimiter struct {
```
edlerd:
```suggestion
type ClientLimiter struct {
```
Shall we drop the `TokenBucket` prefix here and everywhere else?

test/run-backend.sh
```diff
@@ -59,6 +59,12 @@ export MANAGEMENT_API_TLS_PATH="$KEY_DIR"
 export CLUSTER_CONNECTOR_TLS_PATH="$KEY_DIR"
 export CLUSTER_CONNECTOR_PORT=9000
 export PROMETHEUS_BASE_URL=http://$PROMETHEUS_ADDRESS/api/v1/write
+export RATE_LIMIT_REFILL_RATE=1
+export RATE_LIMIT_BUCKET_SIZE=3
+export RATE_LIMIT_TTL_SECONDS=60
+export RATE_LIMIT_CLEANUP_INTERVAL_SECONDS=60
+export RATE_LIMIT_LOG_INTERVAL_SECONDS=60
+export RATE_LIMIT_MAX_CLIENTS=1000
```
edlerd: Do we need those here, or can we rely on defaults?
```suggestion
```

internal/pkg/config/config.go
```diff
@@ -112,9 +119,36 @@ func LoadConfig() (*Config, error) {
...
+	}
+	rateLimitTTLSeconds, err := getEnvAsInt("RATE_LIMIT_TTL_SECONDS", 300)
+	if err != nil {
+		return nil, fmt.Errorf("invalid RATE_LIMIT_TTL_SECONDS: %w", err)
+	}
+	rateLimitCleanupIntervalSeconds, err := getEnvAsInt("RATE_LIMIT_CLEANUP_INTERVAL_SECONDS", 60)
+	if err != nil {
+		return nil, fmt.Errorf("invalid RATE_LIMIT_CLEANUP_INTERVAL_SECONDS: %w", err)
+	}
+	rateLimitLogIntervalSeconds, err := getEnvAsInt("RATE_LIMIT_LOG_INTERVAL_SECONDS", 300)
```
edlerd:
```suggestion
	rateLimitLogIntervalSeconds, err := getEnvAsInt("RATE_LIMIT_LOG_INTERVAL_SECONDS", 60)
```
Let's log at least once a minute.

internal/pkg/config/config.go
```diff
@@ -112,9 +119,36 @@ func LoadConfig() (*Config, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("invalid RATE_LIMIT_LOG_INTERVAL_SECONDS: %w", err)
+	}
+	rateLimitMaxClients, err := getEnvAsInt("RATE_LIMIT_MAX_CLIENTS", 1000)
```
edlerd:
```suggestion
	rateLimitMaxClients, err := getEnvAsInt("RATE_LIMIT_MAX_CLIENTS", 1_000_000)
```
I think this is too low. This puts an upper bound on how many clusters we can connect that send successful heartbeats. The memory footprint of one client should be small enough to securely raise it to 1 million.

edlerd (commented):
Good progress, thanks for making the rate limits configurable. Some more ideas to refine the code below.

internal/app/cluster-connector/core/rate_limit/rate_limiter.go
omarelkashef: do you mean to replace ttl with clientActiveInterval?

test/run-backend.sh
omarelkashef: As discussed, we will not be relying on defaults here.

internal/app/cluster-connector/core/rate_limit/rate_limiter.go
omarelkashef: I wanted to differentiate in case we adopt a different algorithm in the future, we can switch implementation easily.

internal/app/cluster-connector/core/rate_limit/rate_limiter.go
edlerd: Yes, I was thinking about a rename of the ttl field. Wdyt?

internal/app/cluster-connector/core/rate_limit/rate_limiter.go
edlerd: Is it relevant to the code calling this? Or should this stay an implementation secret that is not exposed here? I lean towards the latter, keeping it an implementation detail. Though the config exposes some of this anyhow.

internal/app/cluster-connector/core/rate_limit/rate_limiter.go
omarelkashef: Yes the clearer and consistent naming is better!

internal/app/cluster-connector/core/rate_limit/rate_limiter.go
omarelkashef: Good point, this data structure is not exposed.

internal/app/cluster-connector/core/rate_limit/rate_limiter.go
```diff
@@ -0,0 +1,147 @@
...
+)
+
+type ClientLimiter struct {
+	limiter    *rate.Limiter
+	lastSeen   time.Time
+	lastLogged time.Time
+}
+
+// rateLimiterStore manages per-client limiters with cleanup.
+type TokenBucketRateLimiter struct {
```
edlerd: The comment seems out of sync. Also, can this be `RateLimiter`?

internal/app/cluster-connector/core/rate_limit/rate_limiter.go
```diff
@@ -0,0 +1,147 @@
...
+		cleanupInterval:      cleanupInterval,
+		logInterval:          logInterval,
+	}
+
+	// Start cleanup goroutine
+	go rl.cleanupLoop()
+	return rl
+}
+
+func (rl *TokenBucketRateLimiter) CheckLimit(ctx context.Context, w http.ResponseWriter, r *http.Request) (bool, error) {
```
edlerd: comment on this public function please.

internal/app/cluster-connector/core/rate_limit/rate_limiter.go
```diff
@@ -0,0 +1,147 @@
...
+	bucketSize           int           // maximum tokens in the bucket (requests/second)
+	clientActiveInterval time.Duration // duration after which inactive clients are considered for cleanup
+	maxClients           int
+	clients              map[string]*ClientLimiter
+	cleanupInterval      time.Duration // duration between cleanup runs
+	logInterval          time.Duration // duration between logs to avoid I/O strain
+}
+
+// NewRateLimiterStore creates a new limiter store.
+func NewTokenBucketRateLimiter(refillRate rate.Limit, bucketSize int, clientActiveInterval time.Duration, maxClients int, cleanupInterval time.Duration, logInterval time.Duration) *TokenBucketRateLimiter {
```
edlerd:
```suggestion
func NewRateLimiter(refillRate rate.Limit, bucketSize int, clientActiveInterval time.Duration, maxClients int, cleanupInterval time.Duration, logInterval time.Duration) *TokenBucketRateLimiter {
```

internal/app/cluster-connector/core/rate_limit/rate_limiter.go
```diff
@@ -0,0 +1,147 @@
...
+	refillRate           rate.Limit    // tokens added to bucket per second
+	bucketSize           int           // maximum tokens in the bucket (requests/second)
+	clientActiveInterval time.Duration // duration after which inactive clients are considered for cleanup
+	maxClients           int
+	clients              map[string]*ClientLimiter
+	cleanupInterval      time.Duration // duration between cleanup runs
+	logInterval          time.Duration // duration between logs to avoid I/O strain
+}
+
+// NewRateLimiterStore creates a new limiter store.
```
edlerd: The comment seems out of sync with the code below.

internal/app/cluster-connector/core/rate_limit/rate_limiter.go
```diff
@@ -0,0 +1,147 @@
...
+
+	if cl, exists := rl.clients[key]; exists {
+		cl.lastSeen = time.Now()
+		return cl
+	}
+
+	if len(rl.clients) >= rl.maxClients {
+		logger.Log.Info("rate limiter store is full: maximum %d clients reached", rl.maxClients)
+		return nil
+	}
```
edlerd, line 90: Should we consider the `logInterval` here as well? We'd need a global `lastLogged` as well then.

internal/app/cluster-connector/core/rate_limit/rate_limiter.go
```diff
@@ -0,0 +1,147 @@
...
+		if time.Since(cl.lastSeen) > rl.clientActiveInterval {
+			delete(rl.clients, key)
+		}
+	}
+}
+
+// extractClientIP tries to get the real client IP.
+func extractClientIP(r *http.Request) string {
+	// X-Forwarded-For when behind proxies
+	if xff := r.Header.Get("X-Forwarded-For"); xff != "" {
```
edlerd: Please unpack this into multiple lines to make it easier to read.

edlerd (commented):
Good progress and nearly perfect. A handful of observations below.

internal/app/cluster-connector/core/rate_limit/rate_limiter.go
omarelkashef: Yes, that is a potential attack. Thank you for catching it!

edlerd (dismissed):
Code LGTM, just one issue with the formatter left.

edlerd (approved):

PR #394, opened 2026-01-27
feat(tests): e2e tests for clusters and cluster detail pages WD-32796

omarelkashef:
Add e2e tests for clusters and cluster detail pages.


ui/tests/cluster-detail.spec.ts
```diff
@@ -0,0 +1,190 @@
...
+    await setClusterConfig(page, cluster, "Disk threshold", "1");
+    const container = page.locator(".warning-list");
+    await expect(container).toBeVisible();
+    const notifications = page.locator(".p-notification--caution");
+    const initialCount = await notifications.count();
+
+    await setClusterConfig(page, cluster, "Memory threshold", "100");
+    await setClusterConfig(page, cluster, "Disk threshold", "100");
+    const finalCount = await notifications.count();
+    expect(initialCount === 0 || finalCount < initialCount).toBe(true);
```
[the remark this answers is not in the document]
omarelkashef: The intended logic is to make sure that if there were any warning, they should be gone.

ui/tests/cluster-detail.spec.ts
```diff
@@ -0,0 +1,190 @@
...
+      page.getByText("Storage pool details", { exact: true }),
+    ).not.toBeVisible();
+  });
+});
+
+test.describe("Cluster Members", () => {
+  test("should display cluster members graph", async ({ page }) => {
+    const cluster = randomExistingClusterName();
+    await goToClusterDetail(page, cluster);
+    const container = page.locator(".cluster-detail-members-graph");
```
kimanhou: Are the classnames you added used only in tests ?
I would prefer to keep classnames for styling only. Can you find another way to get the container, like `page.getByRole('button', { name: /save/i })` or `page.getByText('Cluster initialized')`. If not, you can use `data-testid`.

ui/src/pages/clusters/actions/ClusterEditDescriptionBtn.tsx
```diff
@@ -19,6 +19,7 @@ const ClusterEditDescriptionBtn: FC<Props> = ({
   return (
     <Button
       appearance={appearance}
+      name="cluster-edit-description-btn"
```
edlerd:
```suggestion
```
Seems unused.

ui/tests/cluster-detail.spec.ts
```diff
@@ -0,0 +1,194 @@
+import { test } from "@playwright/test";
+import {
+  ensureClusterExists,
+  goToClusterDetail,
+  setClusterConfig,
+} from "./helpers/remote-clusters";
+import { expect } from "./fixtures/test-extension";
+
+const randomExistingClusterName = (): string => {
```
edlerd: Let's move this to the helper.

ui/tests/cluster-detail.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+  });
+
+  test("should display warnings for low thresholds and hide them when thresholds are high", async ({
+    page,
+  }) => {
+    const cluster = randomExistingClusterName();
+    await ensureClusterExists(page, cluster);
+    await setClusterConfig(page, cluster, "Memory threshold", "1");
+    await setClusterConfig(page, cluster, "Disk threshold", "1");
+    const container = page.locator(".warning-list");
```
edlerd: Let's better check for the text of the warning to be on the screen. That way the test is not depending on implementation details, but directly on what the user sees.

ui/tests/cluster-detail.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+    await expect(container).toBeVisible();
+    const notifications = page.locator(".p-notification--caution");
+    const initialCount = await notifications.count();
+
+    await setClusterConfig(page, cluster, "Memory threshold", "100");
+    await setClusterConfig(page, cluster, "Disk threshold", "100");
+    const finalCount = await notifications.count();
+    expect(initialCount === 0 || finalCount < initialCount).toBe(true);
```
edlerd, line 38:
```suggestion
    expect(finalCount === 0).toBe(true);
```
Maybe a bit simpler?

ui/tests/cluster-detail.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+
+  test("should show all storage pool columns", async ({ page }) => {
+    const cluster = randomExistingClusterName();
+    await goToClusterDetail(page, cluster);
+    await page.getByText("View details", { exact: true }).click();
+    await expect(page.locator("th", { hasText: "Storage pool" })).toBeVisible();
+    await expect(page.locator("th", { hasText: "Location" })).toBeVisible();
+    await expect(page.locator("th", { hasText: "Percent used" })).toBeVisible();
+    await expect(page.locator("th", { hasText: "Total used" })).toBeVisible();
+    await expect(page.locator("th", { hasText: "Total size" })).toBeVisible();
```
edlerd: I think we can check for the modal heading and don't have to evaluate every single column heading here.

edlerd (commented):
Looks already very good, some smallish comments below.

ui/tests/cluster-detail.spec.ts
omarelkashef: Some warning are not related to thresholds so we are not guaranteed to have no warnings after setting threshold to 100. Therefore, I check that the final warnings count is less than initial or the initial count is zero.

ui/tests/cluster-detail.spec.ts
```diff
@@ -0,0 +1,186 @@
...
+      page.getByText("Storage pool details", { exact: true }),
+    ).toBeVisible();
+  });
+
+  test("should show all storage pool columns", async ({ page }) => {
+    const cluster = randomSeedClusterName();
+    await goToClusterDetail(page, cluster);
+    await page.getByText("View details", { exact: true }).click();
+    await expect(page.locator(".p-modal")).toBeVisible();
+  });
```
edlerd:
```suggestion
```
This seems not necessary anymore considering the test above. Wdyt?

ui/tests/clusters.spec.ts
```diff
@@ -0,0 +1,43 @@
+import { test } from "@playwright/test";
+import { expect } from "./fixtures/test-extension";
+
+test.describe("Cluster Configuration", () => {
```
edlerd:
```suggestion
test.describe("Cluster Bulk Configuration", () => {
```
A bit more detail here.

ui/tests/clusters.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+        `All ${rowCount} cluster${rowCount !== 1 ? "s" : ""} selected.`,
+      )
+      .click();
+    await page.getByRole("button", { name: "Configure clusters" }).click();
+
+    const diskRow = page
+      .getByText("Disk threshold")
+      .locator("..")
+      .locator("..")
+      .locator("..");
```
edlerd: What is this? It looks suspicious. Can we find a better expression of this?

ui/tests/clusters.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+    diskRow.getByRole("button").click();
+    await diskRow.locator("#diskThreshold").fill("100");
+
+    const memoryRow = page
+      .getByText("Memory threshold")
+      .locator("..")
+      .locator("..")
+      .locator("..");
```
edlerd: Same as above.

edlerd (dismissed):
Some tiny nitpicks, then it should be good to merge from my side.

ui/tests/clusters.spec.ts
omarelkashef: No really because the rows for memory and disk threshold are identical except for the label. So we have to locate the label and walk up. Another option is to add a class to the row or a name to the buttons to differentiate.

edlerd: There should be an edit button that we can target here directly. If the button has no unique description, we should probably adjust it.

For the input, we should have a unique label as well. Otherwise, accessibility is broken. Maybe refine the implementation so we can target it here more easily.

ui/tests/clusters.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+    await page.getByLabel("Select all").click();
+    await page
+      .getByText(
+        `All ${rowCount} cluster${rowCount !== 1 ? "s" : ""} selected.`,
+      )
+      .click();
+
+    await page.getByRole("button", { name: "Configure clusters" }).click();
+    await page.getByTitle("Set Disk threshold").click();
+    await page.locator("#diskThreshold").fill("100");
```
edlerd: Can we use `getByLabel` here instead of an id based locator? This ensures two goals. 1. The code gets a bit simpler, and the test is closer to the actual user experience, and 2. We cover accessibility with tests. We might need to adjust the implementation to ensure this has the right label assigned.

ui/tests/clusters.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+    await page.getByTitle("Set Memory threshold").click();
+    await page.locator("#memoryThreshold").fill("100");
```
edlerd: Here as well, better use the label to target the input.

edlerd (commented):
Thanks, nearly perfect, just one last bit.

edlerd (approved):
LGTM, thanks for increasing the test coverage!

PR #395, opened 2026-01-28
fix validation for cluster configuration, allow unsetting cluster config

### LINKED ISSUE #392: Validate cluster threshold values
Cluster thresholds can currently be set to invalid values >100% and <0% without error messages.

Image

### LINKED ISSUE #393: Allow deleting cluster description
There is no way to remove a cluster description. Once set, the description can only be changed but not deleted.

edlerd:
fixes #393 
fixes #392

image

### DIFF
--- a/internal/app/management-api/api/v1/remote_cluster.go
+++ b/internal/app/management-api/api/v1/remote_cluster.go
@@ -3,6 +3,7 @@ package v1
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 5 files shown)


internal/app/management-api/api/v1/remote_cluster.go
```diff
@@ -149,28 +149,22 @@ func remoteClusterPatch(rc types.RouteConfig) types.EndpointHandler {
...
-			if payload.MemoryThreshold > 0 {
-				err = store.UpdateRemoteClusterConfig(ctx, tx, store.RemoteClusterConfig{
-					RemoteClusterID: existingRemoteCluster.ID,
-					Key:             store.MemoryThresholdKey,
-					Value:           fmt.Sprintf("%d", payload.MemoryThreshold),
-				})
-				if err != nil {
-					return err
-				}
+			err = store.UpdateRemoteClusterConfig(ctx, tx, store.RemoteClusterConfig{
```
omarelkashef, line 169: Why did we leave the validation for the UI?  We can have `if payload.MemoryThreshold >= 0 && payload.MemoryThreshold <= 100` and similarly for DiskThreshold.

ui/src/pages/clusters/BulkConfigureClusterPanel.tsx
```diff
@@ -116,7 +123,7 @@ const BulkConfigureClusterPanel: FC = () => {
                 }}
                 firstValue={clusters[0]?.disk_threshold}
                 defaultValue={80}
-                label="Disk threshold"
+                label="Disk threshold percentage"
```
omarelkashef, line 126: This will break tests from #394. We should make changes according whichever gets merged first.

ui/src/pages/clusters/BulkConfigureClusterPanel.tsx
```diff
@@ -130,6 +137,11 @@ const BulkConfigureClusterPanel: FC = () => {
                   onBlur={formik.handleBlur}
                   onChange={formik.handleChange}
                   value={formik.values.diskThreshold}
+                  error={
+                    formik.touched.diskThreshold
```
omarelkashef: `formik.touched.diskThreshold && formik.errors.diskThreshold` is easier?

omarelkashef (commented):
Thank you for this PR! A few observations below. I could not do QA because demo server is down.

internal/app/management-api/api/v1/remote_cluster.go
edlerd: Right we should keep validation within the suggested limits. Good catch!

ui/src/pages/clusters/ConfigureClusterPanel.tsx
```diff
@@ -137,6 +152,9 @@ const ConfigureClusterPanel: FC = () => {
                 onBlur={formik.handleBlur}
                 onChange={formik.handleChange}
                 value={formik.values.description}
+                error={
+                  formik.touched.description ? formik.errors.description : null
```
omarelkashef: To keep things consistent, `formik.touched.description &&  formik.errors.description`

internal/app/management-api/api/v1/remote_cluster.go
```diff
@@ -149,10 +149,8 @@ func remoteClusterPatch(rc types.RouteConfig) types.EndpointHandler {
 			if payload.Status != "" {
 				newRemoteCluster.Status = string(payload.Status)
 			}
-			if payload.Description != "" {
-				newRemoteCluster.Description = payload.Description
-			}
-			if payload.DiskThreshold > 0 {
+			newRemoteCluster.Description = payload.Description
+			if payload.DiskThreshold >= 0 || payload.DiskThreshold <= 100 {
```
omarelkashef: Is it intentional to skip and not return err if value is out of range but return err if `UpdateRemoteClusterConfig` returns it?

omarelkashef (commented):
Looks good to me, just a questions and nitpick.

internal/app/management-api/api/v1/remote_cluster.go
edlerd: Yes, changing this to do validation first and throw an error if it fails seems like a better approach here.

omarelkashef: I mean now, we do not return error when validation fails. If user submits an invalid value, they do not get an appropriate error message.

omarelkashef: Also what about simplifying this block, and updating `newRemoteCluster` with the new values with returning appropriate validation errors and calling `UpdateRemoteClusterConfig` at the end?

edlerd: The thresholds are custom configurations in another table. Hence, we update them each at a time and can't update them together.  I changed it to validate each before writing and saving the cluster object last.

omarelkashef (dismissed):
Looks great! Thank you

internal/app/management-api/api/v1/remote_cluster.go
omarelkashef: You are right. I confused `UpdateRemoteCluster` with `UpdateRemoteClusterConfig`

omarelkashef (approved):
Thank you for the new changes!

PR #402, opened 2026-02-06
chore: split install-deps make target into two

omarelkashef:
Split install-deps make target into install-core and install-deps.


edlerd (approved):
LGTM

PR #407, opened 2026-02-10
surface rate limit configuration in the ui settings page

edlerd:
image

### DIFF
--- a/internal/app/management-api/api/v1/configuration.go
+++ b/internal/app/management-api/api/v1/configuration.go
@@ -120,5 +120,35 @@ func mapEnvToConfig(cfg config.Config) models.Configuration {
 			Title:       "Prometheus base URL",
 			Description: "The base url for prometheus.",
 		},
+		RateLimitBucketSize: models.ConfigData{
+			Value:       strconv.Itoa(cfg.RateLimitBucketSize),
+			Title:       "Rate Limit Bucket Size",
+			Description: "Maximum tokens in the rate limiting bucket (requests/second).",
+		},
+		RateLimitClientActiveInterval: models.ConfigData{
+			Value:       cfg.RateLimitClientActiveInterval.String(),
+			Title:       "Rate Limit Client Active Interval",
+			Description: "Time after which inactive clients are available for clean up.",
+		},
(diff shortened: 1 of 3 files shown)


internal/app/management-api/api/v1/configuration.go
```diff
@@ -120,5 +120,35 @@ func mapEnvToConfig(cfg config.Config) models.Configuration {
 			Title:       "Prometheus base URL",
 			Description: "The base url for prometheus.",
 		},
+		RateLimitBucketSize: models.ConfigData{
```
omarelkashef, line 123: The titles of all the configs are sentence case, but the rate limiting config are pascal case

internal/app/management-api/api/v1/configuration.go
```diff
@@ -120,5 +120,35 @@ func mapEnvToConfig(cfg config.Config) models.Configuration {
...
+			Title:       "RateLimitLogInterval",
+			Description: "Time between logs to avoid I/O strain.",
+		},
+		RateLimitMaxClients: models.ConfigData{
+			Value:       strconv.Itoa(cfg.RateLimitMaxClients),
+			Title:       "RateLimitMaxClients",
+			Description: "Maximum number of clients to track in the rate limiter. If reached, new clients will be rejected.",
+		},
+		RateLimitRefillRate: models.ConfigData{
+			Value:       fmt.Sprintf("%.2f", cfg.RateLimitRefillRate),
```
omarelkashef, line 149: Do we need the decimal point here?

omarelkashef (commented):
Looks good to me, just two tiny comments.

internal/app/management-api/api/v1/configuration.go
edlerd: Good point, this is the visual title and it can be a little nicer to read with spaces. Adjust it.

internal/app/management-api/api/v1/configuration.go
edlerd: I think a float is a valid value, so I would not want to hide if this is configured with decimals.

omarelkashef (approved):
Looks nice. Thank you!

PR #411, opened 2026-02-12
feat: add ceph status distribution and warnings if any ceph member is not online

edlerd:
# Done
* Receive and store ceph status in heartbeats from the MicroClouds
* Show a warning in cluster manager if any of the ceph members of a cluster are not online

depending on https://github.com/canonical/microcloud/pull/1222

# Screenshots

image

image

### DIFF
--- a/internal/app/management-api/api/v1/remote_cluster.go
+++ b/internal/app/management-api/api/v1/remote_cluster.go
@@ -192,8 +192,14 @@ func toRemoteClustersAPI(dbEntries []store.RemoteClusterWithDetail) ([]models.Re
 	// generate lookup for remoteCluster details
 	var remoteClusters []models.RemoteCluster
 	for _, e := range dbEntries {
+		var cephStatuses []models.StatusDistribution
+		err := json.Unmarshal(e.CephStatuses, &cephStatuses)
+		if err != nil {
+			return nil, fmt.Errorf("failed to unmarshal ceph statuses: %w", err)
+		}
+
 		var memberStatuses []models.StatusDistribution
-		err := json.Unmarshal(e.MemberStatuses, &memberStatuses)
+		err = json.Unmarshal(e.MemberStatuses, &memberStatuses)
 		if err != nil {
 			return nil, fmt.Errorf("failed to unmarshal member statuses: %w", err)
 		}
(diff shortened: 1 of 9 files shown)


internal/pkg/api/models/v1/remote_cluster.go
```diff
@@ -73,6 +75,7 @@ type RemoteClusterPost struct {
 
 // RemoteClusterStatusPost is sent by LXD with to inform about its current status.
```
omarelkashef: Unrelated to this PR, but this comment has an extra "with"

internal/app/management-api/api/v1/remote_cluster.go
```diff
@@ -192,8 +192,14 @@ func toRemoteClustersAPI(dbEntries []store.RemoteClusterWithDetail) ([]models.Re
 	// generate lookup for remoteCluster details
 	var remoteClusters []models.RemoteCluster
 	for _, e := range dbEntries {
+		var cephStatuses []models.StatusDistribution
+		err := json.Unmarshal(e.CephStatuses, &cephStatuses)
+		if err != nil {
+			return nil, fmt.Errorf("failed to unmarshal instance statuses: %w", err)
```
omarelkashef: Should the error be "failed to unmarshal ceph statuses"?

internal/app/management-api/api/v1/remote_cluster.go
```diff
@@ -192,8 +192,14 @@ func toRemoteClustersAPI(dbEntries []store.RemoteClusterWithDetail) ([]models.Re
 	// generate lookup for remoteCluster details
 	var remoteClusters []models.RemoteCluster
 	for _, e := range dbEntries {
```
omarelkashef, line 194: Is it possible to report the entries that  we could not unmarshal and the UI can show a warning instead of returning an error for the whole request? This might be out of scope for this PR as it will involve some UI work too.

ui/src/util/clusterWarnings.tsx
```diff
@@ -48,5 +52,16 @@ export const getClusterWarnings = (cluster: Cluster): string[] => {
     );
   }
 
+  cluster.ceph_statuses.forEach((cephDistribution) => {
+    const status = capitalizeFirstLetter(cephDistribution.status.toLowerCase());
+    const count = cephDistribution.count;
+    if (status === "Online" && count > 0) {
```
omarelkashef: This can push  "Ceph has 0 members with status Online" to `results`. It that intentional?

omarelkashef (commented):
Looks good! I left a couple of comments below.

internal/app/management-api/api/v1/remote_cluster.go
edlerd: yes!

ui/src/util/clusterWarnings.tsx
edlerd: good catch!

internal/app/management-api/api/v1/remote_cluster.go
edlerd: As we fully control the database in terms of what goes in, I think it is ok to surface the error here. If this throws, then we have broken entries in postgres.

omarelkashef (approved):
LGTM. Thank you!

PR #418, opened 2026-02-18
chore: refactor run-backend.sh and run-coverage.sh

omarelkashef:
refactor run-backend.sh and run-coverage.sh and extracted setup operations into setup-env.sh

### DIFF
--- a/test/run-backend.sh
+++ b/test/run-backend.sh
@@ -3,68 +3,11 @@
 set -e
 set -x
 
-# setup /etc/hosts entries for local domains
-make add-hosts
-
-# generate keys if they don't exist
 SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
-KEY_DIR="$SCRIPT_DIR/keys"
-if [ ! -d "$KEY_DIR" ]; then
-  mkdir -p $KEY_DIR
-  openssl req -nodes -x509 -newkey ec \
-    -pkeyopt ec_paramgen_curve:secp384r1 \
-    -sha384 \
-    -keyout $KEY_DIR/tls.key \
-    -out $KEY_DIR/tls.crt \
-    -subj "/C=GB/ST=London/L=London/O=Microcloud Cluster Manager/OU=dev/CN=ma.lxd-cm.local" \
-    -days 3000 \
-    -addext "subjectAltName=DNS:ma.lxd-cm.local,DNS:cc.lxd-cm.local"
-
-  cp $KEY_DIR/tls.crt $KEY_DIR/ca.crt
-fi
-export management_api_cert_secret="$KEY_DIR"
-export cluster_connector_cert_secret="$KEY_DIR"
-
-# run postgres
-docker rm -f my-postgres 2>/dev/null || true
-docker run -d \
-  --name my-postgres \
-  -e POSTGRES_USER=admin \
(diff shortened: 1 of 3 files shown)


ui/tests/cluster-detail.spec.ts
```diff
@@ -152,7 +152,10 @@ test.describe("Navigation and UI", () => {
   test("should navigate back to clusters list", async ({ page }) => {
     const cluster = randomSeedClusterName();
     await goToClusterDetail(page, cluster);
-    await page.getByRole("link", { name: "Clusters", exact: true }).click();
+    await page
+      .getByLabel("Breadcrumbs")
+      .getByRole("link", { name: "Clusters" })
+      .click();
```
edlerd: Seems like an unrelated change. Can it be reverted?

test/setup-env.sh
```diff
@@ -0,0 +1,65 @@
...
+export MANAGEMENT_API_TLS_PATH="$KEY_DIR"
+export CLUSTER_CONNECTOR_TLS_PATH="$KEY_DIR"
+export CLUSTER_CONNECTOR_PORT=9000
+export PROMETHEUS_BASE_URL=http://$PROMETHEUS_ADDRESS/api/v1/write
+export RATE_LIMIT_REFILL_RATE=1000
+export RATE_LIMIT_BUCKET_SIZE=1000
+export RATE_LIMIT_CLIENT_ACTIVE_INTERVAL_SECONDS=60
+export RATE_LIMIT_CLEANUP_INTERVAL_SECONDS=60
+export RATE_LIMIT_LOG_INTERVAL_SECONDS=60
+export RATE_LIMIT_MAX_CLIENTS=1000000
```
edlerd:
```suggestion
export RATE_LIMIT_MAX_CLIENTS=1000000

```

test/run-coverage.sh
```diff
@@ -1,69 +1,9 @@
-#!/bin/bash
-
-set -e
-set -x
-
```
edlerd, line 5: We should keep this in the head of both files.

test/run-backend.sh
edlerd: Can we put this into a function to stop Postgres and another one to stop Prometheus? Then we can keep both function in the setup-env file right next to where we start the containers. To keep the definition of `my-postgres` within one file.

We should also stop Prometheus here.

internal/app/management-api/api/v1/static/placeholder.html
edlerd, line 1: We should not remove this file.

test/run-coverage.sh
edlerd: Let's use the extracted functions to stop postgres and prometheus here as well.

edlerd (commented):
good start, some observations below.

ui/tests/cluster-detail.spec.ts
omarelkashef: This test was reported to be failing on the PR, so I used this PR to fix it as well.

edlerd: That shouldn't be necessary. I see all the previous runs [succeeded in main branch](https://github.com/canonical/microcloud-cluster-manager/actions/workflows/tics.yaml), including very recent runs from yesterday evening. Can you revert this?

edlerd (approved):
LGTM, thanks for improving those scripts by extracting the common parts :)

PR #448, opened 2026-03-19
test: validate api against openapi schema WD-34916

omarelkashef:
Validate api against openapi schema as part of CI.


.github/workflows/pr.yaml
```diff
@@ -197,6 +197,31 @@ jobs:
...
+          test/run-backend.sh &
+          curl --head --fail --retry-delay 2 --retry 100 --retry-connrefused --insecure https://localhost:30000
+      - name: Install Go
+        uses: actions/setup-go@4b73464bb391d4059bd26b0524d20df3927bd417 # v6.3.0
+        with:
+          go-version: 1.25.x
+
+      - name: Fetch OpenAPI description file from microcloud repo
+        run: |
+          curl -L https://raw.githubusercontent.com/canonical/microcloud/main/doc/_extra/cluster-manager-api.yaml -o test/openapi/cluster-manager-api.yaml
```
edlerd: Can we move this part to the make target, so we can easily run the test locally?

omarelkashef: Good idea!

test/helpers/openapi.go
```diff
@@ -0,0 +1,223 @@
...
+	"github.com/getkin/kin-openapi/openapi3filter"
+	"github.com/getkin/kin-openapi/routers"
+)
+
+// TestOpenAPIOperation validates a request with a JSON body and its response against the OpenAPI spec.
+func TestOpenAPIOperation(t *testing.T, env *Environment, router routers.Router, method string, path string, body any) error {
+	reqInput, res, err := validateAndSendRequest(env, router, method, path, body)
+	if err != nil {
+		return err
+	}
```
edlerd, line 23: We shouldn't have `and` in a function name. Or at least it hints at a code smell. I'd suggest to inline the validate and send-request calls in this function directly. Makes the hierarchy more equal within the three functions and each is clearly defined in scope.

test/helpers/openapi.go
```diff
@@ -0,0 +1,223 @@
...
+	"testing"
+	"time"
+
+	"github.com/canonical/lxd/shared/api"
+	"github.com/getkin/kin-openapi/openapi3filter"
+	"github.com/getkin/kin-openapi/routers"
+)
+
+// TestOpenAPIOperation validates a request with a JSON body and its response against the OpenAPI spec.
+func TestOpenAPIOperation(t *testing.T, env *Environment, router routers.Router, method string, path string, body any) error {
```
edlerd, line 19:
```suggestion
func EnforceSuccessSchema(t *testing.T, env *Environment, router routers.Router, method string, path string, body any) error {
```
Maybe a better name?

edlerd, line 19:
```suggestion
func TestOpenAPIOperation(env *Environment, router routers.Router, method string, path string, body any) error {
```
t is unused?

test/helpers/openapi.go
```diff
@@ -0,0 +1,223 @@
...
+	reqInput, res, err := validateAndSendRequest(env, router, method, path, body)
+	if err != nil {
+		return err
+	}
+	return validateResponse(reqInput, res)
+}
+
+// TestOpenAPIOperationInvalid sends an intentionally invalid request (skipping request validation)
+// and validates only that the error response conforms to the OpenAPI spec.
+func TestOpenAPIOperationInvalid(t *testing.T, env *Environment, router routers.Router, method string, path string, body any) error {
```
edlerd, line 29:
```suggestion
func TestOpenAPIOperationInvalid(env *Environment, router routers.Router, method string, path string, body any) error {
```
unused?

test/helpers/openapi.go
```diff
@@ -0,0 +1,223 @@
...
+	reqInput := &openapi3filter.RequestValidationInput{
+		Request:    req,
+		PathParams: pathParams,
+		Route:      route,
+		Options: &openapi3filter.Options{
+			AuthenticationFunc:                openapi3filter.NoopAuthenticationFunc,
+			MultiError:                        true,
+			RejectWhenRequestBodyNotSpecified: false,
+		},
+	}
```
edlerd, line 50: Let's reuse `getRequestValidationInput` here.

test/helpers/openapi.go
edlerd, line 29:
```suggestion
func EnforceErrorSchema(t *testing.T, env *Environment, router routers.Router, method string, path string, body any) error {
```

edlerd (commented):
Some thoughts on the structure below, also CI is not passing yet.

test/helpers/openapi.go
```diff
@@ -0,0 +1,223 @@
...
+
+	res, err := sendRequest(env, req)
```
edlerd, line 52: We have to ensure `res.Body.Close()` is called after this, best with a `defer.

test/helpers/openapi.go
```diff
@@ -0,0 +1,223 @@
...
+	"time"
+
+	"github.com/canonical/lxd/shared/api"
+	"github.com/getkin/kin-openapi/openapi3filter"
+	"github.com/getkin/kin-openapi/routers"
+)
+
+// TestOpenAPIOperation validates a request with a JSON body and its response against the OpenAPI spec.
+func TestOpenAPIOperation(t *testing.T, env *Environment, router routers.Router, method string, path string, body any) error {
+	reqInput, res, err := validateAndSendRequest(env, router, method, path, body)
```
edlerd, line 20: Here as well, ensure `res.Body.Close()` is called.

omarelkashef:
Please refer to #455 instead.

PR #455, opened 2026-03-24
test: validate api against openapi schema WD-34916

omarelkashef:
Validate api against openapi schema as part of CI.


test/helpers/openapi.go
```diff
@@ -0,0 +1,203 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	tlsClient, err := NewTLSHTTPClient(api.URL{}, nil, certPublicKey, env.ManagementAPIHost())
+	if err != nil {
+		return nil, err
+	}
+
+	if err := headers(req); err != nil {
```
edlerd:
```suggestion
  err = headers(req)
  if err != nil {
```
Best not use the shorthand style.

test/openapi/openapi_remote_cluster.go
```diff
@@ -0,0 +1,128 @@
...
+		condition := "500: patching a cluster with internal server error returns error response."
+		malformedClusterName := "%g1" // A cluster name that is improperly escaped to trigger internal error in handler
+		err := helpers.EnforceErrorSchema(env, router, http.MethodPatch, fmt.Sprintf("/1.0/remote-cluster/%s", malformedClusterName), nil)
+		helpers.LogTestOutcome(t, condition, err)
+	}
+}
+
+func testDeleteRemoteClusterSuccess(env *helpers.Environment, router routers.Router) (testName string, testFunc func(t *testing.T)) {
+	return "testDeleteRemoteClusterSuccess DELETE /1.0/remote-cluster/{remoteClusterName} 200", func(t *testing.T) {
+		clusterName := "cluster-2" // From seed data
```
edlerd: It would be nice to make the test suite idempotent so we can rerun it without hassle. I am not sure if that is feasible, though.

For any create call, we could use randomized names. For the delete here, we would need to do the setup on the fly with a randomized name. So create a cluster first, and then the test starts to delete it. That way we would get completely independent of the seed data.

edlerd (commented):
Looks good already. Thanks for the iteration. Some linter issues are left.

edlerd (approved):
LGTM, thanks.

PR #456, opened 2026-03-25
Reverse tunnel for microclouds

edlerd:
# Done
- added endpoint to open a tunnel from each cluster
- proxy endpoint to serve the LXD-UI

Draft because there are open issues

Follow up of #207

### DIFF
--- a/internal/app/cluster-connector/api/v1/remote_cluster.go
+++ b/internal/app/cluster-connector/api/v1/remote_cluster.go
@@ -4,9 +4,13 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
+	"net/url"
+	"os"
 	"strings"
 	"time"
 
@@ -16,10 +20,14 @@ import (
 	"github.com/canonical/microcloud-cluster-manager/internal/app/cluster-connector/core/certificate"
 	"github.com/canonical/microcloud-cluster-manager/internal/app/cluster-connector/core/rate_limit"
 	"github.com/canonical/microcloud-cluster-manager/internal/pkg/api/models/v1"
+	"github.com/canonical/microcloud-cluster-manager/internal/pkg/config"
 	"github.com/canonical/microcloud-cluster-manager/internal/pkg/database/store"
 	"github.com/canonical/microcloud-cluster-manager/internal/pkg/logger"
 	"github.com/canonical/microcloud-cluster-manager/internal/pkg/types"
 	"github.com/golang/snappy"
+	"github.com/google/uuid"
+	"github.com/gorilla/mux"
+	"github.com/gorilla/websocket"
 	"github.com/jmoiron/sqlx"
 	"github.com/prometheus/common/expfmt"
 	"github.com/prometheus/common/model"
@@ -54,8 +62,39 @@ var RemoteClusterProtected = types.RouteGroup{
 			Handler: remoteClusterStatusPost,
 		},
 		{
+			Path:    "ws",
+			Method:  http.MethodGet,
+			Handler: remoteClusterWsGet,
+		},
+		{
 			Method:  http.MethodDelete,
 			Handler: remoteClusterDelete,
+		},
+	},
+}
+
+// RemoteClusterInternal are the API endpoints that cluster manager calls itself.
+var RemoteClusterInternal = types.RouteGroup{
+	Prefix: "remote-cluster",
+	Middlewares: []types.RouteMiddleware{
+		rate_limit.RateLimitMiddleware,
+		auth.InternalAuthMiddleware,
+	},
+	Endpoints: []types.Endpoint{
+		{
+			Path:    "{remoteClusterName}/tunnel/{path:.*}",
+			Method:  http.MethodGet,
+			Handler: remoteClusterTunnel,
+		},
+		{
+			Path:    "{remoteClusterName}/tunnel/{path:.*}",
+			Method:  http.MethodPost,
+			Handler: remoteClusterTunnel,
+		},
+		{
+			Path:    "{remoteClusterName}/tunnel/{path:.*}",
+			Method:  http.MethodPut,
+			Handler: remoteClusterTunnel,
 		},
 	},
 }
@@ -182,7 +221,7 @@ func remoteClustersPost(rc types.RouteConfig) types.EndpointHandler {
 	}
 }
 
-// apply mtls for this endpoint.
+// mtls applied by the auth middleware for this endpoint.
 func remoteClusterStatusPost(rc types.RouteConfig) types.EndpointHandler {
 	return func(w http.ResponseWriter, r *http.Request) error {
 		payload := models.RemoteClusterStatusPost{}
@@ -201,10 +240,6 @@ func remoteClusterStatusPost(rc types.RouteConfig) types.EndpointHandler {
 			dbRemoteCluster, err := store.GetRemoteClusterWithDetailByID(ctx, tx, remoteClusterID)
 			if err != nil {
 				return err
-			}
-
-			if dbRemoteCluster == nil {
-				return fmt.Errorf("remote cluster not found")
 			}
 
 			dbRemoteClusterDetail, err := store.GetRemoteClusterDetail(ctx, tx, remoteClusterID)
@@ -356,7 +391,12 @@ func forwardMetricsToPrometheus(timeSeries []prompb.TimeSeries, rc types.RouteCo
 	if err != nil {
 		return fmt.Errorf("failed to send metrics to Prometheus: %w", err)
 	}
-	defer resp.Body.Close()
+	defer func() {
+		err := resp.Body.Close()
+		if err != nil {
+			logger.Log.Warnw("Failed to close Prometheus response body", "error", err)
+		}
+	}()
 
 	if resp.StatusCode > 299 {
 		body, err := io.ReadAll(resp.Body)
@@ -369,6 +409,7 @@ func forwardMetricsToPrometheus(timeSeries []prompb.TimeSeries, rc types.RouteCo
 	return nil
 }
 
+// mtls applied by the auth middleware for this endpoint.
 func remoteClusterDelete(rc types.RouteConfig) types.EndpointHandler {
 	return func(w http.ResponseWriter, r *http.Request) error {
 		remoteClusterID, err := request.GetContextValue[int](r.Context(), auth.CtxRemoteClusterID)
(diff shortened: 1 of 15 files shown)


internal/pkg/database/store/remote_cluster_detail.go
```diff
@@ -17,24 +17,25 @@ import (
...
+	CPULoad15         string          `db:"cpu_load_15"`         // CPU load (15 minute average)
+	MemoryTotalAmount int64           `db:"memory_total_amount"` // Total memory in bytes
+	MemoryUsage       int64           `db:"memory_usage"`        // Memory usage in bytes
+	InstanceCount     int64           `db:"instance_count"`      // Number of instances
+	InstanceStatuses  json.RawMessage `db:"instance_statuses"`   // JSON array of instance statuses
+	MemberCount       int64           `db:"member_count"`        // Number of members
+	MemberStatuses    json.RawMessage `db:"member_statuses"`     // JSON array of member statuses
+	StoragePoolUsages json.RawMessage `db:"storage_pool_usages"` // JSON array of storage pool usages
+	UIURL             string          `db:"ui_url"`              // UI URL
+	TunnelMemberURL   string          `db:"tunnel_member_url"`   // Cluster manager member url that holds the tunnel
```
omarelkashef: The name made me think the variable refers to a member in the cluster but the comment implies a member in cluster manager.

internal/pkg/database/store/remote_cluster_detail.go
```diff
@@ -269,6 +275,41 @@ func UpdateRemoteClusterDetail(ctx context.Context, tx *sqlx.Tx, remoteClusterID
...
+	if err != nil {
+		return fmt.Errorf("update remote_cluster_details entry failed: %w", err)
+	}
+
+	n, err := result.RowsAffected()
+	if err != nil {
+		return fmt.Errorf("fetch affected rows: %w", err)
+	}
+
+	if n != 1 {
```
omarelkashef, line 290: Are these changes rolled back? If not, is not it weird if we return an error and persist the db changes at the same time?

internal/pkg/middleware/request.go
```diff
@@ -27,15 +31,20 @@ func RequestTrace(next http.Handler) http.Handler {
...
-		ctx = context.WithValue(ctx, request.RequestKey(), &v)
+		ctx := context.WithValue(r.Context(), request.RequestKey(), &v)
 		next.ServeHTTP(w, r.WithContext(ctx))
 	})
 }
 
 // LogRequest is a middleware that logs a request/response cycle.
 func LogRequest(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// Skip logging wrapper for WebSocket upgrade requests to preserve Hijacker
```
omarelkashef, line 39: Why skip the wrapper if it implements Hijacker interface below?

internal/app/management-api/api/v1/remote_cluster.go
```diff
@@ -241,6 +261,7 @@ func toRemoteClustersAPI(dbEntries []store.RemoteClusterWithDetail) ([]models.Re
 			InstanceStatuses:   instanceStatuses,
 			StoragePoolUsages:  storagePoolUsages,
 			UIURL:              e.UIURL,
+			TunnelConnected:    e.TunnelMemberURL != "",
```
omarelkashef: Should this be a more thorough check to make sure the member is reachable?

internal/app/management-api/api/v1/remote_cluster.go
```diff
@@ -253,3 +274,113 @@ func toRemoteClustersAPI(dbEntries []store.RemoteClusterWithDetail) ([]models.Re
...
+		if _, err = io.Copy(w, resp.Body); err != nil {
+			logger.Log.Errorf("failed to stream response body", "err", err)
+			return nil
+		}
+
+		needsTokenRefresh := resp.StatusCode == 401 || resp.StatusCode == 403
+		if needsTokenRefresh {
+			refreshCtx, cancelRefresh := context.WithTimeout(context.Background(), 30*time.Second)
+			defer cancelRefresh()
+			_, err = sessionProvider.RefreshSessionTokens(refreshCtx)
```
omarelkashef: Do we not retry the request after refreshing the tokens? If not, what is the most appropriate error to return?

internal/app/cluster-connector/api/v1/remote_cluster.go
```diff
@@ -392,3 +432,350 @@ func remoteClusterDelete(rc types.RouteConfig) types.EndpointHandler {
 		return response.EmptySyncResponse.Render(w, r)
 	}
 }
+
+// apply mtls for this endpoint.
```
omarelkashef: Is not mtls already applied through the auth middleware?

omarelkashef (commented):
I did an initial review. Very good PR! A few comments below

internal/pkg/database/store/remote_cluster_detail.go
edlerd: Any more specific terminology I can think of become very bulky, hence I went with this one. Do you have a suggestion?

internal/pkg/database/store/remote_cluster_detail.go
edlerd: All database changes are done in transactions. If we throw here, the transaction is not comitted.

internal/pkg/middleware/request.go
edlerd: Forgot to clean up the approach below. It was for the other direction, trying to enable websockets from browser back to LXD. But we discarded that idea. I removed the Hijacker interface below. We have to keep the skip here as the tunnel doesn't work without.

internal/app/management-api/api/v1/remote_cluster.go
edlerd: I would trust the database state here.

internal/app/cluster-connector/api/v1/remote_cluster.go
edlerd: Yes, exactly. The comment is meant to highlight that fact. I'll adjust it slightly.

internal/app/management-api/api/v1/remote_cluster.go
edlerd: Good point, we should re-send the request with the new token.

edlerd:
Thanks for the review @omarelkashef please have another pass at the changes.

internal/pkg/database/store/remote_cluster_detail.go
omarelkashef: What do you think of `TunnelManagerMemberURL` or `TunnelHostMemberURL`?

internal/pkg/types/rest.go
```diff
@@ -47,6 +49,39 @@ type RouteConfig struct {
...
+
+// Tunnel is the reverse tunnel with a specific MicroCloud.
+type Tunnel struct {
+	Mu           sync.RWMutex
+	WsConn       *websocket.Conn
+	PendingCalls map[string]chan ClusterManagerTunnelResponse
+	UserSessions map[string]string
+}
+
+// ClusterManagerTunnelRequest represents the request received through the tunnel.
```
omarelkashef: These comments are confusing. How about "request forwarded over reverse tunnel" here and "response received from to the reverse tunnel" below?

internal/app/management-api/api/v1/remote_cluster.go
omarelkashef: > I would trust the database state here.

The database state is correct, but it does not guarantee reachability. That would be fine if we do not want the overhead of testing reachability and we will return the error to the caller.

internal/pkg/types/rest.go
```diff
@@ -47,6 +49,39 @@ type RouteConfig struct {
...
+type TunnelStore struct {
+	Mu              sync.RWMutex
+	TunnelByCluster map[int]*Tunnel
+}
+
+// Tunnel is the reverse tunnel with a specific MicroCloud.
+type Tunnel struct {
+	Mu           sync.RWMutex
+	WsConn       *websocket.Conn
+	PendingCalls map[string]chan ClusterManagerTunnelResponse
```
omarelkashef: PendingCalls sounds like “pending send” rather than “waiting response”. How about PendingResponses, InFlightResponseCh, ResponseChByRequestID, etc.?

internal/app/cluster-connector/api/v1/remote_cluster.go
```diff
@@ -392,3 +433,364 @@ func remoteClusterDelete(rc types.RouteConfig) types.EndpointHandler {
...
+		defer func() {
+			logger.Log.Warnw("Client disconnected")
+			tunnel.Mu.Lock()
+			if tunnel.WsConn != conn {
+				// someone else updated the tunnel already, do not close their connection
+				tunnel.Mu.Unlock()
+				return
+			}
+
+			tunnel.WsConn = nil
```
omarelkashef, line 476: Should we also clear UserSessions and PendingCalls (the whole tunnel entry in TunnelByCluster) here to avoid memory leakage or leave UserSessions for future connection? We can also do the cleanup on remote cluster delete.

internal/app/management-api/api/v1/remote_cluster.go
omarelkashef: If we want to keep it this way, we can use "established" or "registered" (at least in the UI) instead of "connected" because I also noticed in the UI we inform the user that the tunnel is connected which implies that they can use it.

omarelkashef (commented):
I did another round of review. Logically, it looks good. Only a few suggestions.

internal/app/cluster-connector/api/v1/remote_cluster.go
```diff
@@ -392,3 +433,364 @@ func remoteClusterDelete(rc types.RouteConfig) types.EndpointHandler {
...
+		}()
+
+		// send request
+		tunnel.Mu.Lock()
+		wsConn := tunnel.WsConn
+		if wsConn == nil {
+			tunnel.Mu.Unlock()
+			return response.SmartError(errors.New("Tunnel not connected")).Render(w, r)
+		}
+		err = wsConn.WriteJSON(req)
```
omarelkashef, line 714: I am not sure that this part where we send the request here and wait for the channel response coming from remoteClusterWsGet is easily understood on the first try, but I can not think of a better way to do it.

internal/pkg/database/store/remote_cluster_detail.go
edlerd: `TunnelManagerMemberURL` seems good, yes will adapt it.

internal/app/management-api/api/v1/remote_cluster.go
edlerd: > If we want to keep it this way, we can use "established" or "registered" (at least in the UI) instead of "connected" because I also noticed in the UI we inform the user that the tunnel is connected which implies that they can use it.

yes, good idea, will adjust it.

internal/pkg/types/rest.go
edlerd: yes, PendingResponses is the right name here!

internal/app/cluster-connector/api/v1/remote_cluster.go
edlerd: Good catch with the pending calls. Yes, we should clear them here and in one other place.

The user sessions stay valid even when the tunnel closes and reopens. They track sessions with LXD, that are independent of which tunnel connection is used.

internal/pkg/types/rest.go
```diff
@@ -47,6 +49,39 @@ type RouteConfig struct {
...
+// ClusterManagerTunnelRequest is forwarded over reverse tunnel.
+type ClusterManagerTunnelRequest struct {
+	UUID    string      `json:"uuid"`
+	Method  string      `json:"method"`
+	Path    string      `json:"path"`
+	Headers http.Header `json:"headers"`
+	Body    []byte      `json:"body"`
+}
+
+// ClusterManagerTunnelResponse is received from to the reverse tunnel.
```
omarelkashef: typo: there is an extra "to" here

omarelkashef (dismissed):
LGTM, thank you for this PR.  A tiny nitpick below.

omarelkashef (approved):

PR #458, opened 2026-03-27
feat: admin vs non-admin access control WD-34391

omarelkashef:
Require admin group to access management-api after authentication.

With admin access
image

Without admin access

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,11 @@ juju integrate self-signed-certificates:send-ca-cert microcloud-cluster-manager-
 juju integrate traefik-k8s:traefik-route microcloud-cluster-manager-k8s
 ```
 
-For authentication you need an OIDC provider. You can use Auth0, Ory Hydra, Keycloak, Microsoft Entra among others. See [LXD documentation on OIDC](https://documentation.ubuntu.com/lxd/latest/howto/oidc/) for how to configure the provider side. The callback path is `/oidc/callback`, but in contrast to LXD it must be without the port 8443 as cluster manager listens on the default port 443.
+For authentication and authorization, you need an OIDC provider. Supported providers include Auth0, Ory Hydra, Keycloak, and Microsoft Entra. See the [LXD documentation on OIDC](https://documentation.ubuntu.com/lxd/latest/howto/oidc/) for guidance on configuring the provider side.
+
+On the provider side, set the OIDC callback path to `/oidc/callback`. Note that unlike LXD, Cluster Manager listens on the default port 443, so the callback URL does not need to include a port number.
+
+To enable group-based authorization, add a custom claim `mcm-idp-groups` to your users' ID tokens. Users must be in the `admins` group to gain access.
 
 Configure the cluster manager charm with your auth provider:
 
--- a/cmd/cluster-connector/cluster-connector.go
+++ b/cmd/cluster-connector/cluster-connector.go
@@ -22,6 +22,7 @@ import (
 	"github.com/canonical/microcloud-cluster-manager/internal/pkg/database/seed"
 	"github.com/canonical/microcloud-cluster-manager/internal/pkg/logger"
 	"github.com/canonical/microcloud-cluster-manager/internal/pkg/middleware"
+	"github.com/canonical/microcloud-cluster-manager/internal/pkg/types"
 	"go.uber.org/automaxprocs/maxprocs"
 	"go.uber.org/zap"
 )
@@ -130,6 +131,14 @@ func Run() (err error) {
 	mtlsAuthenticator := auth.NewMtlsAuthenticator(db)
 
 	// =========================================================================
+	// Initialize authorizor support
+
+	authorizor, err := auth.NewClusterConnectorAuthorizor()
+	if err != nil {
+		return fmt.Errorf("authorizor error: %w", err)
+	}
+
+	// =========================================================================
 	// Initialize rate limiting support
 	tokenBucketRateLimiter := rate_limit.NewRateLimiter(cfg.RateLimitRefillRate, cfg.RateLimitBucketSize, cfg.RateLimitClientActiveInterval, cfg.RateLimitMaxClients, cfg.RateLimitCleanupInterval, cfg.RateLimitLogInterval)
 
@@ -147,7 +156,7 @@ func Run() (err error) {
 		Shutdown:    shutdown,
 		DB:          db,
 		EnvConfig:   cfg,
-		Auth:        mtlsAuthenticator,
+		Auth:        types.Auth{Authenticator: mtlsAuthenticator, Authorizor: authorizor},
 		RateLimiter: tokenBucketRateLimiter,
 	})
 
--- a/internal/app/cluster-connector/core/auth/authorizor.go
+++ b/internal/app/cluster-connector/core/auth/authorizor.go
@@ -1,0 +1,16 @@
+package auth
+
+import (
+	"context"
+)
+
+type ClusterConnectorAuthorizor struct{}
+
+func NewClusterConnectorAuthorizor() (*ClusterConnectorAuthorizor, error) {
+	return &ClusterConnectorAuthorizor{}, nil
+}
+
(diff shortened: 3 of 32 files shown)


cmd/cluster-connector/cluster-connector.go
```diff
@@ -148,6 +156,7 @@ func Run() (err error) {
 		DB:          db,
 		EnvConfig:   cfg,
 		Auth:        mtlsAuthenticator,
+		Authorizor:  authorizor,
```
edlerd: Do we need the Authorizor here? I think we only need it on the management api.

internal/app/management-api/core/auth/oidc.go
```diff
@@ -141,66 +144,99 @@ func (o *Verifier) Auth(ctx context.Context, w http.ResponseWriter, r *http.Requ
...
+		Claims:  claims,
+	}
+}
+
+// extractIdpGroups extracts the user's groups from the OIDC claims and sets the IsAdmin field on the AuthenticationResult
+// if the user is in the "admin" group.
+func extractIdpGroups(claims *oidc.IDTokenClaims, result *AuthenticationResult) {
+	raw, ok := claims.Claims["mcm-idp-groups"]
+	if !ok {
+		logger.Log.Info("AUTHN OIDC groups claim missing")
```
edlerd: I think this message should mention the `mcm-idp-groups` field name we look at.

internal/app/management-api/core/auth/oidc.go
```diff
@@ -141,66 +144,99 @@ func (o *Verifier) Auth(ctx context.Context, w http.ResponseWriter, r *http.Requ
...
+	for _, v := range rawSlice {
+		s, ok := v.(string)
+		if !ok {
+			logger.Log.Info("AUTHN OIDC groups claim malformed")
+			return
+		}
+		groups = append(groups, s)
+	}
+
+	if slices.Contains(groups, "admin") {
```
```suggestion
	if slices.Contains(groups, "admins") {
```
Let's use the same group name as LXD has.

internal/pkg/api/rest.go
```diff
@@ -70,6 +70,18 @@ func registerEndpoint(mux *mux.Router, prefix string, e types.Endpoint, rc types
 			}
 		}
 
+		if !e.AllowUnauthorized {
+			err := rc.Authorizor.CheckPermissions(r.Context(), e.AllowedEntitlements)
+			if err != nil {
+				logger.Log.Errorw("authorization failed", "path", path.Join(prefix, e.Path), "ERROR", err)
```
edlerd: this should have the `AUTH*` prefix

README.md
```diff
@@ -22,7 +22,11 @@ juju integrate self-signed-certificates:send-ca-cert microcloud-cluster-manager-
 juju integrate traefik-k8s:traefik-route microcloud-cluster-manager-k8s
 ```
 
-For authentication you need an OIDC provider. You can use Auth0, Ory Hydra, Keycloak, Microsoft Entra among others. See [LXD documentation on OIDC](https://documentation.ubuntu.com/lxd/latest/howto/oidc/) for how to configure the provider side. The callback path is `/oidc/callback`, but in contrast to LXD it must be without the port 8443 as cluster manager listens on the default port 443.
+For authentication and authorization, you need an OIDC provider. Supported providers include Auth0, Ory Hydra, Keycloak, and Microsoft Entra. See the [LXD documentation on OIDC](https://documentation.ubuntu.com/lxd/latest/howto/oidc/) for guidance on configuring the provider side.
+
+On the provider side, set the OIDC callback path to `/oidc/callback`. Note that unlike LXD, Cluster Manager listens on the default port 443, so the callback URL does not need to include a port number.
+
+To enable group-based authorization, add a custom claim `mcm-idp-groups` to your users' ID tokens containing a list of their group memberships. Users must be in the `admin` group to gain full access.
```
edlerd:
```suggestion
To enable group-based authorization, add a custom claim `mcm-idp-groups` to your users' ID tokens. Users must be in the `admins` group to gain access.
```
A bit clearer?

Also, we should update the cluster manager docs in the microcloud repository with the group step. Maybe file a ticket to do that as a follow up.

internal/app/cluster-connector/core/auth/authorizor.go
edlerd, line 1: I don't think we need this one?

internal/app/management-api/core/auth/authorizor.go
```diff
@@ -0,0 +1,27 @@
...
+}
+
+// CheckPermissions checks if the user has permissions to perform an action based on the user information in the context.
+// It returns an error if the user does not have permissions or if there was an error getting the user information from the context.
+func (a *ManagementAPIAuthorizor) CheckPermissions(ctx context.Context, allowedEntitlements []string) error {
+	userInfo, ok := ctx.Value(types.UserInfoKey).(*types.UserInfo)
+	if !ok {
+		return fmt.Errorf("failed to get user information from the request context")
+	}
+	if !userInfo.IsAdmin {
```
edlerd: Instead of a boolean, let's make this a slice of the groups. So we can check the `admins` group here. Wdyt?

ui/src/components/NotAuthorized.tsx
```diff
@@ -0,0 +1,16 @@
+import { Icon } from "@canonical/react-components";
+import type { FC } from "react";
+
+const NotAuthorized: FC = () => {
+  return (
+    <div className="not-authorized">
+      <Icon name="warning" />
```
edlerd: Icon sits a bit too close to the text, and the text overflows on narrow screens

Image

internal/app/management-api/core/auth/authorizor.go
edlerd: FYI I changed the group in auth0 to `admins` already for the cluster-manager-e2e-tests@example.org
user.

edlerd (commented):
Excellent start! Some observations are below.

cmd/cluster-connector/cluster-connector.go
omarelkashef: `Authorizor` is part of the `RouteConfig` interface, so it either needs to be implemented in cluster connector (even if it is a dummy authorizor) or it can be made nullable and when we check for permission, we let the request through if the authorizor is null. I do not like the latter option because it does not differentiate between an accidental null (forgot to initialize authorizor) or an intentional null.

internal/app/management-api/core/auth/oidc.go
omarelkashef: We will also need to update auth0 config to reflect this.

omarelkashef: I see you addressed that below.

README.md
omarelkashef: Yes, I will create a ticket for it.

cmd/cluster-connector/cluster-connector.go
edlerd: I see, makes sense. :+1: 

Another observation: shall we rename `Auth` to `Authenticator`? Or maybe we should even have a nested structure such as `Auth: { authn, authz }`. Wdyt?

omarelkashef: Yeah, that would be better.

test/helpers/auth.go
```diff
@@ -89,3 +92,17 @@ func LoginToManagementAPI(e *Environment, username string, password string, serv
 
 	return cookies, nil
 }
+
+func GetManagementAPIAuthorizor() (*auth.ManagementAPIAuthorizor, error) {
+	return auth.NewManagementAPIAuthorizor()
+}
+
+func GetContextWithUserInfo(isAdmin bool) context.Context {
```
edlerd: Let's instead pass in the groups as a slice instead of the boolean; then the calling code will be clearer.

edlerd (commented):
Code LGTM, just tiny linter and one nitpick on the test below. Still need to do another round of QA when code is done.

edlerd (approved):
LGTM

PR #468, opened 2026-04-06
ci: add pre-commit hook

omarelkashef:
Add linters to git pre-commit hook.


scripts/pre-commit.sh
edlerd, line 1: Let's move the whole scripts folder to the top level. It seems better suited there. Wdyt?

edlerd (commented):
LGTM, one suggestion below.

scripts/pre-commit.sh
omarelkashef: Yes, the scripts were under test because they did not have a dedicated folder, but now it makes sense to move the dedicated folder to the root.

scripts/setup-env.sh
```diff
@@ -8,7 +8,7 @@ make add-hosts
 
 # generate keys if they don't exist
 SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
-KEY_DIR="$SCRIPT_DIR/keys"
+KEY_DIR="$SCRIPT_DIR/../keys"
```
edlerd: Should this be inside the tests directory as `../tests/keys`?

CONTRIBUTING.md
```diff
@@ -109,6 +110,25 @@ sudo snap install rockcraft --classic
...
+```
+
+This copies `scripts/pre-commit.sh` to `.git/hooks/pre-commit` and makes it executable.
+
+If you prefer to do this manually:
+
+```
+cp scripts/pre-commit.sh .git/hooks/pre-commit
+chmod +x .git/hooks/pre-commit
+```
```
edlerd:
```suggestion
```
That seems overly verbose. One can see the makefile to get the details of it.

edlerd (dismissed):
LGTM, one last nitpick on the docs that I missed last time.

scripts/setup-env.sh
edlerd: already resolved.

edlerd (approved):
LGTM, thanks.

PR #471, opened 2026-04-08
fix(coverage): Copy coverage folder to and from /scripts instead of /tests

omarelkashef:
Copy coverage folder from /scripts instead of /tests


edlerd (approved):
LGTM, thanks :)

PR #499, opened 2026-04-28
feat(test): add auth tests

omarelkashef:
Add e2e auth tests.

### DIFF
--- a/test/e2e/auth.go
+++ b/test/e2e/auth.go
@@ -1,0 +1,441 @@
+package main
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/http"
+	"testing"
+
+	"github.com/canonical/lxd/shared/api"
+	"github.com/canonical/microcloud-cluster-manager/test/helpers"
+	"github.com/google/uuid"
+)
+
+// testAuthAdminUserAllowsAccess verifies that a fully authenticated session with admin group
+// can access a protected endpoint.
+func testAuthAdminUserAllowsAccess(env *helpers.Environment) (testName string, testFunc func(t *testing.T)) {
+	return "admin user is granted access to authenticated endpoint", func(t *testing.T) {
+		const condition = "Should return 200 when admin session cookies are present"
+		cookies, err := helpers.GetValidCookies(env, "cluster-manager-e2e-tests@example.org", "cluster-manager-e2e-password")
+		if err != nil {
+			helpers.LogTestOutcome(t, condition, err)
+			return
+		}
+
+		statusCode, err := helpers.DoRequestWithCookies(env, "/1.0/remote-cluster", http.MethodGet, cookies)
+		if err != nil {
+			helpers.LogTestOutcome(t, condition, fmt.Errorf("request error: %w", err))
+			return
+		}
+
+		if statusCode != http.StatusOK {
+			helpers.LogTestOutcome(t, condition, fmt.Errorf("expected 200, got %d", statusCode))
+			return
+		}
+
+		helpers.LogTestOutcome(t, condition, nil)
+	}
+}
+
+// testAuthNonAdminDenyAccess verifies that a non-admin user is denied access to a protected endpoint.
+func testAuthNonAdminDenyAccess(env *helpers.Environment) (testName string, testFunc func(t *testing.T)) {
+	return "non-admin user is denied access to authenticated endpoint", func(t *testing.T) {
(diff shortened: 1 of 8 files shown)


test/helpers/auth.go
```diff
@@ -104,3 +105,50 @@ func GetContextWithUserInfo(groups []string) context.Context {
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get management API cert: %w", err)
+	}
+
+	cookies, err := LoginToManagementAPI(env, username, password, certPublicKey)
+	if err != nil {
+		return nil, fmt.Errorf("login failed: %w", err)
+	}
+
+	return cookies, nil
```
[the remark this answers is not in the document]
omarelkashef: Caching is not appropriate here since we test with different credentials to test permissions, and adding the logic to verify if the cache has the same credentials as the new request seems out of scope.

test/e2e/auth.go
```diff
@@ -0,0 +1,441 @@
...
+			return
+		}
+
+		helpers.LogTestOutcome(t, condition, nil)
+	}
+}
+
+// testAuthOIDCCallbackWithInvalidStateReturns500 verifies that hitting the callback
+// endpoint with a malformed state parameter returns an error response.
+func testAuthOIDCCallbackWithInvalidStateReturns500(env *helpers.Environment) (testName string, testFunc func(t *testing.T)) {
```
edlerd: Should this be a client error, because the invalid state is provided by them?

edlerd (commented):
This looks pretty good as is, just one concern found below plus the commented-out code.

test/e2e/auth.go
omarelkashef: Yes and we need to change the endpoint response also.

omarelkashef:
> This looks pretty good as is, just one concern found below plus the commented-out code.

The commented out test is for the logout flow which we decided to do postpone and work on a spec first.

edlerd:
> > This looks pretty good as is, just one concern found below plus the commented-out code.
> 
> The commented out test is for the logout flow which we decided to do postpone and work on a spec first.

Perhaps we should remove it and merge without logging out test coverage. Then add it back in later, together with the changes agreed upon in the spec at some future time.

test/e2e/remote_cluster_fail_status.go
```diff
@@ -54,10 +54,10 @@ func testRemoteClusterStatusInvalidCert(env *helpers.Environment) (testName stri
 			}
 
 			err = sendStatusUpdateInvalidCert(env, *tokenData)
-			if err != nil && err.Error() == "Not Found" {
+			if err != nil && err.Error() == "Forbidden" {
 				err = nil
 			} else {
-				err = errors.New("expected not found error not received")
+				err = errors.New("expected forbidden error not received")
```
edlerd, line 60: I think both should be "not found" errors. Otherwise an unauthenticated user can enumerate and check the existence of clusters. I assume the test was masking this error before. Would you mind looking if this can be corrected?

edlerd (commented):
Mostly good, thanks for the tests, one concern raised below.

test/e2e/remote_cluster_fail_status.go
omarelkashef: The certificate validation happens in the auth middleware. An unauthenticated user can not enumerate the clusters because the request will always be rejected with 403 regardless of which cluster they target. Let me know if I understood your concern correctly.

edlerd: Does that mean both existing and nonexistent cluster requests will return `forbidden`?

I was concerned that both cases could be differentiated by an unauthenticated client by comparing the different error messages `not found` vs `forbidden`.

omarelkashef: The request goes through an auth middleware which checks if the certificate is valid/trusted or not. If it is not, it immediately returns Forbidden. If it is, the remote cluster id is extracted and the endpoint runs. If the cluster is not found in the db, it returns Not Found. My understanding of an unauthenticated client here means is one with no valild cluster certificate and they will always get Forbidden.  

https://github.com/canonical/microcloud-cluster-manager/blob/main/internal/app/cluster-connector/core/auth/mtls.go#L37-L69

edlerd: Okay, perfect, that means we do not leak existence information through the error message :+1:

edlerd (approved):
LGTM, thanks :)

PR #504, opened 2026-04-30
chore: lazy load routes and use suspense component WD-36458

omarelkashef:
lazy load routes and use suspense component

### DIFF
--- a/ui/src/App.tsx
+++ b/ui/src/App.tsx
@@ -2,12 +2,13 @@ import type { FC } from "react";
 import { lazy, Suspense } from "react";
 import { Navigate, Route, Routes } from "react-router-dom";
 import NoMatch from "pages/NoMatch";
-import Settings from "pages/Settings";
 import { useAuth } from "context/auth";
 import { logout } from "util/helpers";
-import ClusterDetail from "pages/clusters/ClusterDetail";
+import { Spinner } from "@canonical/react-components";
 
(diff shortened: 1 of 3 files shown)


ui/src/App.tsx
```diff
@@ -36,15 +37,19 @@ const App: FC = () => {
   );
 
   if (isAuthLoading) {
-    return;
+    return <Spinner className="u-loader" text="Loading..." isMainComponent />;
```
edlerd: I think we should not show a spinner here, the loading states now look fairly odd, jumping around. Maybe a blank page as before is better.

https://github.com/user-attachments/assets/3f5117f9-a667-452f-994d-d2d4dc8f34f1

ui/src/App.tsx
```diff
@@ -36,15 +37,19 @@ const App: FC = () => {
...
 
   if (!isAuthenticated) {
     logout();
   }
 
   return (
-    <Suspense>
+    <Suspense
+      fallback={
+        <Spinner className="u-loader" text="Loading..." isMainComponent />
```
edlerd, line 50: The spacing for the loading state is a bit off. There should be more left padding on the loading animation, so around 5 seconds in the video above.

edlerd (commented):
Thanks for opening this PR, the loading states need some design refinement. See video and comments below.

edlerd (approved):
LGTM

PR #532, opened 2026-06-04
feat(auth):  sessions WD-36740

omarelkashef:
Add sessions support.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.26.2
 require (
 	github.com/canonical/lxd v0.0.0-20260105151203-f0abf58d180f
 	github.com/getkin/kin-openapi v0.140.0
+	github.com/golang-jwt/jwt/v5 v5.3.1
 	github.com/golang/snappy v1.0.0
 	github.com/google/uuid v1.6.0
 	github.com/gorilla/mux v1.8.1
--- a/internal/app/management-api/core/auth/oidc.go
+++ b/internal/app/management-api/core/auth/oidc.go
@@ -14,8 +14,12 @@ import (
 	"github.com/canonical/lxd/lxd/response"
 	"github.com/canonical/lxd/lxd/util"
 	"github.com/canonical/lxd/shared"
+	"github.com/canonical/lxd/shared/api"
+	"github.com/canonical/microcloud-cluster-manager/internal/pkg/config"
+	"github.com/canonical/microcloud-cluster-manager/internal/pkg/database"
 	"github.com/canonical/microcloud-cluster-manager/internal/pkg/logger"
 	"github.com/canonical/microcloud-cluster-manager/internal/pkg/types"
+	"github.com/golang-jwt/jwt/v5"
 	"github.com/google/uuid"
 	"github.com/gorilla/securecookie"
 	"github.com/zitadel/oidc/v3/pkg/client/rp"
@@ -25,14 +29,18 @@ import (
 )
 
 const (
-	// cookieNameIDToken is the identifier used to set and retrieve the identity token.
-	cookieNameIDToken = "oidc_identity"
+	// cookieNameSessionToken is used to identify the session. It does not need to be encrypted.
+	cookieNameSessionToken = "oidc_session"
 
-	// cookieNameRefreshToken is the identifier used to set and retrieve the refresh token.
-	cookieNameRefreshToken = "oidc_refresh"
+	// cookieNameUserSecret is the identifier used to set and retrieve the user secret for encrypting tunnel data.
+	cookieNameUserSecret = "user_secret"
 
-	// cookieNameSessionID is used to identify the session. It does not need to be encrypted.
-	cookieNameSessionID = "session_id"
+	// SessionCookieExpiryBuffer extends browser cookie retention after session JWT expiry.
+	// This helps ensure clients still send stale cookies so the server can respond with a deterministic re-authentication path.
+	SessionCookieExpiryBuffer = time.Hour * 24 * 7
+
+	// OIDCSessionExpiry is the duration for the session JWT expiry.
+	OIDCSessionExpiry = time.Hour * 24 * 7
 )
 
 const (
@@ -59,15 +67,13 @@ type Verifier struct {
 	// and will be refreshed. This refreshes the cookie encryption keys that the relying party uses.
 	configExpiry         time.Time
 	configExpiryInterval time.Duration
+
+	sessionHandler *SessionHandler
 }
 
-// AuthenticationResult represents an authenticated OIDC client.
-type AuthenticationResult struct {
-	Subject string
-	Email   string
-	Name    string
-	Claims  *oidc.IDTokenClaims
-	Groups  []string
+// sessionTokenClaims represents claims for the signed session token.
+type sessionTokenClaims struct {
+	jwt.RegisteredClaims
 }
 
 // AuthError represents an authentication error. If an error of this type is returned, the caller should call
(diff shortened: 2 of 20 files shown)


internal/app/management-api/core/auth/session.go
```diff
@@ -0,0 +1,149 @@
...
+			return nil, fmt.Errorf("Failed to update identity: %w", err)
+		}
+		return identity, nil
+	}
+
+	if !sharedapi.StatusErrorCheck(err, http.StatusNotFound) {
+		return nil, fmt.Errorf("Failed to get identity: %w", err)
+	}
+
+	identityRow := store.Identity{
```
edlerd:
```suggestion
    identity = &store.Identity{
```
Let's use the same variable name here.

internal/app/management-api/core/auth/session.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+// UpdateSessionTokens persists re-encrypted tokens for the given session.
+func (s *SessionHandler) UpdateSessionTokens(ctx context.Context, session *store.Session, encryptedTokens tokenSet) error {
+	if s == nil || s.db == nil {
+		return errors.New("Missing session handler")
+	}
+
+	updatedSession := *session
+	updatedSession.IDTokenEncrypted = encryptedTokens.IDToken
```
edlerd:
```suggestion
	session.IDTokenEncrypted = encryptedTokens.IDToken
```
I think we don't need the alias here.

internal/app/management-api/core/auth/oidc.go
```diff
@@ -120,39 +135,251 @@ func DecodeStateToken(token string) (StateToken, error) {
...
+
+	err = o.sessionHandler.UpdateSessionTokens(ctx, session, encryptedTokens)
+	if err != nil {
+		return nil, AuthError{Err: fmt.Errorf("Failed to update session tokens: %w", err)}
+	}
+
+	// Keep in-memory session data in sync so callers can immediately reuse refreshed values.
+	session.IDTokenEncrypted = encryptedTokens.IDToken
+	session.AccessTokenEncrypted = encryptedTokens.AccessToken
+	session.RefreshTokenEncrypted = encryptedTokens.RefreshToken
```
```suggestion
	session.IDTokenEncrypted = encryptedTokens.IDToken
	session.AccessTokenEncrypted = encryptedTokens.AccessToken
	session.RefreshTokenEncrypted = encryptedTokens.RefreshToken

	err = o.sessionHandler.UpdateSessionTokens(ctx, session)
	if err != nil {
		return nil, AuthError{Err: fmt.Errorf("Failed to update session tokens: %w", err)}
	}
```
This is a bit backwards. Let's update it first and then save it. We might not need the `UpdateSessionTokens` function after all.

internal/app/management-api/core/auth/oidc.go
```diff
@@ -120,39 +135,251 @@ func DecodeStateToken(token string) (StateToken, error) {
 	return stateToken, nil
 }
 
-// Auth extracts OIDC tokens from the request, verifies them, and returns the subject.
+// tokenSet groups the three OIDC token strings passed between authentication helpers.
+type tokenSet struct {
+	IDToken      string
+	AccessToken  string
+	RefreshToken string
+}
```
edlerd: Can we pass around the plain `oidc.Tokens[*oidc.IDTokenClaims]`. I don't think we need the indirection here.

We might keep this, use it only for the encrypted token set, so would recommend renaming it as such. Wdyt?

internal/app/management-api/core/auth/oidc.go
```diff
@@ -120,39 +135,251 @@ func DecodeStateToken(token string) (StateToken, error) {
...
+	userSecret, err := o.getUserSecretFromCookie(r, sessionID)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to get user secret: %w", err)
 	}
 
-	if idToken == "" && refreshToken == "" {
-		return false, fmt.Errorf("No ID or refresh tokens present in request")
+	session, err := o.sessionHandler.GetSessionByID(r.Context(), sessionID)
+	if err != nil {
+		o.deleteCookies(w)
```
edlerd, line 186: Deleting cookies might be wrong here. This means if the database is locked or unresponsive we delete users sessions.

internal/pkg/database/store/sessions.go
```diff
@@ -0,0 +1,256 @@
...
+		return nil
+	}
+
+	err := query.Scan(ctx, tx, q, dest)
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch from \"sessions\" table: %w", err)
+	}
+
+	return objects, nil
+}
```
edlerd: This is unused, can we remove it? 
```suggestion
```

go.mod
```diff
@@ -32,6 +32,7 @@ require (
 	github.com/go-openapi/jsonpointer v0.23.1 // indirect
 	github.com/go-openapi/swag/jsonname v0.26.0 // indirect
 	github.com/gogo/protobuf v1.3.2 // indirect
+	github.com/golang-jwt/jwt/v5 v5.3.1 // indirect
```
edlerd: Should this be a direct dependency?

edlerd (commented):
Some observations below.

internal/app/management-api/core/auth/oidc.go
```diff
@@ -588,11 +783,75 @@ func NewVerifier(issuer string, clientID string, clientSecret string, audience s
...
+	if err != nil {
+		return tokenSet{}, fmt.Errorf("Failed to decrypt access token: %w", err)
+	}
+
+	refreshToken, err := config.DecryptUserValue(ts.RefreshToken, userSecret)
+	if err != nil {
+		return tokenSet{}, fmt.Errorf("Failed to decrypt refresh token: %w", err)
+	}
+
+	return tokenSet{
```
edlerd: Here we should not return a token set as it is now unencrypted.

internal/app/management-api/core/auth/oidc.go
edlerd: Let's check if the error is not found and only delete in that case.

internal/app/management-api/core/auth/oidc.go
omarelkashef: It looks like that is unnecessary. `UpdateSessionTokens` already handles updating the session fields.

edlerd: I was thinking to write the new tokens into the session here and turn UpdateSessionTokens into a store new tokens function.

omarelkashef: The database is not exposed in the Verifier. It is only available for the session handler, so all session-related db operations go through the session handler. Do you want to expose the db in the verifier?

internal/pkg/database/store/sessions.go
```diff
@@ -0,0 +1,219 @@
...
+	IdentityID            int       `db:"identity_id"`             // Foreign key to identities
+	ExpiresAt             time.Time `db:"expires_at"`              // Session expiration timestamp
+	CreatedAt             time.Time `db:"created_at"`              // Creation timestamp
+	IDTokenEncrypted      string    `db:"id_token_encrypted"`      // Encrypted ID token
+	AccessTokenEncrypted  string    `db:"access_token_encrypted"`  // Encrypted access token
+	RefreshTokenEncrypted string    `db:"refresh_token_encrypted"` // Encrypted refresh token
+}
+
+// GetSessionByIdentityID returns the session ID by identity ID.
+func GetSessionByIdentityID(ctx context.Context, tx *sqlx.Tx, identityID int) (string, error) {
```
edlerd:
```suggestion
func GetSessionIDByIdentityID(ctx context.Context, tx *sqlx.Tx, identityID int) (string, error) {
```
This only returns the id, no?

internal/pkg/database/store/sessions.go
```diff
@@ -0,0 +1,219 @@
...
+
+	if err != nil {
+		return "", fmt.Errorf("failed to get \"sessions\" ID: %w", err)
+	}
+
+	return id, nil
+}
+
+// SessionExists checks if a session exists for a given identity ID.
+func SessionExists(ctx context.Context, tx *sqlx.Tx, identityID int) (bool, error) {
```
edlerd:
```suggestion
func SessionExistsForIdentity(ctx context.Context, tx *sqlx.Tx, identityID int) (bool, error) {
```
A bit more precise?

internal/app/management-api/core/auth/oidc.go
```diff
@@ -252,19 +384,33 @@ func (o *Verifier) Login(w http.ResponseWriter, r *http.Request, stateTokenStr s
...
-		if err != nil {
-			logger.Log.Errorw("Failed rendering internal server error response due to failed deletion of login information: %w", err)
-		}
+		// Not logged in.
+		return
+	}
+
+	sessionID, err := o.verifySessionAndGetID(sessionCookie.Value)
+	if err != nil {
+		// Not logged in.
```
edlerd, line 441: On expiration error we should clean up the db here.

internal/app/management-api/core/auth/oidc.go
```diff
@@ -120,119 +127,244 @@ func DecodeStateToken(token string) (StateToken, error) {
...
 	}
-	result.Groups = groups
 }
 
-// authenticateIDToken verifies the identity token and returns the authentication result and the ID token. If no identity token is given (or
-// verification fails) it will attempt to refresh the ID token.
-func (o *Verifier) authenticateIDToken(ctx context.Context, w http.ResponseWriter, token string, refreshToken string) (result *AuthenticationResult, idToken string, e error) {
+// validateAndRefreshTokens validates tokens and, if they are expired or invalid, refreshes and persists them in the session and db.
+// userSecret is the per-user AES-256 key used to re-encrypt refreshed tokens before persisting them.
+func (o *Verifier) validateAndRefreshTokens(ctx context.Context, session *store.Session, idToken, accessToken, refreshToken, userSecret string) (*oidc.IDTokenClaims, error) {
```
edlerd: accessToken is unused.

internal/app/management-api/core/auth/oidc.go
```diff
@@ -120,119 +127,244 @@ func DecodeStateToken(token string) (StateToken, error) {
...
-			return
-		}
-		groups = append(groups, s)
+
+	refreshToken, err := config.DecryptUserValue(session.RefreshTokenEncrypted, userSecret)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to decrypt refresh token: %w", err)
+	}
+
+	claims, err := o.validateAndRefreshTokens(ctx, session, "", "", refreshToken, userSecret)
```
edlerd: This internally writes to the db and deletes the old session. It would be cleaner if `refreshAndPersistTokens` would only return the new `EncryptedTokenSet` without any persisting. The persisting can be done by `StartSession`.

internal/app/management-api/core/auth/oidc.go
```diff
@@ -120,119 +127,244 @@ func DecodeStateToken(token string) (StateToken, error) {
...
 
-	idTokenAny := tokens.Extra("id_token")
-	if idTokenAny == nil {
-		return nil, "", AuthError{Err: errors.New("ID tokens missing from OIDC refresh response")}
+	return claims, nil
+}
+
+// refreshAndPersistTokens fetches a new token set using the refresh token, verifies the refreshed ID token,
+// re-encrypts all tokens with the user secret, and persists them to the session and db.
+func (o *Verifier) refreshAndPersistTokens(ctx context.Context, session *store.Session, refreshToken, userSecret string) (*oidc.IDTokenClaims, error) {
```
edlerd: This is a code smell. Any function should only do one thing, so if there is an and in the name, better break it down. See my other comment, I think we don't have to persist here and it would simplify the flow.

edlerd (commented):
Good progress, few concerns left. Didn't manage to review the tests yet.

internal/app/management-api/core/auth/oidc.go
omarelkashef: We already clean up old sessions when user logs in again. Inside the if statement you highlighted, we do not have the session ID to delete it.

edlerd: I think there is an option that `verifySessionAndGetID` returns a session id and an error. In that case we miss to clean up the session from the db.

edlerd: Yes, the session gets cleaned up on login, but we don't know here if the user will successfully re-login.

internal/app/management-api/core/auth/oidc.go
```diff
@@ -588,20 +734,36 @@ func NewVerifier(issuer string, clientID string, clientSecret string, audience s
...
+
+	return key, nil
+}
+
 func getCallbackURL(host string) string {
 	return fmt.Sprintf("https://%s/oidc/callback", host)
 }
 
-func setUserInfoInRequest(authResult *AuthenticationResult, r *http.Request) {
+func setUserInfoInRequest(email, name string, groups []string, r *http.Request) {
```
edlerd, line 799: We also need to store the session or the details of it in the context. Later, consumers will need to read the access token to use it for the tunnel request. It might be enough to add only the access token to the context.

edlerd: Actually might be better not to mix the concerns here, instead set something like this as a new context:

```
	accessTokenContext := context.WithValue(r.Context(), types.AccessTokenKey, accessToken)
	*r = *r.WithContext(accessTokenContext)
```

Because the user info context is exposed via api later on.

edlerd: Actually I also need the user secret, so maybe expose both as individual contexts.
```
	accessTokenContext := context.WithValue(r.Context(), types.UserAccessToken, accessToken)
	*r = *r.WithContext(accessTokenContext)

	userSecretContext := context.WithValue(r.Context(), types.UserSecret, userSecret)
	*r = *r.WithContext(userSecretContext)
```

 Or better would be to combine them into a single one, but dedicated from the user info context.

omarelkashef: I think a better idea would be to create an AccessTokenProvider and have the tunnel routes access it. That way, we do not expose the implementation of refreshing and decrypting acceess tokens.

edlerd: Agree on a common struct to bundle both into the context. Maybe better call it `UserAuthentication` or `UserCredentials`, wdyt?

internal/app/management-api/core/auth/oidc.go
```diff
@@ -120,119 +126,295 @@ func DecodeStateToken(token string) (StateToken, error) {
...
+	setUserInfoInRequest(claims.Email, claims.Name, groups, r)
+
+	provider := &SessionCredentialsProvider{
+		verifier:             o,
+		sessionID:            sessionID,
+		userSecret:           userSecret,
+		encryptedAccessToken: encryptedAccessToken,
+	}
+
+	providerCtx := context.WithValue(r.Context(), SessionCredentialsProviderKey, provider)
```
edlerd, line 157: Nice 👍

internal/app/management-api/core/auth/oidc.go
```diff
@@ -120,119 +126,295 @@ func DecodeStateToken(token string) (StateToken, error) {
...
+		if sessionIDClaim == "" {
+			return nil, errors.New("Missing session ID in JWT payload")
+		}
+
+		parsedSessionID, err := uuid.Parse(sessionIDClaim)
+		if err != nil {
+			return nil, fmt.Errorf("Invalid session ID in JWT payload: %w", err)
+		}
+
+		sessionID = parsedSessionID
```
```suggestion
		sessionID, err := uuid.Parse(sessionIDClaim)
		if err != nil {
			return nil, fmt.Errorf("Invalid session ID in JWT payload: %w", err)
		}
```
We can skip the alias here. Wdyt?

internal/app/management-api/core/auth/oidc.go
```diff
@@ -120,119 +126,295 @@ func DecodeStateToken(token string) (StateToken, error) {
...
+		IDToken:      session.IDTokenEncrypted,
+		AccessToken:  session.AccessTokenEncrypted,
+		RefreshToken: session.RefreshTokenEncrypted,
+	}, userSecret)
+
+	if err != nil {
+		return nil, uuid.Nil, "", "", fmt.Errorf("Failed to decrypt tokens for expired session handling: %w", err)
+	}
+
+	claims, err := rp.VerifyIDToken[*oidc.IDTokenClaims](ctx, idToken, o.relyingParty.IDTokenVerifier())
```
edlerd, line 282: We are inside `handleExpiredSession` and I think we should not try to validate the id token again, but directly go into refresh tokens. Why do we try to validate the id token again in here?

If we go directly into the refresh tokens block, we can remove lines 305 to 319 below.

internal/app/management-api/core/auth/oidc.go
```diff
@@ -120,119 +126,295 @@ func DecodeStateToken(token string) (StateToken, error) {
...
 	}
 
-	extractIdpGroups(result.Claims, result)
-	setUserInfoInRequest(result, r)
+	// Decrypt tokens stored in the database using the per-user secret from the cookie.
+	idToken, _, refreshToken, err := DecryptTokens(EncryptedTokenSet{
+		IDToken:      session.IDTokenEncrypted,
+		AccessToken:  session.AccessTokenEncrypted,
+		RefreshToken: session.RefreshTokenEncrypted,
+	}, userSecret)
```
edlerd: We have to do this translation from `session` to `EncryptedTokenSet` a few times. Can we change the session interface, so that after loading a session we directly have `session.encryptedTokens` instead of `session.IDTokenEncrypted`, `session.AccessTokenEncrypted`, etc.?

So that on loading a session, we get a nested `EncryptedTokenSet` right away. I think that will simplify a bit in here and above.

internal/app/management-api/core/auth/session.go
```diff
@@ -0,0 +1,154 @@
...
+
+	if !sharedapi.StatusErrorCheck(err, http.StatusNotFound) {
+		return nil, fmt.Errorf("Failed to get identity: %w", err)
+	}
+
+	identity = &store.Identity{
+		Subject:     claims.Subject,
+		Name:        claims.Name,
+		Email:       claims.Email,
+		DisplayName: claims.Name,
```
edlerd: Should we guard against empty values in the `claims`?

internal/app/management-api/core/auth/session_credentials_provider.go
```diff
@@ -0,0 +1,68 @@
...
+	accessToken, err := DecryptToken(p.encryptedAccessToken, p.userSecret)
+	if err != nil {
+		return "", err
+	}
+
+	return accessToken, nil
+}
+
+// RefreshSessionTokens refreshes OIDC tokens, persists them, and returns the refreshed token.
+func (p *SessionCredentialsProvider) RefreshSessionTokens(ctx context.Context) (string, error) {
```
edlerd, line 38: Nice, this makes it straightforward and clear in the other PR to adjust.

edlerd (commented):
Reviewed all code, left the test code for future review. Few comments below.

internal/app/management-api/core/auth/oidc.go
omarelkashef: Removing the validation will simplify the code indeed. The intention was to ensure that new sessions start with tokens that are not expired. The tokens are not checked later in auth, so it would be fine to remove this part. The only thing I can think of is that if the user claims change, we will start the new session with the old ones, and upon subsequent requests the expired tokens will be detected and the claims will be updated.

omarelkashef: A middle ground that will simplify the code but keep the logic is to only do the validation in `authenticate()` and not return early in line 176. That way `handleExpiredSession()` will only be responsible for starting a new session and then all session authentication and validation logic happen in `authenticate()`

edlerd: > Removing the validation will simplify the code indeed. The intention was to ensure that new sessions start with tokens that are not expired. The tokens are not checked later in auth, so it would be fine to remove this part. The only thing I can think of is that if the user claims change, we will start the new session with the old ones, and upon subsequent requests the expired tokens will be detected and the claims will be updated.

I don't think we should check expiration status of the tokens in here. The function `handleExpiredSession` should safely assume the tokens are expired.

I don't understand the other concern on changed `claims`. The call `o.refreshTokens` returns the new claims as far as I can see.

edlerd: Or does the verification happen locally? My main concern was that we do too many round trips to the oidc provider here causing delays.

omarelkashef: Yes, all the verification by VerifyIDToken is local, but for simplicity we can always refresh the tokens with the new sessions.

edlerd (approved):
LGTM, thanks. The oidc.go file is still with a lot of responsibility, but I have no direct idea how to simplify it.

PR #558, opened 2026-06-26
chore: update db code to return db errors instead of api errors

omarelkashef:


edlerd (commented):
I am not sure if this is the right thing to do. It seems an established pattern also in LXD to return http statuses from the database layer. An example is in [profiles.mapper.go](https://github.com/canonical/lxd/blob/main/lxd/db/cluster/profiles.mapper.go#L528). To me it seems okay, and not fundamentally wrong, also it seems simpler.

Shall we keep things as they are right now?

omarelkashef:
> I am not sure if this is the right thing to do. It seems an established pattern also in LXD to return http statuses from the database layer. An example is in [profiles.mapper.go](https://github.com/canonical/lxd/blob/main/lxd/db/cluster/profiles.mapper.go#L528). To me it seems okay, and not fundamentally wrong, also it seems simpler.
> 
> Shall we keep things as they are right now?

I also noticed it is a pattern in LXD, but I am not sure it is correct. It seems wrong having api errors in the db layer.

edlerd:
> I also noticed it is a pattern in LXD, but I am not sure it is correct. It seems wrong having api errors in the db layer.

Can you reach out to the LXD team? Let's follow their standards and only make an exception if they would like to go that way too.

omarelkashef:
Closed to keep the things aligned with the existing pattern in LXD.

PR #606, opened 2026-08-10
fix(api) ensure user values are escaped on api requests

edlerd:
# Done
- ensure user values are escaped on api requests


omarelkashef (approved):
LGTM, thank you!

PR #613, opened 2026-08-21
feat(cluster-links) enable fetching cluster links from remote cluster

edlerd:
# Done
- feat(cluster-links) enables fetching cluster links from a remote cluster
- lock down remote cluster tunnels

### DIFF
--- a/internal/app/cluster-connector/api/v1/remote_cluster.go
+++ b/internal/app/cluster-connector/api/v1/remote_cluster.go
@@ -82,19 +82,9 @@ var RemoteClusterInternal = types.RouteGroup{
 	},
 	Endpoints: []types.Endpoint{
 		{
-			Path:    "{remoteClusterName}/tunnel/{path:.*}",
+			Path:    "{remoteClusterName}/cluster-links",
 			Method:  http.MethodGet,
-			Handler: remoteClusterTunnel,
-		},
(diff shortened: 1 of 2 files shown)


omarelkashef (dismissed):
Logic looks good to me. One question and it would be ready to merge. Should we also make the cluster-connector tunnel endpoints opinionated and not expose more than we need?

edlerd:
> Should we also make the cluster-connector tunnel endpoints opinionated and not expose more than we need?

Good question, I was wondering the same but then kept it unchanged. Thinking about it now, we could have a list of allowed paths that we maintain and check against. I would want to avoid one API endpoint per path to make it easy to maintain in the future. What do you think @omarelkashef ?

omarelkashef:
> > Should we also make the cluster-connector tunnel endpoints opinionated and not expose more than we need?
> 
> Good question, I was wondering the same but then kept it unchanged. Thinking about it now, we could have a list of allowed paths that we maintain and check against. I would want to avoid one API endpoint per path to make it easy to maintain in the future. What do you think @omarelkashef ?

Yes, my concern about my suggested approach was maintainability. So, your idea of an allowed list seems reasonable to me!

omarelkashef (approved):
I like the new structure. Thank you!

PR #632, opened 2026-09-09
feat(cluster-link) spike for displayal and managing of cluster links

edlerd:
# Done
- display cluster links
- allow creation of cluster links

# Screenshots

## Create

image

## View

image

## Delete

image

### DIFF
--- a/internal/app/management-api/api/v1/remote_cluster.go
+++ b/internal/app/management-api/api/v1/remote_cluster.go
@@ -1,6 +1,7 @@
 package v1
 
 import (
+	"bytes"
 	"context"
 	"crypto/tls"
 	"crypto/x509"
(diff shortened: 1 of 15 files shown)


internal/app/management-api/api/v1/remote_cluster.go
```diff
@@ -313,12 +343,28 @@ func forwardToClusterConnector(rc types.RouteConfig, w http.ResponseWriter, r *h
...
+	}
+
+	defer func() {
+		err := r.Body.Close()
+		if err != nil {
+			logger.Log.Warnw("Failed to close request body", "error", err)
+		}
+	}()
+
+	resp, err := sendInternalRequest(r, accessToken, userSecret, reqURL, rc, body)
```
omarelkashef: What do you think about creating a function `sendRetryableRequest` that takes care of the first attempt, token refresh, and retry. We can also start using it in all tunnel requests?

omarelkashef (commented):
I only reviewed the stable golang part. A small comment. QA looks good.

internal/app/management-api/api/v1/remote_cluster.go
edlerd: I assumed the current parent of this `forwardToClusterConnector` is already generic enough. It should be fine as an entry point for future UI features that rely on the tunnel. Wdyt?

omarelkashef: Yes, that is true. But it is confusing why sendInternalRequest is used twice and the retry block takes many lines of code and have some implementation details that the parent function does not care about. But this is not a blocker.

PR #641, opened 2026-09-15
feat: rename ui_url field to lxd_url

omarelkashef:
## ⚠️ Do not merge before canonical/microcloud#1539 ⚠️

This PR matches the API contract change in canonical/microcloud#1539 (still in review), which renames the `ui_url` field to `lxd_url` in the status payload MicroCloud sends to cluster manager. Merging this before the MicroCloud PR would break status updates, since MicroCloud still sends `ui_url`.

## Done

- Renamed `ui_url` → `lxd_url` in `RemoteClusterStatusPost` (incoming status payload) and `RemoteCluster` (API response) models, matching the updated `cluster-manager-api.yaml` spec in the MicroCloud PR
- Renamed the `ui_url` column to `lxd_url` in `remote_cluster_details` via a new goose migration (`00004_db.sql`, with down migration)
- Updated the UI (`cluster.lxd_url`, `ClusterUiButton` prop) to match the renamed response field

## QA

1. Check out canonical/microcloud#1539 and set up a MicroCloud cluster connected to this cluster manager
2. Verify status updates are accepted and stored (`lxd_url` populated in `remote_cluster_details`)
3. Verify `GET /1.0/remote-cluster` and `GET /1.0/remote-cluster/{name}` return `lxd_url` (and no `ui_url`)
4. Verify the "LXD UI" button on the clusters page links to the LXD URL
5. Verify migration up/down: `ui_url` ↔ `lxd_url` column rename on `remote_cluster_details`


internal/pkg/database/store/remote_cluster_detail.go
```diff
@@ -350,8 +350,7 @@ var baseDetailQuery = `
 		remote_cluster_details.member_count,
 		remote_cluster_details.member_statuses,
 		remote_cluster_details.storage_pool_usages,
-		remote_cluster_details.ui_url,
-		remote_cluster_details.tunnel_manager_member_url,
```
edlerd: We should not remove the `tunnel_manager_member_url` here.

internal/pkg/database/store/remote_cluster_detail.go
```diff
@@ -394,8 +393,7 @@ func getRemoteClusterWithDetails(ctx context.Context, tx *sqlx.Tx, sql string, a
 			&c.MemberCount,
 			&c.MemberStatuses,
 			&c.StoragePoolUsages,
-			&c.UIURL,
-			&c.TunnelManagerMemberURL,
```
edlerd: We need the `TunnelManagerMemberURL`, don't we?

edlerd (commented):
There seems a bug being introduced below. Also can we keep it as draft until the blocked PRs got in?

internal/pkg/database/store/remote_cluster_detail.go
omarelkashef: Yes that should not have been removed!

internal/pkg/database/store/remote_cluster_detail.go
omarelkashef: Fixed!

edlerd (approved):

edlerd (changes requested):
openapi spec needs adjustments

omarelkashef:
> openapi spec needs adjustments

The API schema tests are failing because the microcloud pr to rename the field is not merged yet.

PR #643, opened 2026-09-16
chore(cluster) align instance status stopped icon with lxd ui

edlerd:
# Done
- chore(cluster) align instance status stopped icon with lxd ui


omarelkashef (approved):
LGTM!

